PR #2, opened 2016-12-07
Build setup improvements + node-sdk packages

emmenko:
#### Summary
I finally took some time to continue work on the new sdk. Since I want to have it split into different packages, what better chance to start doing it in our new monorepo ;)

> Please **don't bother much** reviewing the `packages/*` stuff, it's still very much in progress. It's just there as a starting point, and **to test the different script commands** for packages usage.

So **focus on the setup changes** if you are reviewing this PR 😉 .
I'll add some inline comments to explain some of the changes. Note that some stuff got removed / refactored because it didn't really work as expected with the packages, since we didn't really have proper packages to test it yet.

> My plan is to merge this, to set the basic necessary setup, and continue working on the packages in separate PRs

TODO:
- [x] fix travis build


.babelrc
```diff
@@ -1,8 +1,7 @@
 {
   "presets": ["es2015"],
-  "env": {
-    "test": {
-      "plugins": [ "istanbul" ]
-    }
-  }
+  "plugins": [
+    "transform-export-extensions",
```
emmenko: This is useful for having proxy exports

```js
// index.js
export * from './my-module'
```

.babelrc
```diff
@@ -1,8 +1,7 @@
...
+    "transform-object-rest-spread"
```
emmenko: This is useful for spreading objects, instead of using `Object.assign`

```js
const state = { foo: 'bar' }
const o = Object.assign({}, state, { foo: 'baz' })

const o = { ...state, { foo: 'baz' } }
```

package.json
```diff
@@ -1,66 +1,74 @@
...
-    "check-updates": "run-p -s check-updates:*",
-    "check-updates:root": "ncu --error-level 2",
-    "check-updates:packages": "lerna exec -- ncu --error-level 2",
-    "upgrade": "run-p -s check-updates:*",
-    "upgrade:root": "ncu --loglevel minimal --upgrade",
-    "upgrade:packages": "lerna exec -- ncu --loglevel minimal --upgrade",
+    "postinstall": "check-node-version --package && lerna bootstrap",
+    "check-updates": "./scripts/check-updates",
+    "clean": "rm -rf ./node_modules && rm -rf ./packages/*/node_modules",
+    "build": "lerna exec -- npm run build",
```
emmenko: This does not work at the moment, as no packages have any build scripts

package.json
```diff
@@ -1,66 +1,74 @@
...
+    "lint": "./scripts/lint packages",
+    "test": "lerna exec -- jest",
+    "test:package": "./scripts/test-package",
```
emmenko: Useful to run tests for a single package.

Usage:

```js
npm run test:package commercetools-sdk
npm run test:package commercetools-sdk --watch
npm run test:package commercetools-sdk --coverage
```

package.json
```diff
@@ -70,5 +78,28 @@
         }
       }
     }
+  },
+  "jest": {
```
emmenko: 🎉

scripts/check-updates
```diff
@@ -0,0 +1,16 @@
...
+eval "$updateCmd --error-level 2"
+resultRoot="$?"
+
+# packages dependencies
+eval "$lernaCmd exec -- $updateCmd --error-level 2"
+resultPackages="$?"
+
+if [ "$resultRoot" -eq 1 ] || [ "$resultPackages" -eq 1 ]; then
+  exit 1
+fi
```
emmenko: This allows to run the check in the root folder and in all packages first, then exiting with an error based on the command results.
So basically the script will run to the end before e.g. erroring

scripts/symlink-bin.js
```diff
@@ -1,42 +0,0 @@
...
-
-  files.forEach(function (file) {
-    const rootBinPath = path.resolve(rootBin, file);
-    const packageBinPath = path.resolve(packageBin, file);
-
-    forceSymlink(rootBinPath, packageBinPath);
-  });
-
-  console.log(`Created symlinks for .bin in ${process.cwd()}`);
-});
```
emmenko: I didn't find this necessary anymore, it was causing a bit of confusion. That's also part of the changes in the main scripts.
Additionally, packages don't need to define any "scripts" themselves, besides is something really specific.

packages/commercetools-sdk-middleware-http/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "author": "",
+  "main": "",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/commercetools/nodejs.git"
+  },
+  "dependencies": {
+    "isomorphic-fetch": "^2.2.1"
+  }
+}
```
emmenko: Really simple package definition. Only metadata and no boilerplate 🍻

packages/commercetools-sdk-middleware-auth/README.md
```diff
@@ -0,0 +1,39 @@
...
+- `refresh-token-flow` ([link](http://dev.commercetools.com/http-api-authorization.html#refresh-token-flow)) _(TBD)_
+- `anonymous-session-flow` ([link](http://dev.commercetools.com/http-api-authorization.html#tokens-for-anonymous-sessions)) _(TBD)_
+
+## Usage
+
+```js
+// Middleware for client credentials flow
+import { createClient } from '@commercetools/sdk'
+import {
+  createAuthMiddlewareForClientCredentialsFlow,
```
PhilippSpo: 👀

.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
hisabimbola: what is this file for?

package.json
```diff
@@ -1,66 +1,77 @@
...
-    "check-updates:packages": "lerna exec -- ncu --error-level 2",
-    "upgrade": "run-p -s check-updates:*",
-    "upgrade:root": "ncu --loglevel minimal --upgrade",
-    "upgrade:packages": "lerna exec -- ncu --loglevel minimal --upgrade",
+    "postinstall": "check-node-version --package && lerna bootstrap",
+    "check-updates": "./scripts/check-updates",
+    "clean": "./scripts/clean-all",
+    "build": "./scripts/build-packages",
+    "lint": "./scripts/lint packages",
+    "test": "lerna exec -- jest",
```
hisabimbola: How do we intend to pass project credentials?

packages/commercetools-sdk-middleware-auth/README.md
```diff
@@ -0,0 +1,39 @@
+# commercetools-sdk-middlware-auth
+
+Auth middlewares collection for usage with `@commercetools/sdk`
```
hisabimbola: Shouldn't we add _nodejs_ to this. More like `@commercetools/nodejs-sdk`

packages/commercetools-sdk-middleware-auth/README.md
hisabimbola: Erm, too long I guess

packages/commercetools-sdk-middleware-auth/src/index.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable max-len */
```
hisabimbola: 😺 😈 

Possible to find new variable names?

packages/commercetools-sdk-middleware-http/package.json
```diff
@@ -0,0 +1,28 @@
+{
```
hisabimbola: Any reason why the file names have `commercetools` in them?

Shouldn't that be more explicit

scripts/check-npm.sh
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env bash
-
-if [[ `npm -v` != 3* ]]
-  then
-    npm install --global npm
-fi
```
hisabimbola: Sweet 🍬

scripts/clean-all
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+cmd="rm -rf node_modules; rm -rf coverage; rm -rf lib"
```
hisabimbola: Under which scenario do we want to _clean-all_ including *node_modules*?

scripts/lint
```diff
@@ -0,0 +1,11 @@
...
+
+# Lints all files passed as arguments,
+# unless they are ignored by .eslintignore
+
+# folders/files from arguments to lint
+FILES="${*:1}"
+
+./node_modules/.bin/eslint $FILES \
+  --format=node_modules/eslint-formatter-pretty\
+  --rule 'import/no-unresolved: 2'
```
hisabimbola: why do we have this extra rule?

emmenko:
@Siilwyn @hisabimbola so, I drafted a script to execute custom scripts from within the individual packages, if defined. See https://github.com/commercetools/nodejs/pull/2/commits/e6310918887186a6a2e7a4ed0c1ee2d37f969f33

task-run

What do you think?

scripts/check-npm.sh
Siilwyn: `check-node-version` does not install the latest npm version right? This script is in place so that Travis works with Node 4, though I don't remember exactly why.

package.json
```diff
@@ -1,66 +1,79 @@
 {
   "private": true,
-  "license": "MIT",
```
Siilwyn: Having the license field in our root is nice because GH will show it in the repo information.

package.json
```diff
@@ -1,66 +1,79 @@
...
-    "commit": "git-cz || true",
-    "check-updates": "run-p -s check-updates:*",
-    "check-updates:root": "ncu --error-level 2",
-    "check-updates:packages": "lerna exec -- ncu --error-level 2",
-    "upgrade": "run-p -s check-updates:*",
-    "upgrade:root": "ncu --loglevel minimal --upgrade",
-    "upgrade:packages": "lerna exec -- ncu --loglevel minimal --upgrade",
+    "postinstall": "check-node-version --package && lerna bootstrap",
+    "check-updates": "./scripts/check-updates",
+    "clean": "rm -rf coverage && ./scripts/run-task clean",
```
Siilwyn: Why not `lerna clean --yes`?

package.json
```diff
@@ -1,66 +1,79 @@
...
-    "clean": "lerna exec -- rm -rf lib/* && lerna clean --yes",
-    "lint": "lerna exec -- npm run lint -s",
-    "build": "lerna exec -- npm run build -s",
-    "test": "lerna exec -- npm run test -s",
-    "coverage": "lerna exec -- npm run coverage -s",
-    "codecov": "lerna exec -- npm run codecov -s",
-    "commit": "git-cz || true",
-    "check-updates": "run-p -s check-updates:*",
-    "check-updates:root": "ncu --error-level 2",
-    "check-updates:packages": "lerna exec -- ncu --error-level 2",
```
Siilwyn: Why remove this granular control of where to check for updates?
And below what to upgrade.

packages/commercetools-sdk-middleware-auth/README.md
Siilwyn: I would prefer `@commercetools/sdk` since we are talking about a 'node package manager' package.

package.json
```diff
@@ -1,66 +1,79 @@
...
-    "upgrade": "run-p -s check-updates:*",
-    "upgrade:root": "ncu --loglevel minimal --upgrade",
-    "upgrade:packages": "lerna exec -- ncu --loglevel minimal --upgrade",
+    "postinstall": "check-node-version --package && lerna bootstrap",
+    "check-updates": "./scripts/check-updates",
+    "clean": "rm -rf coverage && ./scripts/run-task clean",
+    "build": "./scripts/run-task build",
```
Siilwyn: Maybe we can simplify this script by using `lerna exec --` for the looping?

packages/commercetools-sdk/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "main": "./lib/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/commercetools/nodejs.git"
+  },
+  "scripts": {
+    "task:clean": "rm -rf lib"
+  },
+  "devDependencies": {
+    "@commercetools/sdk-middleware-auth": "file:../commercetools-sdk-middleware-auth",
```
Siilwyn: `file:`?
Lerna bootstrap should link all packages as dependencies.

package.json
emmenko: Hmm, `learn clean` removes `node_modules` but this `clean` task removes other stuff (not `node_modules`)

package.json
emmenko: This won't work the way we want because we're not just looping.
The `./scripts/run-task` has specific logic (see https://github.com/commercetools/nodejs/pull/2/commits/e6310918887186a6a2e7a4ed0c1ee2d37f969f33)

package.json
emmenko: It's not removed, the check part has been moved to the `check-updates` script. It's also something you won't do that often and we should use something like `greenkeeper` for this sort of stuff.

This is what you get now anyway, which I think is more then enough:

```bash
$ npm run check-updates

> @ check-updates /Users/emmenko/dev/src/nodejs
> ./scripts/check-updates

Using /Users/emmenko/dev/src/nodejs/package.json
⸨░░░░░░░░░░░░░░░░░░⸩ ⠧ :
 eslint-config-commercetools  4.0.2  →  5.0.0
 eslint-plugin-jsx-a11y       2.2.3  →  3.0.1
Dependencies not up-to-date
Lerna v2.0.0-beta.30
Independent Versioning Mode
Using /Users/emmenko/dev/src/nodejs/packages/commercetools-sdk-middleware-auth/package.json
Using /Users/emmenko/dev/src/nodejs/packages/commercetools-sdk/package.json
Using /Users/emmenko/dev/src/nodejs/packages/commercetools-sdk-middleware-http/package.json
⸨░░░░░░░░░░░░░░░░░░⸩ ⠴ :
All dependencies match the latest package versions :)

⸨░░░░░░░░░░░░░░░░░░⸩ ⠧ :
All dependencies match the latest package versions :)

⸨░░░░░░░░░░░░░░░░░░⸩ ⠦ :
All dependencies match the latest package versions :)

Waiting for 1 child process to exit. CTRL-C to exit immediately.
```

The upgrade part I don't find it necessary and can be dangerous if done "automatically" without thinking much what to upgrade.

packages/commercetools-sdk-middleware-auth/README.md
emmenko: Yep, like @Siilwyn said. The "node" part is redundant since it's a node module per-se ;)

packages/commercetools-sdk-middleware-auth/README.md
emmenko: ...but very informative ;)

packages/commercetools-sdk/package.json
emmenko: Yeah I know, something is still not working properly though. I still have to fix this.

scripts/check-npm.sh
emmenko: We're just going with node >= 6, no more workarounds ;)

emmenko:
> Can we use #!/usr/bin/env bash in the script shebangs and also add the .sh file extension (.js for the JS scripts ofc.)?

Personally I prefer without extension, it's more clear it's a script file. Also those scripts are just used for development. I don't have a strong opinion on this though. What would be your argument for having the file extensions?

packages/commercetools-sdk-middleware-http/package.json
hisabimbola: The folder names: `commercetools-sdk-middleware-http`

hisabimbola: and yeah, packages name.

scripts/check-npm.sh
Siilwyn: We still want to support Node 4, so Travis should test against 4 right?

package.json
Siilwyn: What I tried to say is that the control to run them individually is removed, not that it's completely removed, but okay. As for upgrading if we remove that we'd need to do it manually or what is the alternative you propose? I was thinking it's a nice script to ease keeping everything up-to-date (so not something CI would run but for all maintainers to use).

Siilwyn:
Well the extension would make it more clear it's a script file, but since one of them is a JS file maybe not. Though they're all located in the 'scripts' directory. So no strong opinion on it either, I guess it's a bit of personal preference, I like always having file extensions so that the OS can easily detect what kind of file it is.

That said the shebang is way more important to do right.

emmenko:
> yarn is missing as dev dependency

The recommended way of installing `yarn` is via `brew`
https://yarnpkg.com/en/docs/install#mac-tab

(see also "Install via npm" section in the alternative installations page https://yarnpkg.com/en/docs/install#alternatives-tab)

scripts/check-npm.sh
emmenko: Well, we want to support LTS version right? The current one is `6`. I guess we can also support node `4`. Using node `6` by default comes with `npm@3` included, although if we keep using `yarn` as dependency manager it doesn't really matter.

So we can support node `4` as well, I don't see a problem for now ;)

emmenko: Here you go https://github.com/commercetools/nodejs/commit/df446fc597181b37d3b254eeb1298b773eb71d7f

emmenko:
Alright guys, I'm going to merge this as it's more then good enough as a basis. There are obviously things to improve here and there to make it a smooth development process but overall I'm confident that this is going to work out!
I'll continue working on the sdk packages for now.

PR #21, opened 2017-01-18
chore(package): update cz-lerna-changelog to version 0.3.2

Siilwyn:
Version 0.3.2 of `cz-lerna-changelog` fixes compatiblity with `lerna beta.30`.

#### Todo

- Tests
    - [ ] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "codecov": "1.0.1",
     "commitizen": "2.9.5",
     "cross-env": "3.1.4",
-    "cz-lerna-changelog": "1.0.0",
+    "cz-lerna-changelog": "0.3.2",
     "eslint": "3.12.2",
     "eslint-config-airbnb": "13.0.0",
     "eslint-config-commercetools": "5.0.0",


package.json
```diff
@@ -43,7 +43,7 @@
     "codecov": "1.0.1",
     "commitizen": "2.9.5",
     "cross-env": "3.1.4",
-    "cz-lerna-changelog": "1.0.0",
+    "cz-lerna-changelog": "0.3.2",
```
emmenko: Wait, I'm confused here...

Siilwyn: Haha yeah, sorry forgot to explain why. There is no version `1.0.0` the latest is `0.3.2`.

emmenko: Hmm that's weird. They did a wrong release maybe?

image

> Btw. why are we not using the carets?

For dev deps we might, but we shouldn't use that for prod deps. Talking for experience seeing how many "patch", "minor" changes simply break stuff. Also, since we are using `yarn`,  we have a lockfile which locks the deps version.

> Having exact versions like this will require a lot of maintenance.

Not really, using something like e.g. greenkeeper does the maintainance for us.

Siilwyn: I mean that is still a lot of GreenKeeper PRs. ^^ And having the yarn lockfile would actually make using carets less of an issue than vice versa. I get your point though. Will move this discussion regarding the devDependencies to a separate issue.
乁། ˵ ◕ – ◕ ˵ །ㄏ
Either way it looks like they messed up the v1 release, can you approve this PR?

emmenko: FYI: we cannot upgrade lerna yet, until `lerna-semantic-release` is published. They merged my PR https://github.com/atlassian/lerna-semantic-release/pull/42 but didn't release it yet 🙈

PR #22, opened 2017-01-18
chore(package): add command to commit following the commit template

Siilwyn:
#### Todo

- Tests
    - [ ] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "postinstall": "check-node-version --package && lerna bootstrap && npm run build",
     "check-updates": "./scripts/check-updates",
     "clean": "rm -rf coverage; lerna exec -- rimraf lib dist",
+    "commit": "git-cz",
     "build": "lerna exec -- babel src --out-dir lib",
     "dist": "lerna run dist",
     "lint": "flow check && ./scripts/lint packages",


emmenko:
I think you can simply include this in #23

Siilwyn:
@emmenko I split them up in two PRs because we can only do 'Squash and merge' for PRs. This would mean that the commit message is not correct because it has two tags with different scopes: `docs` and `chore`. It was more hassle than needed though maybe we can also enable 'Rebase and merge'.

emmenko:
If that's the main reason it's not really a problem. When you squash usually it lists all the commits in the commit description anyway. It's up to you then to keep it or not

https://i.stack.imgur.com/cyHZY.png

emmenko:
And btw, in such cases where running a CI build is not necessary, we should use `[skip ci]` in the commit message.

/cc @hisabimbola

Siilwyn:
I don't understand, "When you squash usually it lists all the commits in the commit description anyway". That still doesn't solve the wrong commit message right?

hisabimbola:
I totally agree to the `[skip ci]` flag, my only concern is making sure it's not _abused_

We probably need to state the usage of the `[skip ci]` in the contributing.md

hisabimbola:
and to the other discussion @Siilwyn I really think you don't need to have two commits, all can be squash into one if they are tightly related.

It's just like saying, if you implement a feature, you create a PR for the feature and another PR for test and another PR for docs, and several other PRs for minor changes...😱

emmenko:
Yep, if there are related things they can be squashed together. Obviously you need to provide a more generic message for the all PR.

At the end all that matters is:
- `fix` (patch)
- `feat` (minor)
- `BREAKING CHANGE` (major)

So even if the PR contains several "fix" commits, some "chore" commits and so on, when you squash it / merge it, it only matters what version you want to be released

Siilwyn:
> it only matters what version you want to be released

In the future I will just put closely related things in one PR and squash but I really do not agree with this. A commit should tell about what changes are made and the message is part of that. Degrading it to just a triggering mechanism for version releases doesn't do justice.

Sidenote: shouldn't the owner of the PR merge?

emmenko:
Yeah, like I said in our last meeting, if we want to keep using semantic-release, this is the best way I found to be safe and working. It's really easy to "forget" defining the commit properly and miss the release.
Additionally, the single commits are still linked in the PR, it's just that you "merge" only 1 general commit. It keeps the history a bit cleaner also. Many repository do that for open source stuff.

From my side, this is something we still need to define and decide by "making our hands dirty" first, meaning we should try it out, understand pros and cons, and see first hand what at the end makes more sense.

> shouldn't the owner of the PR merge?

Yep sorry, I was just merging a bunch of PRs since they were all reviewed. But you're right ;)

PR #31, opened 2017-01-19
build(travis): fail fast and avoid running it tests for PRs coming from forks

### LINKED ISSUE #30: Disable integration tests for PRs coming from forks
We started this discussion some time ago. This is the problem that PRs from external contributors don't have access to encrypted variables in travis, for security reasons.
I think for now we can simply disable e.g. integration tests when travis builds such PRs.

@commercetools/node-js wdyt?

emmenko:
#### Summary
- builds now will fail fast if an error occurs `set -e`
- don't run *it* tests if PR is coming from a fork (Closes #30)

### DIFF
--- a/.travis/scripts.sh
+++ b/.travis/scripts.sh
@@ -13,6 +13,8 @@ npm run coverage:ci
 # thus tasks that rely on that should not run.
 if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ]; then
   npm run it
+else
+  echo "Build does not have secure env variables, skipping integration tests!"
 fi
 
 # 4. run dist tasks where necessary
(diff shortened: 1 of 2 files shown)


PhilippSpo:
> `Note: ensure to have `set -e` in order to fail the build fast.`
@emmenko where is `set -e` actually being set?

emmenko:
@PhilippSpo `#!/bin/bash -e`

.travis/scripts.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
```
Siilwyn: Let's use a proper bash shebang here for portability: `#!/usr/bin/env bash`. Then on the next line `set -e`.

emmenko: How is that different? This script also will be run only from travis, I don't see a problem...or am I missing something?

Siilwyn: Well since this only runs on Travis it's no biggie. The moment somebody wants to run this script locally it is because the location of `bash` is dependant on the OS. Here's a good [stackoverflow answer](http://stackoverflow.com/a/10383546).

emmenko: Sure, feel free to change it if you want ;)

PR #33, opened 2017-01-19
feat(api-request-builder): add missing services to cover all API endpoints

emmenko:
#### Summary
Define remaining services for API endpoints.

Note: there are a few extra things that should also be supported but they can be done separately.

#### Todo
- Tests
    - [x] Unit

### DIFF
--- a/packages/commercetools-api-request-builder/src/create-service.js
+++ b/packages/commercetools-api-request-builder/src/create-service.js
@@ -92,6 +92,7 @@ export default function createService (
     // Pass some options to further configure the URI:
     // - `projectKey`: will prefix the URI with the given projectKey
     build (options: BuildOptions = {}): string {
+      // FIXME: make it an option of `createRequestBuilder`
       const { projectKey } = options
 
       const queryParams = buildQueryString(this.params)
@@ -99,11 +100,19 @@ export default function createService (
       const uri =
         (projectKey ? `/${projectKey}` : '') +
         endpoint +
-        (this.params.id ? `/${this.params.id}` : '') +
+        getIdOrKey(this.params) +
         (queryParams ? `?${queryParams}` : '')
 
       setDefaultParams.call(this)
       return uri
     },
   })
+}
+
+function getIdOrKey (params: Object): string {
+  if (params.id)
+    return `/${params.id}`
+  else if (params.key)
+    return `/key=${params.key}`
+  return ''
 }
--- a/packages/commercetools-api-request-builder/src/create-service.js
+++ b/packages/commercetools-api-request-builder/src/create-service.js
@@ -92,7 +92,6 @@ export default function createService (
     // Pass some options to further configure the URI:
     // - `projectKey`: will prefix the URI with the given projectKey
     build (options: BuildOptions = {}): string {
-      // FIXME: make it an option of `createRequestBuilder`
       const { projectKey } = options
 
       const queryParams = buildQueryString(this.params)
--- a/packages/commercetools-api-request-builder/src/default-services.js
+++ b/packages/commercetools-api-request-builder/src/default-services.js
@@ -155,7 +155,6 @@ export default {
       features.projection,
     ],
   },
-  // TODO: productSuggestions
   products: {
     type: 'products',
     endpoint: '/products',
--- a/packages/commercetools-api-request-builder/src/default-services.js
+++ b/packages/commercetools-api-request-builder/src/default-services.js
@@ -1,6 +1,24 @@
 import * as features from './features'
 
 export default {
+  cartDiscounts: {
+    type: 'cart-discounts',
+    endpoint: '/cart-discounts',
+    features: [
+      features.query,
+      features.queryOne,
+      features.queryExpand,
+    ],
+  },
+  carts: {
+    type: 'carts',
+    endpoint: '/carts',
+    features: [
+      features.query,
+      features.queryOne,
+      features.queryExpand,
+    ],
+  },
   categories: {
     type: 'categories',
     endpoint: '/categories',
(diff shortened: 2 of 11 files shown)


nkuehn:
Wonderful!   

From a "build a shopfront" use case you probably also want to be able to access the project metadata under  the root url an want to trigger the stuff that's a path below the customer like http://dev.commercetools.com/http-api-projects-customers.html#change-customers-password  and the password reset token.  Not a high priority though.

packages/commercetools-api-request-builder/src/create-service.js
```diff
@@ -92,18 +92,27 @@ export default function createService (
     // Pass some options to further configure the URI:
     // - `projectKey`: will prefix the URI with the given projectKey
     build (options: BuildOptions = {}): string {
+      // FIXME: make it an option of `createRequestBuilder`
```
Siilwyn: Should resolve this before merging or remove it & open an issue about it so we can track it there.

packages/commercetools-api-request-builder/src/create-service.js
```diff
@@ -92,18 +92,27 @@ export default function createService (
...
+        getIdOrKey(this.params) +
         (queryParams ? `?${queryParams}` : '')
 
       setDefaultParams.call(this)
       return uri
     },
   })
 }
+
+function getIdOrKey (params: Object): string {
```
Siilwyn: This looks odd, why not two functions and put the `if - else` logic where needed?

packages/commercetools-api-request-builder/src/default-services.js
```diff
@@ -57,6 +155,7 @@ export default {
       features.projection,
     ],
   },
+  // TODO: productSuggestions
```
Siilwyn: Would also put this is an issue.

packages/commercetools-api-request-builder/src/query-id.js
```diff
@@ -6,11 +6,35 @@
  * @throws If `id` is missing.
  * @return {Object} The instance of the service, can be chained.
  */
-// eslint-disable-next-line import/prefer-default-export
```
Siilwyn: ⬆️

packages/commercetools-api-request-builder/src/query-projection.js
```diff
@@ -11,3 +11,9 @@ export function staged (isStaged: boolean = true): Object {
   this.params.staged = isStaged
   return this
 }
+
+// TODO: add price selection query params
```
Siilwyn: Convert to issue.

packages/commercetools-api-request-builder/src/create-service.js
emmenko: We should create an issue yes, but this should be done separately and it's a breaking change

packages/commercetools-api-request-builder/src/create-service.js
emmenko: ??

packages/commercetools-api-request-builder/src/default-services.js
emmenko: I've added it in the project backlog, which can be converted into an issue

https://github.com/commercetools/nodejs/projects/1

packages/commercetools-api-request-builder/src/query-projection.js
emmenko: Also in the backlog ;)

https://github.com/commercetools/nodejs/projects/1

packages/commercetools-api-request-builder/src/create-service.js
Siilwyn: Oh nvm, this function confused me 😅 .

emmenko:
@nkuehn 

> From a "build a shopfront" use case you probably also want to be able to access the project metadata under the root url an want to trigger the stuff that's a path below the customer like http://dev.commercetools.com/http-api-projects-customers.html#change-customers-password and the password reset token. Not a high priority though.

Yeah, although this package mostly cares about building the query parameters. If you just have an endpoint like `/customers/password` without any query parameters you might as well just type it yourself ;)
But I still haven't fully decided how to deal with such simple (special) cases

packages/commercetools-api-request-builder/src/create-service.js
emmenko: 😄

packages/commercetools-api-request-builder/src/query-id.js
hisabimbola: Not needed anymore I guess.

packages/commercetools-api-request-builder/src/create-service.js
```diff
@@ -92,18 +92,27 @@ export default function createService (
     // Pass some options to further configure the URI:
     // - `projectKey`: will prefix the URI with the given projectKey
     build (options: BuildOptions = {}): string {
```
hisabimbola: @Siilwyn doesn't this have something to do with the error you were having this evening?

packages/commercetools-api-request-builder/src/create-service.js
```diff
@@ -92,18 +92,27 @@ export default function createService (
     // Pass some options to further configure the URI:
     // - `projectKey`: will prefix the URI with the given projectKey
```
hisabimbola: @Siilwyn doesn't this have something to do with the error you were having this evening?

packages/commercetools-api-request-builder/src/create-service.js
hisabimbola: So it's going to be remove 😑 (and probably a better context)

packages/commercetools-api-request-builder/src/create-service.js
hisabimbola: You will get use to **type** function soon @Siilwyn 😜

packages/commercetools-api-request-builder/src/create-service.js
Siilwyn: :o it is! That fixed it, thank you! :3

emmenko:
Alright, TODOs / FIXMEs have been converted to issues.

PR #37, opened 2017-01-20
fix(middleware-http): set options parameter to an object

Siilwyn:
affects: @commercetools/sdk-middleware-http

Set the `options` parameter of the function to an object by default so that passing nothing will
work.

#### Summary
I think the `options` parameter is optional? It's not documented so we might also want to indicate it's optional, any ideas for that? I've seen putting the parameter in _italic style_ but not sure if that's explicit enough.


Siilwyn:
Looks like `flow` is complaining, but I don't know why. 🐧  Any help would be appreciated!

packages/commercetools-sdk-middleware-http/src/http.js
```diff
@@ -18,15 +18,13 @@ import getErrorByCode, {
...
 export default function createHttpMiddleware (
-  options: HttpMiddlewareOptions,
+  options: HttpMiddlewareOptions = {
+    host: defaultApiHost
+  },
 ): Middleware {
-  const {
-    host = defaultApiHost,
-    // TODO: other options?
-  } = options
```
emmenko: Just do `} = (options || {})`

Siilwyn:
@emmenko I actually wrote it with ES6's default parameter values on purpose, looks a bit nicer and one less declaration needed.

emmenko:
> I actually wrote it with ES6's default parameter values on purpose, looks a bit nicer and one less declaration needed.

I know, it's just because it's a bit tricky to make it work with flow, that's all. I can give it another try if you like.

> Regarding the documentation I mean it's not written if the parameters are optional or not, right? (So I mean the gitbooks page)

Ah yeah, if you have suggestions to make it clearer please do ;)

Siilwyn:
I will take a look at it, currently looking more into how Flow works so this seems like a good task.

types/sdk.js
```diff
@@ -60,7 +60,7 @@ export type AuthMiddlewareOptions = {
   scopes: Array<string>;
 }
 export type HttpMiddlewareOptions = {
-  host?: string;
+  host: string;
```
emmenko: This should stay optional

Siilwyn: Well as far as I could see not since we're always assigning a value to it, otherwise Flow will complain.

emmenko: Sure. The only thing is that this type can not really be "used" for documentation, since it says it's a required field.

Siilwyn: Are the docs generated based on the type definition?

emmenko: They are not used to generate the doc, but it's good for knowing the object shape when coding or whatever. The default value that we provide should not affect the type definition. If there is no other way I would prefer to do it like I did, so that the type definition is correct.

Siilwyn: Playing the "devil's advocate" here: what type definition is 'correct'. If looking it from the other way putting a Maybe type on it is not correct since we set it to a String if nothing is passed.

emmenko: Alright, let's do it this way then. Can you change it also for the other middleware option types?

Siilwyn: Alright, feels like outside of the scope of this PR but will do.
Also found this: "That is, Flow understands that providing a default value makes the parameter optional. You don't need to explicitly make the parameter type optional—although it doesn't hurt."
- http://stackoverflow.com/questions/32893892/what-is-the-syntax-for-defining-a-type-when-parameter-has-a-default-value

Siilwyn:
@emmenko :D I'm glad you asked to change the other middlewares too because I realised it's broken this way when having more than one parameter.

PR #53, opened 2017-01-27
fix(middleware-auth): fix usage of cached token, no default options

emmenko:
BREAKING CHANGE: `host` is now a required option, no more default value

#### Summary
Yesterday I realized that the usage of the token cache was not properly done (new token would be retrieved for each new request).

Additionally, after some conversation yesterday, we should not put any _default_ `host` values, as our platform as already multiple data centers. It also keeps the middleware more decoupled from our platform. 

#### Todo
- Tests
    - [x] Unit
    - [ ] Integration
- [ ] Documentation

### DIFF
--- a/packages/commercetools-sdk-middleware-auth/src/build-requests.js
+++ b/packages/commercetools-sdk-middleware-auth/src/build-requests.js
@@ -12,34 +12,34 @@ type BuiltRequestParams = {
   body: string;
 }
 
-const defaultAuthHost = 'https://auth.sphere.io'
-
 // POST https://{host}/oauth/token?grant_type=client_credentials&scope={scope}
 // Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==
 export function buildRequestForClientCredentialsFlow (
-  {
-    host = defaultAuthHost,
-    projectKey,
-    credentials,
-    scopes,
-  }: AuthMiddlewareOptions = {},
+  options: AuthMiddlewareOptions,
 ): BuiltRequestParams {
-  if (!projectKey)
+  // TODO: use a better validator
+  if (!options)
+    throw new Error('Missing required options')
+
+  if (!options.host)
+    throw new Error('Missing required option (host)')
+
+  if (!options.projectKey)
     throw new Error('Missing required option (projectKey)')
 
-  if (!credentials)
+  if (!options.credentials)
     throw new Error('Missing required option (credentials)')
 
-  const { clientId, clientSecret } = credentials
+  const { clientId, clientSecret } = options.credentials
 
   if (!(clientId && clientSecret))
     throw new Error('Missing required credentials (clientId, clientSecret)')
 
-  const defaultScope = `${authScopes.MANAGE_PROJECT}:${projectKey}`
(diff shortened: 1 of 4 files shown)


packages/commercetools-sdk-middleware-auth/src/build-requests.js
```diff
@@ -12,34 +12,34 @@ type BuiltRequestParams = {
...
-  {
-    host = defaultAuthHost,
-    projectKey,
-    credentials,
-    scopes,
-  }: AuthMiddlewareOptions = {},
+  options: AuthMiddlewareOptions,
 ): BuiltRequestParams {
-  if (!projectKey)
+  // TODO: use a better validator
```
emmenko: This is something we might consider in general, whether to do validation manually (like we do now) or to have some sort of schema that does validation automatically.

Siilwyn: We can put the required arguments in an array, loop and return an error if one of them misses.

Siilwyn:
I don't think it's nice to have to specify the API host value. The cache fix is very nice though.

emmenko:
> I don't think it's nice to have to specify the API host value

This is something that @nkuehn also brought up, that our US folks had some problems because the host was hardcoded to the EU one (not specific to this SDK, but in general). Since we don't have a single "main" host anymore in our API it doesn't make sense to have one. Additionally, by not having one, it keeps the middleware decoupled from **our** API.

packages/commercetools-sdk-middleware-auth/src/build-requests.js
emmenko: Yes and no :) For example, the `credentials` is an object, and we need to validate that both keys in it are provided.

My question was mostly if we do this manually (array or not array) or have some kind of schema validator.

hisabimbola: How were you able to a _TODO_ comment though, I thought our eslint config forbids that...🤔

Nevermind, I just saw the _:_ that makes the rule useless 🙄

hisabimbola: and yeah, we can have a custom validator

Siilwyn:
What if you can pass 'EU' and 'US' to simplify the usage of the SDK?

nkuehn:
@Siilwyn  a) there will be more locations b) clients wrap our API behind API gateways etc.  c) there are datacenter-internal API URLs that clients can use to save traffic cost d) etc. pp.  

Short version: no, it's good that people need to understand and explicitly what actual API endpoint they are working against.

Siilwyn:
@nkuehn thank you for the extra info, didn't know it was used this way.

packages/commercetools-sdk-middleware-auth/src/build-requests.js
Siilwyn: ESLint rule dodging. :P

Siilwyn:
Docs on https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html are out of date? Host has no default value anymore.

PR #56, opened 2017-01-28
refactor(middleware-http): make host option required

emmenko:
BREAKING CHANGE: `host` is now a required option, no more default value

Follow discussion in #53


Siilwyn:
https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareHttp.html states there is a default value still.

PR #58, opened 2017-01-28
Safely check for browser env when accessing global.window

### LINKED ISSUE #46: sdk-middleware-auth is broken in Node.js environment
### Description
Using the auth middleware does not work in Node.js [because of this piece of code](https://github.com/commercetools/nodejs/blob/master/packages/commercetools-sdk-middleware-auth/src/build-requests.js#L62-L76):
```js
export function getBasicAuth (
  username: string,
  password: string,
  windowObject: Object = window,
): string {
  const basicAuth = `${username}:${password}`
  if (
    windowObject &&
    windowObject.btoa &&
    typeof windowObject.btoa === 'function'
  )
    return windowObject.btoa(basicAuth)

  return new Buffer(basicAuth).toString('base64')
}
```
So `windowObject` falls back to `window` which will be undefined.

### Possible Solution
I propose to follow this approach which Abi found online:
https://nolanlawson.com/2017/01/09/how-to-write-a-javascript-package-for-both-node-and-the-browser/
Funny enough the author created a package for this which we can use: `base64-encode-string`.

emmenko:
Fixes #46

#### Summary
This PR fixes the problem in `node` environment caused by `global.window`.

#### Description
For `auth`, since we were using `Buffer` anyway, we can simply use the polyfill provided by webpack.
For `user-agent` it's fixed (see comment) in a different way, but allows to easily test it.

Additionally, I realised that running the tests scoped by package didn't use the jest config defined in `package.json` (because of the directory context), so I've put the config into a `test-config.json` file and each script running `jest` will reference to the file config.
And tests run in `node` environment now by default (jest uses `jsdom` by default).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,13 +13,13 @@
     "build": "lerna exec -- babel src --out-dir lib",
     "dist": "lerna run dist",
     "lint": "flow check && ./scripts/lint packages",
-    "test": "npm run build && lerna exec -- jest",
+    "test": "npm run build && jest --config test-config.json",
     "test:package": "npm run build && ./scripts/test-package",
     "it": "npm run it:install && npm run it:run",
     "it:run": "jest --config integration-tests/config.json",
     "it:install": "npm run build && ./scripts/install-npm-packages-for-integration-tests",
-    "coverage": "npm run build && jest --no-watchman --coverage",
-    "coverage:ci": "jest --no-watchman --coverage && codecov",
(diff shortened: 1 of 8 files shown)


emmenko:
So, since this is an important fix as the current versions are broken, I'm going to release it. Any further discussions about possible changes about the implementation can be done later.
Thanks.

PR #64, opened 2017-01-31
docs(readme): add list of package names, versions and dependency status

### LINKED ISSUE #28: List packages in readme
List the packages with their link to the NPM page, version and dependency status in the readme.

Siilwyn:
Resolves #28


Siilwyn:
Yeah that would be great but how do we differentiate between the different packages? E.g. once we have more packages we would maybe have some groups as: SDK, middlewares, CSV parsers, etc.
If it's all one big table it would be easy. Ideas on that?

emmenko:
On the other hand, you might only need that when adding a new package. So maybe the script can simply print the information for 1 single package. Then you put it in the README wherever you want.
Not really a big issue anyway, more of a nice to have suggestion ;)

Siilwyn:
@emmenko made a small script, indeed we only want to use it when adding a new package. Also looking at Babel they don't list all packages since some are never used directly by users so listing everything automatically is not ideal unless we filter out packages somehow.

hisabimbola:
> Also looking at Babel they don't list all packages since some are never used directly by users
I am not sure how babel does it, but if it's a package that is published to NPM then we automatically assume that it being used. So maybe a good filter will be to check if the 'private' flag is false in the package.json

Siilwyn:
@hisabimbola  Well if we follow Babel's way that wouldn't work. E.g. `babel-helper-regex` is published to npm but not listed in the readme.

Siilwyn:
Abi noticed that the badges have a lot of timeouts it might be worth setting up a simple server that acts as a proxy with a cache. Maybe it's something I can do tomorrow with the tech time.

PR #74, opened 2017-02-07
fix(http-middleware): fix host url. works with or without ending slash

### LINKED ISSUE #67: SDK middleware fails when given a host that ends with a slash
### Description
Passing `https://api.sphere.io/` vs. `https://api.sphere.io` makes a difference.

### Expected Behavior
Both should work the same.

### Current Behavior
My guess is it ends up with a double slash when appending things to the host URL.

### Context
I used the URLs that are given at the dev.commercetools.com documentation which failed. Confirmed this for `sdk-middleware-http`.

### Possible Solution
Merge slashes if needed.

wizzy25:
affects: @commercetools/sdk-middleware-http

fixes #67

#### Summary

resolve host url. works with or without ending slash

#### Description

 host url gets parsed properly whether
 it ends with slash or not

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


packages/sdk-middleware-auth/src/client-credentials-flow.js
```diff
@@ -20,8 +20,7 @@ type Task = {
 }
 
 export default function createAuthMiddlewareForClientCredentialsFlow (
-  options: AuthMiddlewareOptions,
-): Middleware {
```
emmenko: Keep the same formatting

packages/sdk-middleware-auth/test/client-crendentials-flow.spec.js
```diff
@@ -97,6 +97,7 @@ describe('Client Crentials Flow', () => {
       )
       nock(middlewareOptions.host)
         .filteringRequestBody(/.*/, '*')
+        // Check hear
```
emmenko: ??

package.json
```diff
@@ -120,5 +120,8 @@
         "babel-module": {}
       }
     }
+  },
+  "dependencies": {
+    "normalize-url": "^1.9.0"
```
emmenko: Let's not introduce unnecessary dependencies please.

You can simply fix this "problem" with a `replace(/\/$/, '')`

```js
'foo/'.replace(/\/$/, '')
// "foo"
'foo'.replace(/\/$/, '')
// "foo"
```

packages/sdk-middleware-http/package.json
```diff
@@ -31,7 +31,8 @@
     "build:umd:min": "cross-env NODE_ENV=production webpack --config ../../webpack.config.js --output-library CommercetoolsSdkMiddlewareHttp --output-library-target umd src/index.js dist/commercetools-sdk-middleware-http.min.js"
   },
   "dependencies": {
-    "isomorphic-fetch": "2.2.1"
+    "isomorphic-fetch": "2.2.1",
+    "normalize-url": "^1.9.0"
```
emmenko: Remove

packages/sdk-middleware-http/src/http.js
```diff
@@ -19,7 +21,7 @@ export default function createHttpMiddleware (
   options: HttpMiddlewareOptions,
 ): Middleware {
   return next => (request: MiddlewareRequest, response: MiddlewareResponse) => {
-    const url = options.host + request.uri
+    const url = normalizeUrl(options.host + request.uri)
```
emmenko: See comment https://github.com/commercetools/nodejs/pull/74/files#r99773653

package.json
wizzy25: This would assume that the uri always begins with a slash, which may not always be the case

emmenko: We should make it a rule then (e.g. validation)

package.json
```diff
@@ -121,5 +121,6 @@
         "babel-module": {}
       }
     }
-  }
+  },
```
hisabimbola: You can remove the `dependencies` field since it's empty. No need to commit it.

packages/sdk-middleware-auth/test/build-requests.spec.js
```diff
@@ -31,6 +31,28 @@ describe('buildRequestForClientCredentialsFlow', () => {
     })
   })
 
+  it('parses a host that ends with \'/\'', () => {
```
hisabimbola: To make this more readable I would prefer you use backtick _`_ and quotes to emphasise the forward slash

wizzy25:
I think it should be good enough for a merge now?

emmenko:
BTW: there are eslint errors, but you shouldn't have been able to commit. How is that possible?

packages/sdk-middleware-auth/test/build-requests.spec.js
wizzy25: @emmenko The eslint errors are because I replaced ```'``` with ``` ` ``` to avoid escaping and make the code more reasonable (I forced the commit)
Would replace ```/``` with ```"slash"``` to maintain readability and lintability... Then everyone should be happy 😋

emmenko: I mean, you shouldn't have been able to commit since we have git hooks setup. Did you actually set them up correctly?

wizzy25: I forced the commit with ```--no-verify``` flag

emmenko: :man_facepalming: what's the point then of having the commit hooks? 😅

Try to avoid then, unless it's really necessary please.

wizzy25: 😃😃🙏🏾

emmenko:
Alright, let's go ahead and merge this. Good job guys! 👍

hisabimbola:
@wizzy25 Please squash and merge and remember to use `fix` so that a release can be triggered.

emmenko:
And `affects`. Read https://github.com/commercetools/nodejs/blob/master/CONTRIBUTING.md#releases

Siilwyn:
Or just do `git reset --soft HEAD~9 && npm run commit` ;)
No way for human errors that way.

hisabimbola:
@wizzy25 Please delete the branch when you done merging to avoid old branches flying 🚡  around

emmenko:
The release commit was wrong, there should be only one "affects" line with comma separated list of packages. Now only http got release.
This is one of the reasons I have doubts about using semantic releases.

Anyway, no worries and just keep in mind this for the future ;)

PR #87, opened 2017-02-16
feat(packages): add new package get-credentials

### LINKED ISSUE #50: Add API credentials retrieving package
### Description
Have a package that retrieves the API credentials from a system like the one we have in the old utils repository. Trying to simplify the behaviour of getting the credentials I came up with three steps. The package will return a promise since it could need to read from the FS.

### Expected Behavior
1. Project name is passed, environment variable: `CT_<project key>` containing `<client id>:<client secret>`.
_Managing multiple projects having to only pass the name._
2. Check environment variables `CT_PROJECT_KEY`, `CT_CLIENT_ID`, `CT_CLIENT_SECRET`.
_For single project usage, nice to have so no project name needs to be passed._
3. Use [dotenv](https://github.com/motdotla/dotenv) package loading values from a file `ct-credentials.env` from the current or `etc` directory.
_Same as the second step but nice to have for users that are not comfortable with setting environment variables._

### Context
This is needed to move on with #27 and good to have for future packages. Avoiding logic duplication like over multiple packages like in [csv-parser-price](https://github.com/commercetools/csv-parser-price/blob/master/src/get-api-credentials.js).

Thoughts?

Siilwyn:
affects: @commercetools/get-credentials

Add new package to retrieve credentials from the system. Credentials can be given with a 'dotenv'
file or by using environment variables.

Resolves #50

One unusual thing is that `get-credentials` rejects an array of errors. This is so that modules that are using `get-credentials` can display the user the steps it went through. For example reading the different 'dotenv' files before getting environment variable.


Siilwyn:
The linter is complaining:
`sinon should be listed in the project's dependencies, not devDependencies.  import/no-extraneous-dependencies`

Moving it to the root package.json file does not resolve this. :/ @emmenko ideas?

packages/get-credentials/src/get-credentials.js
```diff
@@ -0,0 +1,60 @@
...
+export default function getCredentials (projectKey) {
+  if (!projectKey)
+    return Promise.reject([new Error('Project Key is needed')])
+
+  return Promise.resolve(setCredentialsFromEnvFile())
+    .catch(Promise.resolve)
+    .then(dotenvError => getCredentialsFromEnvironment(projectKey)
+      .catch(environmentError =>
+        Promise.reject([dotenvError, environmentError]),
+      ))
```
emmenko: I don't understand this all catch/then/catch/reject thingy. Can you explain what's going on here? 🤔

packages/get-credentials/src/index.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as getCredentials } from './get-credentials'
+export { getCredentialsFromEnvironment } from './get-credentials'
+export { setCredentialsFromEnvFile } from './get-credentials'
```
emmenko: Do we need to export also the named exports?

packages/get-credentials/src/index.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as getCredentials } from './get-credentials'
```
emmenko: Nitpick: you can simply do `export { default } from '...'`

packages/get-credentials/src/get-credentials.js
```diff
@@ -0,0 +1,60 @@
+import dotenv from 'dotenv'
+import path from 'path'
+
+import { homepage } from '../package.json'
+
+export default function getCredentials (projectKey) {
+  if (!projectKey)
+    return Promise.reject([new Error('Project Key is needed')])
```
emmenko: Nitpick: I would rewrite it to `Missing "projectKey"`

packages/get-credentials/test/get-credentials.spec.js
```diff
@@ -0,0 +1,104 @@
...
+      })
+  })
+
+  test('should reject an array of all errors', (done) => {
+    getCredentials('forgery')
+      .then(done.fail)
+      .catch((errors) => {
+        expect(errors.length).toBe(2)
+        expect(errors[0].message).toBeTruthy()
+        expect(errors[1].message).toBeTruthy()
```
emmenko: Why should we get 2 errors here?

packages/get-credentials/test/get-credentials.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  })
+})
+
+describe('getCredentialsFromEnvironment', () => {
+  const sandbox = sinon.sandbox.create()
+
+  afterEach(() => sandbox.restore())
+
+  test('should resolve credentials from environment variables', () => {
+    sandbox.stub(process, 'env', { CT_STROOPWAFEL: 'nyw:les' })
```
emmenko: You don't need `sinon`. You can simply do `process.env.CT_STROOPWAFEL='nyw:les'`, the real env var won't be overridden.

```bash
$ export FOO=bar
$ echo $FOO
bar
$ npm t

 PASS  packages/xxx.spec.js
 ...
Ran all test suites.
$ echo $FOO
bar
```

and the test was

```js
it('should mock env var', () => {
  process.env.FOO = '123'
  expect(process.env.FOO).toEqual('123')
})
```

packages/get-credentials/test/get-credentials.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  })
+})
+
+describe('setCredentialsFromEnvFile', () => {
+  const sandbox = sinon.sandbox.create()
+
+  afterEach(() => sandbox.restore())
+
+  test('should set environment variables', () => {
+    sandbox.stub(fs, 'readFileSync')
```
emmenko: This also you can do with jest mocks

```js
// on top of file, before the imports
let mockReadFileSync = jest.fn()
jest.mock('fs', () => ({
  readFileSync: (...args) => mockReadFileSync(...args),
}))

// then in your test
describe('setCredentialsFromEnvFile', () => {
  beforeEach(() => {
    mockReadFileSync.mockClear()
  })
  test('should set environment variables', () => {
    mockReadFileSync = jest.fn(() => `
      CT_STROOPWAFEL=dev:null
      CT_PANNENKOEK=dev:urandom
    `)
  })
})
```

packages/get-credentials/readme.md
```diff
@@ -0,0 +1,11 @@
+# get-credentials
+
+Retrieve commercetools platform credentials from environment variables or file system.
```
emmenko: `s/platform/project`

packages/get-credentials/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "license": "MIT",
+  "contributors": [
+    "Selwyn Versteeg <selwyn.versteeg@commercetools.com> (https://selwyn.cc/)"
+  ],
+  "main": "lib/index.js",
+  "files": [
+    "lib"
+  ],
+  "devDependencies": {
+    "sinon": "^1.17.7"
```
emmenko: We don't need `sinon` (see comments in test file)

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
+# `get-credentials`
+Retrieve commercetools platform credentials from environment variables or file system.
```
emmenko: `s/platform/project`

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
+# `get-credentials`
```
emmenko: Nitpick: wouldn't better `load-credentials`?

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Install
+
+```bash
+npm install --save @commercetools/get-credentials
+```
+
+## Usage
+
+First `get-credentials` will try to load the credentials from a file `ct-credentials.env` in the current directory and then from the `/etc` directory. If a project's credentials exist in both files the values from the current directory are used. Secondly credentials will be retrieved from an environment variable containing the project key: `CT_<project key>` containing `<client id>:<client secret>`. Notice that a project key will be uppercased and dashes will be replaced by an underscore. So for example: `my-projectkey` becomes `MY_PROJECTKEY`.
```
emmenko: This needs to be rephrased a bit 😁

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
...
+CT_OTHER_PROJECTKEY=myclientid:mysecret
+```
+
+## `getCredentials(projectKey)`
+
+Set environment variables from a [dotenv](https://github.com/motdotla/dotenv) file and load the credentials from an environment variable based on the passed project key.
+
+### Arguments
+
+1. `projectKey` *(String)*: The project key to get the client ID and client secret for.
```
emmenko: Nitpick: `The project key to retrieve the associated client credentials`

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
...
+
+Example `ct-credentials.env` file:
+```dosini
+CT_MY_PROJECTKEY=myclientid:mysecret
+CT_OTHER_PROJECTKEY=myclientid:mysecret
+```
+
+## `getCredentials(projectKey)`
+
+Set environment variables from a [dotenv](https://github.com/motdotla/dotenv) file and load the credentials from an environment variable based on the passed project key.
```
emmenko: What do you mean by `Set environment variables`?

docs/sdk/api/getCredentials.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Arguments
+
+1. `projectKey` *(String)*: The project key to get the client ID and client secret for.
+
+### Returns
+
+The function returns a promise which if the credentials are found resolves to an object like this:
```
emmenko: Simply say 

> Returns a `Promise` which gets resolved with `{ clientId, clientSecret }`

docs/sdk/api/getCredentials.md
Siilwyn: Not sure, since it only retrieves them and not actually 'load' them into something... Some other opinions would help.

docs/sdk/api/getCredentials.md
Siilwyn: So the `dotenv` package sets `process.env` variables from the file. But maybe I shouldn't mention it since it's something internal.

docs/sdk/api/getCredentials.md
Siilwyn: Merged the two sentences a bit.

docs/sdk/api/getCredentials.md
Siilwyn: Yes I'm not sure how though, maybe with numeric steps like:
1. ...
2. ...
?

packages/get-credentials/src/get-credentials.js
Siilwyn: Throwing it? The function should always return a promise right?

It's in an array because the function can return multiple errors, if it were to only sometimes return an array that would make it harder to use I think.

packages/get-credentials/src/get-credentials.js
Siilwyn: Yes since multiple things can go wrong but don't want to immediately stop and return an error when they do it rejects at the end 'collecting' the errors before that into an array.

packages/get-credentials/src/get-credentials.js
```diff
@@ -0,0 +1,60 @@
...
+    const [clientId, clientSecret] = envValue.split(':')
+
+    return resolve({
+      clientId,
+      clientSecret,
+    })
+  })
+}
+
+export function setCredentialsFromEnvFile () {
```
Siilwyn: Think this is pretty clear, no? Suggestions are welcome.

packages/get-credentials/src/index.js
Siilwyn: If I do that I get this error:
`TypeError: (0 , _index.getCredentials) is not a function`
😕

packages/get-credentials/src/index.js
Siilwyn: Maybe not `setCredentialsFromEnvFile` but `getCredentialsFromEnvironment` is usable for modules where you don't use env files. Think we can use `getCredentialsFromEnvironment` for the integration tests.

packages/get-credentials/test/get-credentials.spec.js
Siilwyn: Since both the `dotenv` fails and the retrieval from the environment variable fails. I've changed the assertions to be more descriptive.

packages/get-credentials/package.json
Siilwyn: The reason I went with a sandbox is that it doesn't influence the other tests. Doing it without a sandbox is a bit dangerous because one test can influence the result of others. Also relying on `process.env` without a sandbox means that developer's environment variables could collide with the tests.

packages/get-credentials/test/get-credentials.spec.js
Siilwyn: The reason I went with a sandbox is that it doesn't influence the other tests. Doing it without a sandbox is a bit dangerous because one test can influence the result of others. Also relying on `process.env` without a sandbox means that developer's environment variables could collide with the tests.

docs/sdk/api/getCredentials.md
hisabimbola: You can follow the format used [here](https://github.com/sphereio/sphere-node-utils#projectcredentialsconfig)

and also does this support _home~_ directory and what's the hierarchic?
does it support JSON also?

Siilwyn: Thanks, good idea. Will split it into two sections 'From a file' and 'From environment variables'.

Siilwyn: It doesn't support JSON and no home directory either. To keep things simple I want to minimize the amount of locations the config can be and have one format. So it looks in the working directory and in `/etc` which is an appropriate location for config files.

Siilwyn:
Removed the multiple error handling because I realised it introduces complexity without that much benefit. I'd like another review round.

packages/get-credentials/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "keywords": [
+    "commercetools",
+    "credentials"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/commercetools/nodejs/tree/master/packages/get-credentials"
+  },
+  "homepage": "https://commercetools.github.io/nodejs/sdk/api/getCredentials.html",
+  "bugs": "https://github.com/commercetools/get-credentials/issues",
```
Siilwyn: Fixed #1e31790

Siilwyn:
@hisabimbola fixed your suggestions, could you review again please?

packages/get-credentials/src/get-credentials.js
emmenko: 🤔 where exactly are the credentials being set?

Siilwyn: The `dotenv` package sets them.

Siilwyn:
@hisabimbola @emmenko  :eyes:

emmenko:
🎉

PR #99, opened 2017-02-27
feat(http): add server headers to response

### LINKED ISSUE #98: [sdk-middleware-http] Include headers in request response
### Description

The headers object is missing from the response.

### Expected Behavior

Should include the headers in the response
### Current Behavior

the response header is striped out.

hisabimbola:
#### Summary

Fixes #98 
#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [x] Documentation

### DIFF
--- a/packages/sdk-client/src/client.js
+++ b/packages/sdk-client/src/client.js
@@ -43,6 +43,7 @@ export default function createClient (options: ClientOptions): Client {
             rs.resolve({
               body: rs.body || {},
               statusCode: rs.statusCode,
+              headers: rs.headers || {},
             })
         }
 
--- a/packages/sdk-client/src/client.js
+++ b/packages/sdk-client/src/client.js
@@ -39,12 +39,17 @@ export default function createClient (options: ClientOptions): Client {
           // override those functions for custom behaviours.
           if (rs.error)
             rs.reject(rs.error)
-          else
-            rs.resolve({
+          else {
+            const resObj: Object = {
               body: rs.body || {},
               statusCode: rs.statusCode,
-              headers: rs.headers || {},
-            })
+            }
+            if (rs.headers)
+              resObj.headers = rs.headers
+            if (rs.request)
+              resObj.request = rs.request
+            rs.resolve(resObj)
+          }
         }
 
         const dispatch = compose(...options.middlewares)(resolver)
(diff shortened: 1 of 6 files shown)


hisabimbola:
I was going through the previous implementation [here](https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/services/base.coffee#L583) and I think we should probably implement this to look something like that

hisabimbola:
Hello, the build is failing because though _raw()_ method is present in the actual implemented, somehow it's not present in the flow definition hence the error.

Digging further into the error made me to discover that the node implementation of the fetch API is behind the official implementation. Some method like _raw_ that is the node version is not in the browser 😓

packages/sdk-middleware-http/src/http.js
```diff
@@ -50,6 +50,8 @@ export default function createHttpMiddleware (
               body: result,
               statusCode: res.status,
             }
+            if (options.includeHeaders)
+              parsedResponse.headers = res.headers.raw()
```
emmenko: Use `parseHeaders`

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -48,6 +48,65 @@ describe('Http', () => {
...
+          'Content-Type': 'application/json',
+        })
+        .get('/foo/bar')
+        .reply(200, { foo: 'bar' })
+
+      httpMiddleware(next)(request, response)
+    }),
+  )
+
+  it('should omit the headers when headers flag is false or omitted', () =>
```
emmenko: You don't need this test

wizzy25: Maybe to ensure in doesn't throw/fail if omitted??

emmenko: It's the same test as `execute a get request (success)` (L24)

packages/sdk-client/src/client.js
```diff
@@ -43,6 +43,7 @@ export default function createClient (options: ClientOptions): Client {
             rs.resolve({
               body: rs.body || {},
               statusCode: rs.statusCode,
+              headers: rs.headers || {},
```
emmenko: Test is missing

hisabimbola:
> Also make sure to update the documentation (e.g. sdk-client)

I read through the documentation and did not see anything to update, maybe something else was desired. I added documentation about the new flag in the http middleware docs

hisabimbola:
Also, I am planning to add `maskSensitiveHeaderData` flag.

hisabimbola:
@emmenko Improved the feature. Please review again.

PR #116, opened 2017-03-28
refactor(api-request-builder): pass project key as main option

### LINKED ISSUE #43: [api-request-builder] pass projectKey as main option
Instead of passing it every time as option of `.build()`, pass it once to `createRequestBuilder`.

> It's a BREAKING CHANGE

wizzy25:
#### Summary

affects: @commercetools/api-request-builder, @commercetools/csv-parser-price

### BREAKING CHANGE!!!
pass the projectKey as a required option to the ```createRequestBuilder``` method

resolves #43

### DIFF
--- a/docs/sdk/api/apiRequestBuilder.md
+++ b/docs/sdk/api/apiRequestBuilder.md
@@ -14,20 +14,23 @@ npm install --save @commercetools/api-request-builder
 <script>// global: CommercetoolsApiRequestBuilder</script>
 ```
 
-## `createRequestBuilder(customServices)`
+## `createRequestBuilder(projectKey, customServices)`
 
 Creates a *request builder* that allows to declaratively build a HTTP API request URI for the commercetools platform.
 
 #### Arguments
 
-1. `customServices` *(Object)*: A map of custom services that are not provided by default. This might be useful to build a request for a different API with similar query parameters.
+1. `projectKey` *(Object)*: A required object specifying the project key to use for the request, with the format:
+`{projectKey: 'test'}`. Even though this is required, the project key can be omitted from the URI by passing
+`{withProjectKey: false}` to `.build()`
+2. `customServices` *(Object)*: A map of custom services that are not provided by default. This might be useful to build a request for a different API with similar query parameters.
 
 A _service_ is created by defining its `features`. Features give a service specific _characteristics_ to correctly build URIs. For example, if a service can query a resource by ID you would include `queryOne`. Available features types are:
 
 - `query`: allows to use standard query capabilities (`page`, `perPage`, `sort`, `where`, `whereOperator`)
 - `queryOne`: allows to query a single resource (`byId`)
 - `queryExpand`: allows to use reference expansion (`expand`)
-- `search`: allows to use search capabilities (`text`, `fuzzy`, `facet`, `filter`, `filterByQuery`, `filterByFacets`)
+- `search`: allows to use search capabilities (`text`, `fuzzy`, `fuzzyLevel`, `facet`, `markMatchingVariants`, `filter`, `filterByQuery`, `filterByFacets`)
 - `projection`: allows to use projections capabilities (`staged`)
 
 ```js
@@ -35,6 +38,7 @@ import {
   createRequestBuilder,
   features,
 } from '@commercetools/api-request-builder'
+const projectKey = { projectKey: 'my-project' }
 const customServices = {
   users: {
     type: 'users',
@@ -45,7 +49,7 @@ const customServices = {
     ],
   },
 }
-const requestBuilder = createRequestBuilder(customServices)
+const requestBuilder = createRequestBuilder(projectKey, customServices)
 ```
 
 #### Usage example
@@ -54,14 +58,14 @@ const requestBuilder = createRequestBuilder(customServices)
 import { createRequestBuilder } from '@commercetools/api-request-builder'
 import { createClient } from '@commercetools/sdk-client'
 
-const requestBuilder = createRequestBuilder()
+const requestBuilder = createRequestBuilder({ projectKey: 'my-project-key' })
 const client = createClient({
   middlewares: [...],
 })
 const channelsUri = requestBuilder.channels
   .where('key = "foo"')
   .perPage(1)
-  .build({ projectKey: 'my-project-key' })
+  .build()
 const channelsRequest = {
   url: channelsUri,
   method: 'GET',
--- a/docs/sdk/api/apiRequestBuilder.md
+++ b/docs/sdk/api/apiRequestBuilder.md
@@ -14,15 +14,18 @@ npm install --save @commercetools/api-request-builder
 <script>// global: CommercetoolsApiRequestBuilder</script>
 ```
 
-## `createRequestBuilder(projectKey, customServices)`
+## `createRequestBuilder(options)`
 
 Creates a *request builder* that allows to declaratively build a HTTP API request URI for the commercetools platform.
 
 #### Arguments
 
-1. `projectKey` *(Object)*: A required object specifying the project key to use for the request, with the format:
-`{projectKey: 'test'}`. Even though this is required, the project key can be omitted from the URI by passing
-`{withProjectKey: false}` to `.build()`
+The `options` argument must be an object with a `projectKey` property, and optionally a `customServices` property
+```js
+    options -> { projectKey, customServices }
+```
+
+1. `projectKey` *(String)*: A required staring specifying the project key to use for the request, with the format. Even though this is required, the project key can be omitted from the URI by passing `{withProjectKey: false}` to `.build()`
 2. `customServices` *(Object)*: A map of custom services that are not provided by default. This might be useful to build a request for a different API with similar query parameters.
 
 A _service_ is created by defining its `features`. Features give a service specific _characteristics_ to correctly build URIs. For example, if a service can query a resource by ID you would include `queryOne`. Available features types are:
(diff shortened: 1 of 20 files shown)


wizzy25:
All builds for this repo are failing, including master @emmenko @hisabimbola

emmenko:
@wizzy25 looks like the integration tests are failing?

wizzy25:
@emmenko They're also failing on master... Reverting commits to see if/where something broke

packages/api-request-builder/src/build-query-string.js
```diff
@@ -1,7 +1,7 @@
 /* @flow */
 import type {
   ServiceBuilderDefaultParams,
-} from 'types/sdk'
+} from '../../../types/sdk'
```
emmenko: Why this change?

packages/api-request-builder/src/create-service.js
```diff
@@ -30,6 +31,7 @@ const requiredDefinitionProps = [
 
 export default function createService (
   definition: ServiceBuilderDefinition,
+  project: BuildOptions = {},
```
emmenko: This should be called `options`

packages/api-request-builder/src/create-service.js
```diff
@@ -42,7 +44,12 @@ export default function createService (
   if (!Array.isArray(definition.features) || !definition.features.length)
     throw new Error('Definition requires `features` to be a non empty array.')
 
+  if (!project.projectKey)
+    throw new Error('No project defined. Please enter a project key')
```
emmenko: You are validating twice (here and in `create-request-builder.js`). Just pick one ;)

packages/api-request-builder/test/create-service.spec.js
```diff
@@ -66,7 +73,7 @@ describe('createService', () => {
     )
   })
 
-  it('should throw if definition prop (features) is a non empty array', () => {
+  it('should throw if definition prop (features) is an empty array', () => {
```
emmenko: Why this change?

types/sdk.js
```diff
@@ -124,6 +124,9 @@ export type ServiceBuilderDefinition = {
 export type ApiRequestBuilder = {
   [key: string]: ServiceBuilder;
 }
+export type BuildOptions = {
```
emmenko: We might find a bit more specific name

packages/api-request-builder/src/build-query-string.js
wizzy25: ```types/sdk``` gives this lint error ```unable to resolve path to module import/no-unresolved```, and manually tracing it resolves to the types folder in the root ```../../../types/sdk``` (which gives no errors)

packages/api-request-builder/src/create-service.js
wizzy25: Thought so too... But is there a scenario where ```createService``` can be called on it's own? (not in ```create-request-builder```)? Or I'll just do the validation here so it's throws no matter where it's called ;-)

packages/api-request-builder/test/create-service.spec.js
wizzy25: The old one isn't correct ;)

types/sdk.js
wizzy25: Didn't want to really mess with the variable names, and since [you feel the variable should be called ```options```](https://github.com/commercetools/nodejs/pull/116#discussion_r108428216), what better could there be?

packages/api-request-builder/src/build-query-string.js
emmenko: This was working fine, there is an alias resolver for that (otherwise linting on master will fail). Are you sure you have the correct dependencies installed?

packages/api-request-builder/src/create-service.js
emmenko: Nope, only `create-request-builder` is exposed

emmenko: And yes, you can just do it here in `create-service.js` ;)

packages/api-request-builder/test/create-service.spec.js
emmenko: 🤔 why? `features` needs to be an array, so if it's **a non** empty array it throws

types/sdk.js
emmenko: Something like `ApiRequestBuilderServiceOptions`

packages/api-request-builder/src/build-query-string.js
wizzy25: Still having the errors straight from a fresh git clone on master.
More like a warning though as everything still "seems" to work fine as it is...

emmenko: Can you post here the output of `npm run lint`?

packages/api-request-builder/test/create-service.spec.js
wizzy25: In that case the error message is wrong: 
>Definition requires `features` to be a non empty array

Should instead be: 
>Definition requires `features` to be an empty array

...and [here](https://github.com/commercetools/nodejs/blob/43-projectKey-main-option/packages/api-request-builder/test/create-service.spec.js#L83), you still expect it to throw when features is **empty** 😏

emmenko: Ok, maybe we should split the test into 2 cases then ;)

packages/api-request-builder/src/build-query-string.js
hisabimbola: @wizzy25 it is a sublime specific error. Info how to fix it [here](https://github.com/benmosher/eslint-plugin-import/blob/master/README.md#sublimelinter-eslint)

emmenko: Please revert the changes then ;)

packages/csv-parser-price/test/main.spec.js
```diff
@@ -16,10 +16,13 @@ const logger = {
   info: () => {},
   verbose: () => {},
 }
+const apiConfig = {
```
hisabimbola: I would have prefered if this refactor is done in another PR...

wizzy25: yeah but then, the tests would fail

packages/csv-parser-price/src/main.js
```diff
@@ -183,9 +183,11 @@ export default class CsvParserPrice {
 // Easiest way to wrap the getCustomTypeDefinition in the memoize method
 CsvParserPrice.prototype.getCustomTypeDefinition = memoize(
   function _getCustomTypeDefinition (customTypeKey) {
-    const getTypeByKeyUri = createRequestBuilder().types
+    const getTypeByKeyUri = createRequestBuilder({
```
hisabimbola: Why are we changing this?

wizzy25: Because we pass the projectkey to `.build()` and now we will be  passing it to `createRequestBuilder()`. 
That's what this PR is about 🙄

wizzy25: That's why it's a breaking change

hisabimbola: Yeah, but that did not fully explain why a code in another package affects this...

wizzy25: This module [uses](https://github.com/commercetools/nodejs/blob/master/packages/csv-parser-price/src/main.js#L13) the `createRequestBuilder()` method, which is part of the `api-request-builder`, which is this PR

hisabimbola: Yeah, the fact that it uses it doesn't still validate the change

Do we change the module everytime another package is updated?

`csv-parser-price` uses `lodash` for example, sure we don't update our module everytime that gets updated

So you should see where I am driving to with this...

wizzy25: So what do you suggest? Because I don't know how we can pass and not pass the projectkey to the `.build()` method at the same time

hisabimbola: A more appropriate reason can be found [here](https://github.com/lerna/lerna#how-bootstrap-works)

Lerna works by symlinking internal modules when you do 'npm install', so when you make a change to a module, all other modules get's the new updated change, and this is the desired behaviour

hisabimbola:
I think this is good to go @emmenko 

Can you take a final look and let's ship this 🚀

wizzy25:
@emmenko updated docs. Please review

docs/sdk/api/apiRequestBuilder.md
```diff
@@ -14,27 +14,31 @@ npm install --save @commercetools/api-request-builder
 <script>// global: CommercetoolsApiRequestBuilder</script>
 ```
 
-## `createRequestBuilder(customServices)`
+## `createRequestBuilder(projectKey, customServices)`
```
emmenko: Sorry, now that I see this I think it's better to have named arguments

```js
// before
createRequestBuilder(projectKey, customServices)

// after
createRequestBuilder(options)
// options -> { projectKey, customServices }
```

/cc @hisabimbola ?

wizzy25: Do you mean to refactor `createRequestBuilder()` to take in one object as argument containing projectKey and customServices? 😶 🙄

hisabimbola: @wizzy25 and I actually discussed this format, but can't remember why we didn't go with it. We can refactor it. Looks better-managed params

emmenko: Cool, thanks and sorry for dragging this longer 😁

wizzy25:
@emmenko Hope this would be last time 😄 
kindly review

docs/sdk/api/apiRequestBuilder.md
```diff
@@ -14,38 +14,47 @@ npm install --save @commercetools/api-request-builder
...
 
 #### Arguments
 
-1. `customServices` *(Object)*: A map of custom services that are not provided by default. This might be useful to build a request for a different API with similar query parameters.
+The `options` argument must be an object with a `projectKey` property, and optionally a `customServices` property
+```js
+    options -> { projectKey, customServices }
+```
+
+1. `projectKey` *(String)*: A required staring specifying the project key to use for the request, with the format. Even though this is required, the project key can be omitted from the URI by passing `{withProjectKey: false}` to `.build()`
```
emmenko: `staring`?

wizzy25: oops... typo. 🙈 
But is that the only thing? So I could fix it all in one commit

wizzy25:
LGTM!!! 
@hisabimbola @emmenko

PR #126, opened 2017-04-11
feat(api-request-builder): support version when deleting

### LINKED ISSUE #112: [api-request-builder] Support version when deleting
### Description

Currently to do a delete, you have to manually add the version to the uri.

I find the user experience of this very bad, especially when compared to the old SDK

### Expected Behavior

Should expose a method like service.delete(versionNumber).build()
### Current Behavior

Example of current behaviour can be found [here](https://github.com/commercetools/nodejs/blob/6dce24c25665289ae7d21329a6f1b026b5a7096f/integration-tests/sdk/channels.it.js#L141)

wizzy25:
#### Description

BREAKING CHANGE:
Expose a method `.toDelete()` to pass in the version of an item to delete instead of passing it
directly to the uri

#### Todo
- Tests
    - [x] Unit
    - [x] Integration
- [x] Documentation

resolves #112


junajan:
Hello @wizzy25 .. am I correct that the method is called `toDelete` instead of just `delete` because the second one is a reserved word in JS?

wizzy25:
@junajan it's called `.toDelete()`

emmenko:
@wizzy25 can you explain the motivation for that?

PS: next time it would be nice to create an issue first, so that it can be discussed, before working on the PR. Thanks ;)

packages/api-request-builder/src/delete.js
```diff
@@ -0,0 +1,16 @@
...
+ * in order to generate a delete uri
+ *
+ * @param  {number} version - The version of the item
+ * @throws if `version` is missing or not a number.
+ * @return {Object} The instance of the service, can be chained.
+ */
+
+export default function toDelete (version: number): Object {
+  if (!version || typeof version !== 'number')
+    throw new Error('Required version to delete missing or not valid')
```
junajan: This error message looks strange to me: `Required version to delete missing or not valid`
What about to use something like this: `A resource version is invalid or missing`

But let's wait for others what do they think about it.

wizzy25:
@emmenko The [issue](https://github.com/commercetools/nodejs/issues/112) is 19 days old :-D

packages/api-request-builder/src/delete.js
wizzy25: @junajan I think the version is [only?] needed in the uri with a `DELETE` request

junajan: Yep, but I would expect there a word "IS" like this: `Required version to delete IS missing or it is not valid`

wizzy25: semantics................ @junajan :')

junajan: oki :grimacing:

emmenko:
@wizzy25 totally missed the link, sorry 😅

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -55,8 +55,7 @@ describe('CSV and CLI Tests', () => {
       })
     })
 
-    // FIXME: fails with `ERR! No type with key \'custom-type\' found`
-    test.skip('should take input from file', (done) => {
+    test('should take input from file', (done) => {
```
emmenko: Why is this change here?

packages/api-request-builder/src/delete.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Set the `version` number to the internal state of the service instance
+ * in order to generate a delete uri
+ *
+ * @param  {number} version - The version of the item
+ * @throws if `version` is missing or not a number.
+ * @return {Object} The instance of the service, can be chained.
+ */
+
+export default function toDelete (version: number): Object {
```
emmenko: Why not just name this function `version`? At the end it's exactly what it does.

```js
service.byId('123').version(1).build()

// or
service.byId('123').withVersion(1).build()
```

packages/api-request-builder/src/delete.js
emmenko: I wouldn't mention any "delete" thingy in this function. If we are considering `version` as just a query parameter, it can be used for anything, not just "delete".

So I would keep this generic. See also my other comment above.

docs/sdk/README.md
```diff
@@ -146,9 +146,13 @@ describe('Channels', () => {
   })
 
   it('delete', () => {
-    const uri = service.byId(channelResponse.id).build()
+    const uri = service
+      .byId(channelResponse.id)
+      .toDelete(channelResponse.version)
```
hisabimbola: Let's go with `delete()` as least to maintain a similar API with the old SDK...

packages/api-request-builder/src/delete.js
```diff
@@ -0,0 +1,16 @@
...
+  if (!version || typeof version !== 'number')
```
hisabimbola: You can just do

```
if (typeof version !== 'number)
```

achieves same...

packages/api-request-builder/src/delete.js
hisabimbola: Let's use this

`A resource version is invalid or missing` or something like this. This explains more what *version* is and easily connected to the platform definition of version

docs/sdk/README.md
hisabimbola: Scratch this, I go with @emmenko suggestion...

integration-tests/cli/csv-parser-price.it.js
wizzy25: I just fixed all tests in this PR... Okay then, I'd create a separete PR for this 👍🏾

docs/sdk/README.md
adnasa: @hisabimbola shouldn't you be in vacation? :)

wizzy25:
Implemented changes... Kindly review again 
cc @junajan @emmenko @adnasa @hisabimbola(Enjoy your holiday;))

emmenko:
🤔 did you push? there is no new commit...

wizzy25:
😃😃 Pushed now @emmenko

Issue #145, opened 2017-04-28
[sync-actions] Can't add/change inventory quantity to 0


lufego:
### Description
At the moment, if you try to add/change a inventory quantity with number "0" the API returns a bad request error

```js
export function buildBaseAttributesActions ({
  actions,
  diff,
  oldObj,
  newObj,
}) {
  return actions
    .map((item) => {
      const key = item.key // e.g.: name, description, ...
      const actionKey = item.actionKey || item.key
      const delta = diff[key]
      const before = oldObj[key]
      const now = newObj[key]

      if (!delta) return undefined

      if (!now && !before) return undefined

      if (now && !before) // no value previously set
        return { action: item.action, [actionKey]: now }

      if (!now && !{}.hasOwnProperty.call(newObj, key)) // no new value
        return undefined

      if (!now && {}.hasOwnProperty.call(newObj, key)) // value unset
        return { action: item.action }
    })
    .filter(action => action)
}
```

### Expected Behavior
It should assemble the correct request body

### Current Behavior
Its assembling the request body without a required field (quantity).
This is happening because 0 is falsy, and on the buildBaseAttributesActions function it relies only on the value in order to define what to return

### Possible Solution
Maybe adding a variable to check the existence of the value (`hasNow` and `hasBefore`), so 0 wouldn't return false in this case.

PR #146, opened 2017-04-28
User can't add/change inventory quantity to 0

lufego:
This will fix https://github.com/commercetools/nodejs/issues/145

#### Summary
At the moment, if you try to add/change a inventory quantity with number "0" the API returns a bad request error

#### Description
This is happening because 0 is falsy, and on the buildBaseAttributesActions function it relies only on the value in order to define what to return
I have added a variable to check the existence of the value (`hasNow` and `hasBefore`), so 0 wouldn't return false in this case.

#### Todo

- Tests
    - [x] Acceptance

### DIFF
--- a/packages/sync-actions/src/utils/common-actions.js
+++ b/packages/sync-actions/src/utils/common-actions.js
@@ -1,3 +1,4 @@
+import isNil from 'lodash.isnil'
 import clone from './clone'
 import * as diffpatcher from './diffpatcher'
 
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/utils/common-actions.js
```diff
@@ -24,18 +25,19 @@ export function buildBaseAttributesActions ({
       const delta = diff[key]
       const before = oldObj[key]
       const now = newObj[key]
-
+      const hasBefore = !isNil(oldObj[key])
+      const hasNow = !isNil(newObj[key])
       if (!delta) return undefined
 
-      if (!now && !before) return undefined
+      if (!hasNow && !hasBefore) return undefined
```
emmenko: You don't need the double negation `!isNil(...)` and `!hasNow`.

```js
const isNotDefinedBefore = isNil(oldObj[key])
const isNotDefinedNow = isNil(oldObj[key])
```

lufego:
@emmenko Your comment has been addressed accordingly, thanks

PR #148, opened 2017-05-03
115 picture upload

### LINKED ISSUE #115: [sdk-middleware-http] Support for product image uploads
### Description
I'm trying to use the node sdk to upload a product image to commercetools.  I'm currently using code that looks like:
```
          // let's upload to the product's images endpoint
          let imgType = mime.lookup(tmpFile);
          let img = fs.readFileSync(tmpFile);
          const pimgUri = requestBuilder.products.byId(existingProduct.id).build({ projectKey: projectKey });
          const addImageRequest = {uri: pimgUri + '/images', method: "POST", headers: {'Content-Type': imgType}, body:img};
          /// This won't work with current ct sdk
          client.execute(addImageRequest).catch((e) => {
            console.log(e);
            process.exit(45);
          });
```
I noticed that under the hood, the node sdk uses node-fetch, which would support sending a stream or buffer object as the request body.  However, the sdk seems to drop that support in favor of ensuring every body is sent as JSON.

### Expected Behavior
I'd expect to be able to upload an image to the product's images endpoint using the node sdk.

### Current Behavior
The sdk middleware seems to be JSON.stringify()'ing my Buffer/Stream object and sending that to commercetools, resulting in a 'This jpg encoding is not supported' error from commercetools.

### Context
While writing an importer for a potential customer, I needed to import their product images as well.  I am currently unable to do so using the node sdk.  I am able to use a call to curl, or a custom Request object to accomplish this in the meantime.

### Possible Solution
I'm not entirely sure, but https://github.com/commercetools/nodejs/blob/master/packages/sdk-middleware-http/src/http.js#L25 looks like it would cause this issue

hisabimbola:
#### Summary

Fixes #115 
#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


emmenko:
The title is a bit misleading, it's more about "Enabling HTTP requests with `Buffer` as a body"

hisabimbola:
Yeah, @wizzy25 tested it

Can you confirm that 😜

PR #147, opened 2017-04-30
feat(discount-codes): implement function to generate codes

wizzy25:
affects: @commercetools/discount-code-generator

#### Summary

Implement the discount code generator function as part of [this](https://github.com/commercetools/nodejs/issues/142#issue-224808585) checklist

Issue #26, opened 2017-01-19
Use caret semver range for development dependencies?


Siilwyn:
Let's discuss if we want to use exact versions or semver ranges for the development dependencies.
I would vote for using semver ranges to decrease the amount of Greenkeeper PRs. And feel confident in doing this because we have the dependency lockfile.

PhilippSpo:
👍 When and how do we update the lockfile then?

emmenko:
Yep, someone has to update the lockfile. I would still tend to be careful with that, having greenkeepr PRs might seem a bit spammy but it kind of forces you to check what changed and maybe try out the new version before simply using it.

Siilwyn:
How is it currently being updated? Does that change when we use semver ranges? 😕

emmenko:
Currently we update it manually and commit the changes. So the process is more or less the same.

I'm fine having **some** deps with caret range, and I don't mind having greenkeeper PRs ;)

Siilwyn:
Okay I think it's best to be consistent so let's go with exact versions then. Before closing I'm wondering why `npm-check-updates` doesn't have an exact version.

hisabimbola:
Personally, I feel using caret might make much sense, especially for dev dependencies.

> I'm fine having some deps with caret range, and I don't mind having greenkeeper PRs ;)

...and the PRs don't get merged anyways 🙄

PR #149, opened 2017-05-03
feat(discount-codes): implement generator cli

wizzy25:
#### Summary

Awaiting #147 to get merged before this will be functional

#### Description

Implement CLI to pass configuration and attributes to the generator.


packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,218 @@
...
+import flatten, { unflatten } from 'flat'
+
+import discountCodeGenerator from './main'
+import { version } from '../package.json'
+
+process.title = 'discountCodeGenerator'
+
+const args = yargs
+  .usage(
+    `\n
```
adnasa: This shouldn't be not needed at all when you are using string-interpolation.

packages/discount-code-generator/package.json
```diff
@@ -18,11 +18,21 @@
   "contributors": [
     "Williams Omayuku <williams.omayuku@commercetools.com>"
   ],
-  "main": "lib/index.js",
+  "main": "./lib/index.js",
+  "bin": {
+    "discountcodegen": "bin/discountcodegen.js"
```
emmenko: Nitpick: it would be better to use dashes like `discount-code-gen`

packages/discount-code-generator/package.json
```diff
@@ -18,11 +18,21 @@
...
   ],
   "dependencies": {
-    "tokgen": "^1.0.0"
+    "csv-parser": "^1.11.0",
+    "flat": "^2.0.1",
+    "json2csv": "^3.7.3",
+    "npmlog": "^4.0.2",
+    "pretty-error": "^2.1.0",
+    "tokgen": "^1.0.0",
+    "yargs": "^7.1.0"
```
emmenko: Note: be aware if you want to use `^` range or be more restrictive about the version that you allow to be used (to avoid possible "unwanted" changes)

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+import PrettyError from 'pretty-error'
+import yargs from 'yargs'
+import json2csv from 'json2csv'
+import csv from 'csv-parser'
+import flatten, { unflatten } from 'flat'
+
+import discountCodeGenerator from './main'
+import { version } from '../package.json'
+
+process.title = 'discountCodeGenerator'
```
emmenko: What does it do?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+
+  .option('version', {
+    alias: 'v',
+    type: 'boolean',
+  })
+  .version('version', 'Show version number.', version)
+
+  .option('quantity', {
+    alias: 'q',
+    describe: 'Quantity of discount codes to generate.',
```
emmenko: Do you allow any number or is there a min / max limit?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+    demandOption: true,
+  })
+  .coerce('quantity', arg => parseInt(arg, 10))
+
+  .option('length', {
```
emmenko: I think it's a bit too generic, maybe `code-length`?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+    alias: 'l',
+    default: 11,
+    describe: 'Length of the discount codes to generate.',
+  })
+  .coerce('length', arg => parseInt(arg, 10))
+
+  .option('prefix', {
```
emmenko: Again, maybe `code-prefix`?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+    alias: 'p',
+    default: '',
+    describe: 'Prefix for each code. No prefix will be used if omitted.',
+  })
+
+  .option('inputFile', {
```
emmenko: I think in this case it's enough to say `input`

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+    demandOption: true,
+  })
+  .coerce('inputFile', (arg) => {
+    if (arg.match(/\.json$/i) || arg.match(/\.csv$/i))
+      return fs.createReadStream(String(arg))
+
+    throw new Error('Invalid input file format. Must be CSV or JSON')
+  })
+
+  .option('outputFile', {
```
emmenko: Same here, just `output`

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,217 @@
...
+        strict: true,
+      }))
+        .on('error', (error) => {
+          reject(error)
+        })
+        .on('data', (data) => {
+          const arrayDelim = _args.multivalueDelimiter
+          if (data.cartDiscounts)
+          // eslint-disable-next-line no-param-reassign
+            data.cartDiscounts = data.cartDiscounts.split(arrayDelim)
```
emmenko: Can you maybe document why this case is necessary?

packages/discount-code-generator/package.json
wizzy25: This still goes back to #26. But then, you don't have a problem with greenkeeper PRs 😉

packages/discount-code-generator/src/cli.js
wizzy25: Sets the process nam. Can be seen on Activity Monitor (Mac) or Task Manager (Windows)

packages/discount-code-generator/src/cli.js
wizzy25: Thinking about this, it works for more than a million (for CSV) and 700 thousand (for JSON) on my machine. Although I think limiting it to 500 thousand will be good for both scenarios, or perhaps something more modest? what do you think? cc @hisabimbola 
And of course, the minimum is 1 :-)

packages/discount-code-generator/src/cli.js
wizzy25: My idea of this cli is to be independent of the imports; such that one could generate the codes they need now and add the cart discounts later before being imported. With this model, the module will throw without this case if the input has no `cartDiscounts`

wizzy25:
Did some refactoring... What do you think? @emmenko @hisabimbola

packages/discount-code-generator/package.json
```diff
@@ -18,11 +18,21 @@
   "contributors": [
     "Williams Omayuku <williams.omayuku@commercetools.com>"
   ],
-  "main": "lib/index.js",
+  "main": "./lib/index.js",
```
hisabimbola: why you changing this?

wizzy25: It works both ways... I think it was just a semantic change on my path

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,227 @@
...
+        resolve(total)
+      })
+    } else if (outputStream.path.match(/\.csv$/i)) {
+      // Convert to csv and write to file
+      const arrayDelim = _args.multivalueDelimiter
+      const flatObjects = outputData.map((obj) => {
+        if (obj.cartDiscounts)
+        // eslint-disable-next-line no-param-reassign
+          obj.cartDiscounts = obj.cartDiscounts.join(arrayDelim)
+        return flatten(obj)
```
hisabimbola: why are you flatten the object again?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,227 @@
...
+        del: _args.delimiter,
+      })
+      outputStream.on('error', (error) => {
+        reject(error)
+      })
+      outputStream.end(csvOutput)
+      outputStream.on('finish', () => {
+        resolve(total)
+      })
+    } else reject(new Error('Invalid file format. Must be CSV or JSON'))
```
hisabimbola: Indent this line properly

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,227 @@
...
+}
+
+resolveInput(args)
+  .then((attributes) => {
+    const options = buildOptions(args)
+    const codes = discountCodeGenerator(options, attributes)
+    return resolveOutput(args, codes)
+  })
+  .then((total) => {
+    process.stdout.write(`Successfully generated ${total} discount codes\n`)
```
hisabimbola: This should be done via the logger module, if the user wants the code to be output to _stdout_ inorder to pipe it to another module, this is going to cause an error.

So use the logger to log stuff and the logger should decide where to log either a file or stdout

packages/discount-code-generator/package.json
hisabimbola: Let's make it consistent with others

packages/discount-code-generator/src/cli.js
wizzy25: It's properly indented... Anything different produces lint errors

packages/discount-code-generator/src/cli.js
wizzy25: Could you give more context on this?

packages/discount-code-generator/package.json
wizzy25: @hisabimbola That would be relative: In this repo, 4 packages are like this (this would be the 5th)

Also, check [this](#discussion_r114959894) out

packages/discount-code-generator/src/cli.js
hisabimbola: I am curious to know how to arrive at those numbers but no hard opinion here, I will prefer a hard limit of 1,000,000

If there are going to be piping this to the discount-code-importer (which most likely is the case) then we probably want to limit it down to reduce the stress on the importer and concurrent API calls

Also, I spoke to martin and he mentioned that there is probably a use-case where a customer wants to generate millions of code 😁

hisabimbola: Also maybe after this, I will like us to rethink how we stream things out but that's a discussion for another time...

wizzy25: I tested with various numbers, and calling `JSON.stringify()` causes memory errors on very large numbers. On my system, this limit is about 750,000 (could be more or less on other systems) hence my reason for limiting to 500,000. The alternative to this which I tested would be to stream the output, stringify and write per object using `JSONStream.stringify()`, but this is not feasible as it takes an insane amount of time (writing 1,000,000 codes takes about 20mins). If there's another alternative which I don't know of, I would be happy to be pushed in the right direction, but for now, I think I'll stick to 500,000 codes for optimum performance. The process can be run multiple times if more codes are desired.
Cc @hisabimbola @emmenko

wizzy25:
I updated this PR... please review again @hisabimbola @emmenko

packages/discount-code-generator/src/cli.js
emmenko: Let's start with something, we can always improve later as we need to ;)

packages/discount-code-generator/src/cli.js
emmenko: Ok, whatever the reason is then maybe add a comment about it next to it, so that other people know why.

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,235 @@
...
+  .version('version', 'Show version number.', version)
+
+  .option('quantity', {
+    alias: 'q',
+    describe: 'Quantity of discount codes to generate. (Between 1 and 500000)',
+    demandOption: true,
+  })
+  .coerce('quantity', (arg) => {
+    const quantity = parseInt(arg, 10)
+    if (quantity <= 0 || quantity > 500000)
```
emmenko: Document the reason of this number please, whatever that is :)

wizzy25:
LGTM??? @emmenko

emmenko:
🚀

adnasa:
@wizzy25 press that button! ✅

wizzy25:
🎯

PR #165, opened 2017-05-18
fix(csv-parser-price): add currencyCode to data

### LINKED ISSUE #164: [csv-parser-price] value.currencyCode is missing from transformed data
### Description

For some weird reasons, the csv-parser-price module does not output the correct json data. The currencyCode is missing from the data returned from the module, hence making the data invalid
```js

```

### Expected Behavior

`centAmount` should be added to the data returned from the module
Expected data format
```
{
	"prices": [{
		"sku": "123",
		"prices": [{
			"variant-sku": "123",
			"value": {
				"centAmount": 5000
				"currencyCode": "EUR"
			}
		}]
	}]
}
```
### Current Behavior

Current data format
```
{
	"prices": [{
		"sku": "123",
		"prices": [{
			"variant-sku": "123",
			"value": {
				"centAmount": 5000
			}
		}]
	}]
}
```

hisabimbola:
affects: @commercetools/csv-parser-price

#### Summary

Fixes #164 
#### Todo

- Tests
    - [ ] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/csv-parser-price/test/main.spec.js
+++ b/packages/csv-parser-price/test/main.spec.js
@@ -51,8 +51,11 @@ describe('CsvParserPrice::parse', () => {
 
     const outputStream = streamtest['v2'].toText((error, result) => {
       const prices = JSON.parse(result).prices
+      const price = prices[0].prices[0]
       expect(prices.length).toBe(2)
       expect(prices[0].sku).toBeTruthy()
+      expect(price.value.centAmount).toBeTruthy()
+      expect(price.value.currencyCode).toBeTruthy()
       done()
     })
     csvParserPrice.parse(readStream, outputStream)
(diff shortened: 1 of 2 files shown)


packages/csv-parser-price/test/main.spec.js
```diff
@@ -51,8 +51,11 @@ describe('CsvParserPrice::parse', () => {
 
     const outputStream = streamtest['v2'].toText((error, result) => {
       const prices = JSON.parse(result).prices
+      const price = prices[0].prices[0]
```
junajan: How does this structure look like? Is the `prices[0].prices[0]` correct?

hisabimbola: So the actual price is in the second `prices`, majorly because this is what is expected from the node-cli

PR #166, opened 2017-05-18
fix(csvParserPrice): Fix log poluting of JSON output in stdout

### LINKED ISSUE #153: [csv-parser-price] Handle logging appropriately
### Description

Unlike the old repo, status will always be printed to stdout. This can "disturb" the standard output

### Expected Behavior

Logging should be separated from the output content

### Possible Solution

[This](https://github.com/commercetools/csv-parser-price/blob/master/src/cli.js#L54), which is missing in this repo will be a good place to start

junajan:
#### Summary
Fixes #153 

#### Description
I had a problem when writing logs to the special log file because when an error occurred, the process was killed before it had a chance to flush all logs. And in a case of a successful processing, I don't know how to wait for highland to process and output all data so I can close the logFile stream (for example `stdout` does not have `end` event so I don't know how to catch the end of processing).

Therefore I fixed it differently - only error logs are printed out when the JSON is sent to stdout.


wizzy25:
>the process was killed before it had a chance to flush all logs

What happens if you modify the [error handler](https://github.com/commercetools/nodejs/blob/2908e768e8cef05f62720f29fb80d93ea51e65e9/packages/csv-parser-price/src/cli.js#L111)? Precisely, setting `process.exitCode` to 1 instead of calling `process.exit(1)`?

hisabimbola:
Nitpick @junajan 

The repo name is `csv-parser-price` so commit message should have `csv-parser-price` not `csvParserPrice` 😁

packages/csv-parser-price/src/cli.js
```diff
@@ -85,14 +87,12 @@ Convert commercetools price CSV data to JSON.`,
...
   .option('logLevel', {
-    default: 'info',
     describe: 'Logging level: error, warn, info or verbose.',
   })
-  .coerce('logLevel', (arg) => {
-    npmlog.level = arg
-  })
   .argv
 
+npmlog.level = args.logLevel || minLogLevel
```
hisabimbola: How does this solve the issue?

junajan: This will set a `logLevel` to error when the output is `stdout` unless user set it to anything else.

So if you want to print a result to stdout, only errors are displayed but you still have an option to turn on more verbose logging if you want to debug it.

hisabimbola: I still don't get how this solves the issue

Let's have a call about this later

junajan: @hisabimbola - let me know what time does suit you best and we can discuss it.

hisabimbola: We already had a working feature of this before, please refer to this [line](https://github.com/commercetools/csv-parser-price/blob/master/src/cli.js#L54) to see how to improve this

junajan: Updated here 89f66b9

csvparserprice.log
```diff
@@ -0,0 +1,5 @@
+info Starting conversion
```
hisabimbola: why are we committing this file?

docs/cli/csv-parser-price.md
```diff
@@ -32,6 +32,8 @@ Options:
                                                                [default: "info"]
 ```
 
+Note that when the stdout is used as an output stream all log messages are written to `.csvparserprice.log` log file.
```
hisabimbola: 💯

packages/csv-parser-price/csvparserprice.log
```diff
@@ -0,0 +1,3 @@
+info Starting conversion
```
hisabimbola: we should not commit log files...

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -111,10 +111,12 @@ describe('CSV and CLI Tests', () => {
...
 
       exec(`${binPath} -p ${projectKey} -i ${csvFilePath} --logLevel verbose`,
-        (error, stdout, stderr) => {
+        (error, stdout) => {
           expect(error.code).toBe(1)
-          expect(stdout).toBeFalsy()
-          expect(stderr).toMatch(/\.js:\d+:\d+/)
+          expect(stdout).toMatch('{"prices":[]}')
+
+          const errorLog = fs.readFileSync('csvparserprice.log', 'utf8')
```
hisabimbola: let's find another way to implement this, but log files should not be committed

junajan: I will put them to ignore list. What output should we return if there was an error during processing?

hisabimbola: we can make the log file customizable and create a tmp file for this test, I feel

but then, if there is an error that makes the script to stop, we probably want to output that to the user *and* log it to a file...

junajan: On other projects, I use multiple log streams which log everything into a log file and if there are any errors/criticals then they send logs also to stdout but it looks like the `npmlog` does support only one stream. I will play with it a little bit.

hisabimbola: `npmlog` is only for *logs* so I will not expect it to handle both, if there is an error, just return it or something..I am not expecting something special...

junajan:
Hello all.. I am closing this PR because it got pretty messy. I created a new one with all requested changes here https://github.com/commercetools/nodejs/pull/208

PR #169, opened 2017-05-19
fix(api-request-builder): Allow SDK to send markMatchingVariants=false

### LINKED ISSUE #168: [api-request-builder] Send markMatchingVariants=false by default
### Description
Based on the [documentation](http://dev.commercetools.com/http-api-projects-products-search.html#search-productprojections) the `markMatchingVariants` parameter will add a flag to all variants which match search criteria.

On the API it is by default set to `true` but in the new SDK we can only [set it](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/query-search.js#L49) to `true` because `false` value [is not sent](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/build-query-string.js#L60) to the API so there is no chance to turn this feature off.

### Expected Behavior
We should have a possibility to set this feature off or it should be turned off by default.

### Possible Solutions
We have two options:
 1) by default send `markMatchingVariants=false`
 2) allow users to set it to false by calling `markMatchingVariants(false)`

junajan:
Resolves #168 

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


hisabimbola:
We should update the documentation about this also

junajan:
@hisabimbola - note in the documentation about default `markMatchingVariants` value was added here 2eef6ee

packages/api-request-builder/src/build-query-string.js
```diff
@@ -57,8 +57,7 @@ export default function buildQueryString (
       queryString.push('fuzzy=true')
     if (fuzzyLevel)
       queryString.push(`fuzzyLevel=${fuzzyLevel}`)
-    if (markMatchingVariants)
-      queryString.push('markMatchingVariants=true')
+    queryString.push(`markMatchingVariants=${markMatchingVariants.toString()}`)
```
adnasa: It is not clear where it has been set to `false` by default.

junajan: Right here https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-params.js#L51

It was already there but query builder was sending just a `true` value.

adnasa: Cool thanks @junajan

PR #150, opened 2017-05-05
docs(discount-codes): Write documentation for discount code generator

wizzy25:
affects: @commercetools/discount-code-generator

Issue #143, opened 2017-04-27
Discount code importer


wizzy25:
### Description

A new module to import new or update existing discount codes to the API

### Todo
- [x] Read batches

- [x] Decide if to update or create

- [x] Create codes

- [x] Update codes

- [x] Write documentation

- [x] Add integration tests

PR #171, opened 2017-05-23
Discount Code importer

wizzy25:
#### Summary

As part of [this project](https://github.com/commercetools/nodejs/projects/2) and #143, implement a module to import discount codes to the CTP

#### Description

This module is used to create new or update existing discount codes on the CTP
Another issue/PR would be created for the documentation


adnasa:
any relation to #150 ?
Can't it be included in this PR as well (then close #150)?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+    this._resetSummary()
+  }
+
+  // Use static method because this is not called on any instance
+  static _buildPredicate (codeObjects: CodeDataArray): string {
+    const predicateArray = codeObjects.map(codeObject => codeObject.code)
+    return `code in ("${predicateArray.join('", "')}")`
+  }
+
+  performStream (chunk: ChunkOptions, cb: () => mixed) {
```
junajan: Why is it done with a callback function and not with the Promise?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+    this._processBatches(chunk)
+      .then(() => cb())
+      .catch(err => cb(err.body))
+  }
+
+  _processBatches (codes: CodeDataArray) {
+    // Batch to `batchSize` to reduce necessary fetch API calls
+    const batchedList = _.chunk(codes, this.batchSize)
+    return Promise.map(batchedList, (codeObjects: CodeDataArray) => {
```
junajan: Maybe add some concurrency limit?

junajan: You can use a `mapSeries`  instead of `map` with concurrency 1.

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+      warn: npmlog.warn.bind(this, ''),
+      info: npmlog.info.bind(this, ''),
+      verbose: npmlog.verbose.bind(this, ''),
+    }
+    this._resetSummary()
+  }
+
+  // Use static method because this is not called on any instance
+  static _buildPredicate (codeObjects: CodeDataArray): string {
+    const predicateArray = codeObjects.map(codeObject => codeObject.code)
```
junajan: Just a shorter version `_.map(codeObjects, 'code')`

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+    const batchedList = _.chunk(codes, this.batchSize)
+    return Promise.map(batchedList, (codeObjects: CodeDataArray) => {
+      // Build predicate and fetch existing code
+      const predicate = DiscountCodeImport._buildPredicate(codeObjects)
+      const service = createRequestBuilder({
+        projectKey: this.apiConfig.projectKey,
+      })
+      const uri = service.discountCodes
+        .where(predicate)
+        .perPage(this.batchSize)
```
junajan: If we use a default `batchSize` which is set to 50 and on the API there will be more items matching this predicate - will this load them all?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+        .build()
+      return this.client.execute({
+        uri,
+        method: 'GET',
+      })
+        .then((response: Object) => {
```
junajan: Just o nitpick here - you can use a destructuring to make it shorter:
`.then(({ body: { results: existingCodes } }: Object)`

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+          const existingCodes = response.body.results
+          return this._createOrUpdate(codeObjects, existingCodes)
+        })
+    }, { concurrency: 1 }) // Run one batch at a time
+      .then(() => Promise.resolve(this.summaryReport()))
```
junajan: Is `Promise.resolve` necessary here? You are already inside of a promise.

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+      .catch(error => Promise.reject({
+        error,
+        summary: this._summary,
+      }),
+      )
+  }
+
+  _createOrUpdate (newCodes: CodeDataArray, existingCodes: CodeDataArray) {
+    return Promise.map(newCodes, (newCode: CodeData) => {
```
junajan: Can you please add some concurrency to Promise.map?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+          }
+          this._summary.createErrorCount += 1
+          this._summary.errors.push(error)
+          return Promise.reject(error)
+        })
+    })
+  }
+
+  _update (newCode: CodeData, existingCode: CodeData) {
+    const service = createRequestBuilder({
```
junajan: Can you move `createRequestBuilder` to a new method and reuse it everywhere?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+      createErrorCount,
+      updateErrorCount,
+    } = this._summary
+    let message = ''
+    if (created === 0 && updated === 0 &&
+      createErrorCount === 0 && updateErrorCount === 0)
+      message = 'Summary: nothing to do, everything is fine'
+    else
+      // eslint-disable-next-line max-len
+      message = `Summary: there were ${created + updated} successfully imported discount codes (${created} were new, ${updated} were updates and ${unchanged} were unchanged).`
```
junajan: `updates` -> `updated`

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+    if (createErrorCount || updateErrorCount)
+      // eslint-disable-next-line max-len
+      message += ` ${createErrorCount + updateErrorCount} errors occured (${createErrorCount} create errors and ${updateErrorCount} update errors.)`
+    const report = {
```
junajan: Is the `report` variable necessary here? You can just return a report object.

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+  // Use static method because this is not called on any instance
+  static _buildPredicate (codeObjects: CodeDataArray): string {
+    const predicateArray = codeObjects.map(codeObject => codeObject.code)
+    return `code in ("${predicateArray.join('", "')}")`
+  }
+
+  performStream (chunk: ChunkOptions, cb: () => mixed) {
+    this._processBatches(chunk)
+      .then(() => cb())
+      .catch(err => cb(err.body))
```
junajan: In the `this._processBatches` you return this error object:
```js
.catch(error => Promise.reject({
	error,
	summary: this._summary,
	}),
)
```

But here you expect that the error will have a `body` property:
`.catch(err => cb(err.body))`

packages/discount-code-importer/test/main.spec.js
```diff
@@ -1,11 +1,253 @@
...
+      codeImport._create = jest.fn(() => Promise.resolve())
+      codeImport._update = jest.fn(() => Promise.resolve())
+    })
+
+    it('should be defined', () => {
+      expect(codeImport._createOrUpdate).toBeDefined()
+    })
+
+    it('should call `_update` if code already exists', async () => {
+      await codeImport._createOrUpdate(codes, existingCodes)
```
junajan: 👍 for async - await

wizzy25:
@adnasa #150 is for a different package (code generator), while this is for the importer; hence separate PRs :-)

packages/discount-code-importer/src/main.js
wizzy25: And also a faster one apparently 😃

packages/discount-code-importer/src/main.js
wizzy25: There can't be more items matching it because the predicate also has the same amount of codes and the codes are unique (there could only be less if the code doesn't yet exist)

packages/discount-code-importer/src/main.js
wizzy25: Using concurrency here is not really necessary, and instead slows it down

packages/discount-code-importer/src/main.js
wizzy25: As we don't know how the success/error may be handled,, it's better to wrap it in a promise, which can be unwrapped when needed.

packages/discount-code-importer/src/main.js
wizzy25: I believe `updates` is correct here

hisabimbola: I think `updated` is more appropriate except you change `unchanged` too

packages/discount-code-importer/test/main.spec.js
adnasa: yeah, finally we get to see this in one for our repos

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+import { createAuthMiddlewareForClientCredentialsFlow }
+  from '@commercetools/sdk-middleware-auth'
+import { createHttpMiddleware } from '@commercetools/sdk-middleware-http'
+import { createUserAgentMiddleware }
+  from '@commercetools/sdk-middleware-user-agent'
+import { createSyncDiscountCodes } from '@commercetools/sync-actions'
+import { version } from '../package.json'
+
+export default class DiscountCodeImport {
+  // Set flowtype annotations
```
adnasa: Is this a necessary annotation?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+    })
+
+    this.syncDiscountCodes = createSyncDiscountCodes()
+
+    this.logger = logger || {
+      error: npmlog.error.bind(this, ''),
+      warn: npmlog.warn.bind(this, ''),
+      info: npmlog.info.bind(this, ''),
+      verbose: npmlog.verbose.bind(this, ''),
+    }
```
adnasa: Don't we have a general logger that can be re-used across all packages? ☝🏽

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+  ) {
+    this.batchSize = options.batchSize || 50
+    this.apiConfig = options.apiConfig
+    this.continueOnProblems = options.continueOnProblems || false
+    this.client = createClient({
+      middlewares: [
+        createAuthMiddlewareForClientCredentialsFlow(this.apiConfig),
+        createUserAgentMiddleware({
+          libraryName: 'discount-code-importer',
+          libraryVersion: version,
```
adnasa: I would prefer a context given on the `version` as variable, because it is easy to think that this is an env variable

```
import pkg from 'package.json'

pkg.version // <-- use this, give it context
```

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+      warn: npmlog.warn.bind(this, ''),
+      info: npmlog.info.bind(this, ''),
+      verbose: npmlog.verbose.bind(this, ''),
+    }
+    this._resetSummary()
+  }
+
+  // Use static method because this is not called on any instance
+  static _buildPredicate (codeObjects: CodeDataArray): string {
+    const predicateArray = _.map(codeObjects, 'code')
```
adnasa: is underscore necessary here? can't we use `Array.map`/`Array.reduce`?

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+          return Promise.reject(error)
+        })
+    })
+  }
+
+  _update (newCode: CodeData, existingCode: CodeData) {
+    const actions = this.syncDiscountCodes.buildActions(newCode, existingCode)
+    // don't call API if there's no update action
+    if (!actions.length)
+      return Promise.resolve({ statusCode: 304 })
```
adnasa: can be a one-liner

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+      created,
+      updated,
+      unchanged,
+      createErrorCount,
+      updateErrorCount,
+    } = this._summary
+    let message = ''
+    if (created === 0 && updated === 0 &&
+      createErrorCount === 0 && updateErrorCount === 0)
+      message = 'Summary: nothing to do, everything is fine'
```
adnasa: _nothing to do_ is unnecessary (not so friendly message) 👶🏽

packages/discount-code-importer/test/main.spec.js
```diff
@@ -1,11 +1,247 @@
...
+      expect(codeImport._update).toHaveBeenCalledTimes(2)
+      expect(codeImport._update).toHaveBeenCalledWith(
+        codes[0],
+        existingCodes[0],
```
adnasa: for future reference (maybe you can do it here if you want, it's up to you);
this is quite hard to understand, it would be nice to see to give these a name

```js
const [iAmANewCode] = codes
const [iAmAnExistingCode] = existingCodes

expect(codeImport._update).toHaveBeenCalledWith(iAmANewCode, iAmAnExistingCode)
```

packages/discount-code-importer/src/main.js
wizzy25: Semantics... then I'd also change `new` to `newly created` or `created`

packages/discount-code-importer/src/main.js
wizzy25: Yes it is. you can read more about it [here](https://flow.org/en/docs/types/classes/#toc-class-fields-properties)

packages/discount-code-importer/src/main.js
wizzy25: Not necessary, but lodash is already used here and it makes the code cleaner (used native map in earlier commits). And lodash is also significantly [faster](https://jsperf.com/native-map-vs-lodash-map).

packages/discount-code-importer/src/main.js
wizzy25: 😃😃 Default message borrowed from other modules

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+  logger: LoggerOptions;
+  _summary: Summary;
+  syncDiscountCodes: SyncAction;
+
+  constructor (
+    logger: LoggerOptions,
+    options: ConstructorOptions,
+  ) {
+    this.batchSize = options.batchSize || 50
+    this.apiConfig = options.apiConfig
```
hisabimbola: Since this is required and no default then we should validate if it's not present

packages/discount-code-importer/src/main.js
junajan: yeah, you are right..  https://jsperf.com/array-prototype-map-vs-map

packages/discount-code-importer/src/main.js
hisabimbola: Since we already added concurrency above, we already limit the number of simultaneous request that the module can send to the API so I am not sure concurrency makes much sense here

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+              this._summary.updated += 1
+            return Promise.resolve()
+          })
+          .catch((error) => {
+            if (this.continueOnProblems) {
+              this._summary.updateErrorCount += 1
+              this._summary.errors.push(error)
+              // eslint-disable-next-line max-len
+              const msg = 'Update error occured but ignored. See summary for details'
+              this.logger.warn(msg)
```
hisabimbola: Let us log the error here also with `this.logger`. Very useful is logs here is being transported to another medium

packages/discount-code-importer/src/main.js
junajan: So it always tries to process `this.batchSize` count of requests in parallel?

packages/discount-code-importer/src/main.js
hisabimbola: and not warn, but maybe `logger.error` depending on usage

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+              return Promise.resolve()
+            }
+            this._summary.updateErrorCount += 1
+            this._summary.errors.push(error)
```
hisabimbola: we should log here also...

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+        summary: this._summary,
+      }),
+      )
+  }
+
+  _createOrUpdate (newCodes: CodeDataArray, existingCodes: CodeDataArray) {
+    return Promise.map(newCodes, (newCode: CodeData) => {
+      const existingCode = _.find(existingCodes, ['code', newCode.code])
+      if (existingCode)
+        return this._update(newCode, existingCode)
```
hisabimbola: we can add a log that we are updating here and creating down...

hisabimbola: It will be more `verbose` logs though or maybe `info`

packages/discount-code-importer/src/main.js
```diff
@@ -1,3 +1,221 @@
...
+          this._summary.errors.push(error)
+          return Promise.reject(error)
+        })
+    })
+  }
+
+  _update (newCode: CodeData, existingCode: CodeData) {
+    const actions = this.syncDiscountCodes.buildActions(newCode, existingCode)
+    // don't call API if there's no update action
+    if (!actions.length)
```
hisabimbola: let's have a field in sumaryReport about data that have no update, maybe something like `ignoredUpdates`

packages/discount-code-importer/src/main.js
junajan: I was confused because of those two different forms in the same sentence "were updates" and "were unchanged".

packages/discount-code-importer/src/main.js
hisabimbola: I think we already have `unchanged` we can use that for this...

packages/discount-code-importer/src/main.js
adnasa: yes of course, but IMO you shouldn't care that much about performance unless the predicate to build here _can_ be up to a hundreds of fields :)

but all good 👍

packages/discount-code-importer/src/main.js
wizzy25: @junajan Yes exactly

packages/discount-code-importer/package.json
```diff
@@ -20,8 +20,20 @@
   "contributors": [
     "Williams Omayuku <williams.omayuku@commercetools.com>"
   ],
-  "main": "lib/index.js",
+  "main": "./lib/index.js",
```
wizzy25: Yes you have, and no it doesn't... Just for consistency; it's about 50:50 in the `monorepo` packages for and against

packages/discount-code-importer/src/main.js
wizzy25: That's the current implementation. unchanged gets incremented if there's no update action

packages/discount-code-importer/src/main.js
wizzy25: I don't think we do. We currently use `npmlog` in our other repos for this

packages/discount-code-importer/src/main.js
wizzy25: Well, it can actually get up to 500 😀

packages/discount-code-importer/src/main.js
wizzy25: I'll handle this logging in the individual functions

packages/discount-code-importer/src/main.js
wizzy25: How do you mean? I'm using `warn` here because the import doesn't fail

packages/discount-code-importer/src/main.js
junajan: What if a user set it to 500?

wizzy25: Then it sends it all at ones. I think the API can handle 500 concurrent requests. I think I will cap it at 500 though @junajan

wizzy25:
I added a limit to the `batchSize` because it's the same number used to [fetch existing codes](https://github.com/commercetools/nodejs/blob/code-importer/packages/discount-code-importer/src/main.js#L94) and this limit is 500 from the API... Kindly review again
Cc @adnasa @junajan @hisabimbola @emmenko

packages/discount-code-importer/src/main.js
junajan: Just a note - we found out that sphere-node-cli expects to call a callback once the process is finished.

wizzy25: ?? @junajan

docs/cli/discount-code-importer.md
```diff
@@ -0,0 +1,10 @@
+# Discount Code Importer
+
+Import [commercetools discount codes](https://dev.commercetools.com/http-api-projects-discountCodes.html) to the commercetools platform.
+This library is built to be used in conjunction with [sphere-node-cli](https://github.com/sphereio/sphere-node-cli)
+
+## Features
+- Import new or update existing discount codes
+- Import codes as a valid JSON format
+
+## Configuration
```
wizzy25: No. There's a separate PR for that #181

PR #177, opened 2017-05-26
feat(product/variants): add support for key field on variants

tdeekens:

The API supports a key and sku field for product variants. The sku field is supported as an action
on the sync actions they key field not so far.

#### Summary

Adds the setProductVariantKey action to be supported on the master variant as any other variant.

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/sync-actions/test/product-sync-variants.spec.js
+++ b/packages/sync-actions/test/product-sync-variants.spec.js
@@ -222,6 +222,7 @@ describe('Actions', () => {
 
   it('should build `addVariant` action', () => {
     const newVariant = {
+      key: 'ddd',
       sku: 'ccc',
       attributes: [{ name: 'color', value: 'red' }],
       images: [{ url: 'http://foo.com', label: 'foo' }],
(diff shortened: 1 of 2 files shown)


packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -345,6 +350,27 @@ describe('Actions', () => {
...
+    const now = {
+      id: '123',
+      masterVariant: {
+        key: '', attributes: [{ name: 'foo', value: 'bar' }],
+      },
+    }
+
+    const actions = productsSync.buildActions(now, before)
+    expect(actions).toEqual([
+      { action: 'setProductVariantKey', key: null, variantId: 1 },
```
emmenko: Nitpick: I think `null` should be fine, but usually we don't set the key at all when "unsetting" a field value.

tdeekens:
Same here. Codecov just seems to lack the base to compare against.

hisabimbola:
Somehow it all passes now, I have no idea what changed though 🙄

PR #182, opened 2017-05-30
142 code generator main

### LINKED ISSUE #142: Discount code generator module
### Description

A new module to generate discount codes

### Todo

- [x] Function to generate codes
- [x] CLI tool to get input from user
- [x] Save generated codes to file
- [x] Write documentation

wizzy25:
#### Summary

As part of [this project](https://github.com/commercetools/nodejs/projects/2), implement a package to generate discount codes

#### Description

This package is now complete and should be merged and released

resolves #142


docs/cli/discount-code-generator.md
```diff
@@ -0,0 +1,155 @@
...
+  --help, -h                 Show help text.                                                     [boolean]
+  --version, -v              Show version number.                                                [boolean]
+  --quantity, -q             Quantity of discount codes to generate. (Between 1 and 500000)     [required]
+  --code-length, -l          Length of the discount codes to generate.                       [default: 11]
+  --code-prefix, -p          Prefix for each code. No prefix will be used if omitted.        [default: ""]
+  --input, -i                Path to code options CSV or JSON file.
+  --output, -o               Path to store generated output file.                      [default: "stdout"]
+  --delimiter, -d            Used CSV delimiter for input and/or output file.               [default: ","]
+  --multivalueDelimiter, -m  Used CSV delimiter in multivalue fields for input/output file. [default: ";"]
+  --logLevel,                Logging level: error, warn, info or verbose.                [default: "info"]
```
junajan: Just a nitpick - some params are written in `camelCase` and some in `kebab-case`

packages/discount-code-generator/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "commercetools",
+    "discount",
+    "code",
+    "generator"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/commercetools/nodejs/tree/master/packages/discount-code-generator"
+  },
+  "homepage": "https://commercetools.github.io/nodejs/sdk/api/discountCodeGenerator.html",
```
junajan: Is this URL correct? Shouldn't there be `cli/...`?

packages/discount-code-generator/readme.md
```diff
@@ -0,0 +1,11 @@
+# discount-code-generator
+
+Generate discount codes to be imported to the commercetools platform.
+
+https://commercetools.github.io/nodejs/docs/cli/discount-code-generator.h
```
junajan: Url should be something like this: `https://commercetools.github.io/nodejs/cli/csv-parser-price.html`

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,238 @@
...
+    if (input === undefined)
+      resolve({})
+    else if (input.path.match(/\.json$/i))
+      // Convert input stream to JSON file
+      input
+        .on('error', (error) => {
+          reject(error)
+        })
+        .on('data', (data) => {
+          _attributes = data
```
junajan: Shouldn't this be `_attributes += data` because the input file is loaded in chunks?

packages/discount-code-generator/src/cli.js
```diff
@@ -0,0 +1,238 @@
...
+        del: _args.delimiter,
+      })
+      outputStream.on('error', (error) => {
+        reject(error)
+      })
+      outputStream.end(csvOutput)
+      outputStream.on('finish', () => {
+        resolve(total)
+      })
+    } else reject(new Error('Invalid file format. Must be CSV or JSON'))
```
junajan: This is already handled when parsing input arguments.

packages/discount-code-generator/src/cli.js
wizzy25: That also works; but it's expecting just one line, so this is better; the data event gets fired once

junajan: So the input file can have only [64kb](https://github.com/nodejs/node-v0.x-archive/blob/523929c9272a53c9429616564a45f2af59670e47/lib/fs.js#L1462) including those descriptions in multiple languages.

wizzy25:
See #184

Issue #205, opened 2017-06-14
[discount-code-generator] fix `cartDiscounts` field


wizzy25:
### Description

`cartDiscounts` field is not properly generated

### Expected Behavior

should be generated as 
```js
"cartDiscounts": [
   {
      "typeId": "cart-discount",
      "id": "some-id"
   },
   {
      "typeId": "cart-discount-2",
      "id": "some-id-2"
   }
  ]
```

### Current Behavior

The field is being generated like this instead:
```js
"cartDiscounts": [ "some-id", "some-id-2" ]
```

PR #207, opened 2017-06-14
fix(discount-code-generator): fix cartDiscounts field

wizzy25:
affects: @commercetools/discount-code-generator

ISSUES CLOSED: #205

### DIFF
--- a/packages/discount-code-generator/src/cli.js
+++ b/packages/discount-code-generator/src/cli.js
@@ -144,13 +144,20 @@ const resolveInput = (_args) => {
         .on('data', (data) => {
           const arrayDelim = _args.multiValueDelimiter
           // Add condition so module doesn't fail if there are no cartDiscounts
-          if (data.cartDiscounts)
-          // eslint-disable-next-line no-param-reassign
+          if (data.cartDiscounts) {
+            // eslint-disable-next-line no-param-reassign
             data.cartDiscounts = data.cartDiscounts.split(arrayDelim)
-          _attributes += data
+            // eslint-disable-next-line no-param-reassign
+            data.cartDiscounts = data.cartDiscounts.map(cartDiscount => ({
+              typeId: 'cart-discount',
+              id: cartDiscount,
+            }),
+            )
+          }
+          _attributes += JSON.stringify(data)
         })
         .on('end', () => {
-          resolve(unflatten(_attributes))
+          resolve(unflatten(JSON.parse(_attributes)))
         })
   })
 }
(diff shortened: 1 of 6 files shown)


packages/discount-code-generator/src/cli.js
```diff
@@ -144,13 +144,20 @@ const resolveInput = (_args) => {
...
             data.cartDiscounts = data.cartDiscounts.split(arrayDelim)
-          _attributes += data
+            // eslint-disable-next-line no-param-reassign
+            data.cartDiscounts = data.cartDiscounts.map(cartDiscount => ({
+              typeId: 'cart-discount',
+              id: cartDiscount,
+            }),
+            )
+          }
+          _attributes += JSON.stringify(data)
```
junajan: @wizzy25  please can you clarify how this should work? `_attributes` is a string which can contain multiple JSON arrays? Shouldn't there be also some delimiter or maybe it should be an array instead of a string?

Because at the end you are calling `unflatten(JSON.parse(_attributes))` which will throw an error if there are concatenated multiple JSON arrays without any delimiter

What about to do this?
```js
const _attributes = []
_attributes.push(data)
_attributes.push(data)
unflatten(_attributes)

Comment (wizzy25) [inline packages/discount-code-generator/src/cli.js]: The data event is emitted only once, because the csv should contain only one row, and hence, `_attributes` should contain only one object. 
If there are multiple JSON arrays, it should throw as that would indicate an invalid file

Comment (hisabimbola) [inline packages/discount-code-generator/src/cli.js]: why are we stringifying and parsing again, why can't we just store the object in an array?

File: packages/discount-code-generator/src/cli.js [hunk 2]
```diff
@@ -144,13 +144,20 @@ const resolveInput = (_args) => {
         .on('data', (data) => {
           const arrayDelim = _args.multiValueDelimiter
           // Add condition so module doesn't fail if there are no cartDiscounts
-          if (data.cartDiscounts)
-          // eslint-disable-next-line no-param-reassign
+          if (data.cartDiscounts) {
```
hisabimbola: do you want to check if it's an array here

packages/discount-code-generator/src/cli.js
```diff
@@ -144,13 +144,20 @@ const resolveInput = (_args) => {
...
+            // eslint-disable-next-line no-param-reassign
             data.cartDiscounts = data.cartDiscounts.split(arrayDelim)
-          _attributes += data
+            // eslint-disable-next-line no-param-reassign
+            data.cartDiscounts = data.cartDiscounts.map(cartDiscount => ({
```
hisabimbola: I think we can do some like this

```
data.cartDiscounts.split(arrayDelim).map()
```

packages/discount-code-generator/src/cli.js
wizzy25: The generator expects an object, not an array. Although this implementation can also be changed. Also, I was setting it equal to the object on data, but changed it due to [this](https://github.com/commercetools/nodejs/pull/182#discussion_r119302889) review

PR #208, opened 2017-06-19
refactor(csv-parser-price): refactor logging on csv-parser-price

### LINKED ISSUE #153: [csv-parser-price] Handle logging appropriately

junajan:
Hello @hisabimbola @wizzy25 - I took our changes on csv-parser-order package and added them also here so the logging and streaming data to stdout should be handled properly now.

affects: @commercetools/csv-parser-price

Resolves #153


integration-tests/cli/csv-parser-price.it.js
```diff
@@ -119,6 +116,26 @@ describe('CSV and CLI Tests', () => {
         },
       )
     })
+
+    test('should log errors to stderr and log file', (done) => {
```
wizzy25: Are you logging to `stderr` and the logFile at the same time? I thought it's one or the other

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -119,6 +116,26 @@ describe('CSV and CLI Tests', () => {
...
+
+      // eslint-disable-next-line max-len
+      exec(`${binPath} -p ${projectKey}  -i ${csvFilePath} --logFile ${tmpFile.name}`,
+        (error, stdout, stderr) => {
+          expect(error).toBeTruthy()
+          expect(stderr).toMatch(expectedError)
+
+          fs.readFile(tmpFile.name, { encoding: 'utf8' }, (err, data) => {
+            expect(data).toContain(expectedError)
+            tmpFile.removeCallback()
```
wizzy25: This is not necessary as you did not pass the keep option on line `121`
See [here](https://www.npmjs.com/package/tmp#synchronous-file-creation)

packages/csv-parser-price/src/main.js
```diff
@@ -63,7 +63,10 @@ export default class CsvParserPrice {
       .sortBy((a, b) => a['variant-sku'].localeCompare(b['variant-sku']))
       // Limit amount of rows to be handled at the same time
       .batch(this.batchSize)
-      .stopOnError(error => this.logger.error(error))
+      .stopOnError((err) => {
+        this.logger.error(err)
+        return output.emit('error', err)
```
wizzy25: Nitpick: Unnecessary `return` statement

packages/csv-parser-price/src/main.js
```diff
@@ -75,9 +78,15 @@ export default class CsvParserPrice {
         this.logger.verbose(`Processed row ${rowIndex}`)
         rowIndex += 1
       })
-      .stopOnError(error => this.logger.error(error))
+      .stopOnError((err) => {
+        this.logger.error(err)
+        return output.emit('error', err)
```
wizzy25: Same as line 68

packages/csv-parser-price/src/main.js
```diff
@@ -75,9 +78,15 @@ export default class CsvParserPrice {
...
+      })
       .reduce({ prices: [] }, this.mergeBySku)
-      .stopOnError(error => this.logger.error(error))
+      .stopOnError((err) => {
+        this.logger.error(err)
+        return output.emit('error', err)
```
wizzy25: Same as line 68 and 83

integration-tests/cli/csv-parser-price.it.js
junajan: In the csv-parser-order package we decided to log all errors to stderr even if we use stdout for output and all messages are being logged to a log file. With this even if there are some errors, a user will see what happened without any need to open a log file.

wizzy25: > we decided to log all errors to stderr even if we use stdout 

I see that makes sense as the module has no `continueOnProblems` option. But isn't this what started the issue in the first place?

junajan: The issue was, that when we used the stdout for data output, we were printing there also other log messages so the resulting JSON was polluted and it was not possible to parse it.

Now we use stdout only for content (all log messages go to log file) and if there is any error, we stop processing and log it to stderr. 

So no more info/warn/debug/.. logs in stdout unless we output data to a file. Also as you said, we do not continueOnProblems so we can print an error message to stderr and quit processing because the output will be an invalid JSON due to an error anyway.

So the main difference is that we do not print non-error messages to stderr.

wizzy25: okay... that's fine by me... just resolve the nitpicks 👍

junajan: Gracias senore :-)

hisabimbola: Probably we change the word _log errors to stderr_. That's the final result or something, not really a log. Logging is somewhat different

packages/csv-parser-price/src/main.js
junajan: Updated here a9eeed6

packages/csv-parser-price/src/main.js
junajan: Updated here a9eeed6

packages/csv-parser-price/src/main.js
junajan: Updated here a9eeed6

integration-tests/cli/csv-parser-price.it.js
junajan: Updated here a9eeed6

integration-tests/cli/csv-parser-price.it.js
junajan: Updated here a9eeed6

hisabimbola:
Test coverage is low @junajan 

Please improve it.

junajan:
@hisabimbola working on it!

hisabimbola:
@junajan refactored the code a bit, all working now...

junajan:
Thx @hisabimbola

PR #201, opened 2017-06-12
198 discount code parser docs

wizzy25:
#### Summary

Add documentation for the [csv-parser-discount-code] #197

resolves #198

PR #197, opened 2017-06-08
Discount code parser

wizzy25:
#### Summary

Implement package to parse commercetools discount codes from CSV to JSON

This PR is for the logic. The documentation for this will be made in another PR

resolves #186

PR #210, opened 2017-06-20
feat(csv-parser-discount-code): add new package

### LINKED ISSUE #186: Discount code CSV parser
### Description

Add module to convert [commercetools discount codes](http://dev.commercetools.com/http-api-projects-discountCodes.html) from CSV to JSON

wizzy25:
#### Summary

The parser (#197) and the documentation (#201) have been individually reviewed. This is the PR that merges it all to master (and publishes)

resolves #186


docs/cli/csv-parser-discount-code.md
```diff
@@ -0,0 +1,170 @@
...
+  --output, -o               Path to output JSON file.                   [default: "stdout"]
+  --delimiter, -d            Used CSV delimiter for input and/or output file. [default: ","]
+  --multiValueDelimiter, -m  Used CSV delimiter in multiValue
+                               fields for input/output file.                  [default: ";"]
+  --continueOnProblems, -c   Flag if parsing should continue if module
+                               encounters an error.               [boolean] [default: false]
+  --logLevel, -l             Logging level: error, warn, info or verbose.  [default: "info"]
+```
+
+#### Info on flags
```
hisabimbola: Very useful section. It should increase our developer experience.

docs/cli/csv-parser-discount-code.md
```diff
@@ -0,0 +1,170 @@
...
+- The `--input` flag specifies the path to the discount codes CSV file. If this flag is omitted, the module will attempt to read the data from the `standard input`.
+- The `--output` flag specifies where to output/save the parsed discount codes as JSON file. Several notes on this flag:
+  - If the file specified already exists, it will be overwritten.
+  - The default location for status report logging is the standard output.
+  - If no output path is specified, the generated codes will be logged to the standard output as a result, status reports will be logged to a `csv-parser-discount-code.log` file
+- The `--delimiter` flag specifies the delimiter used in the input file. Defaults to `','` if omitted.
+- The `--multiValueDelimiter` flag specifies the delimiter for multiValue cells in CSV. Note that only the `cartDiscounts` field if present should contain multiple values. Defaults to `';'` if omitted.
+- The `--continueOnProblems` flag specifies if the module should continue parsing discount codes if it encounters an error. Defaults to `false` if omitted.
```
hisabimbola: Let us add some more info about errors.

Something like the error will be collated and returned in the summaryReport or anyhow we plan to return the error back to them.

docs/cli/csv-parser-discount-code.md
```diff
@@ -0,0 +1,170 @@
...
+
+### JS
+For more direct usage, it is possible to use this module directly:
+```js
+const fs = require('fs')
+const CsvParser = require('CsvParser')
```
hisabimbola: is this correct?
```
require('CsvParser')
```

_and just FYI: npm package name is always lowercase_

packages/csv-parser-discount-code/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "flat": "^2.0.1",
+    "highland": "^2.10.5",
+    "lodash": "^4.17.4",
+    "npmlog": "^4.1.0",
+    "pretty-error": "^2.1.0",
+    "tmp": "0.0.31",
+    "yargs": "^8.0.1"
+  },
+  "devDependencies": {
+    "streamtest": "^1.2.2"
```
hisabimbola: I think we should move all this devdependencies to the root `package.json`. It's something that can be useful in csv-parser modules.

packages/csv-parser-discount-code/readme.md
```diff
@@ -0,0 +1,11 @@
+# commercetools-csv-parser-discount-code
+
+Convert [commercetools discount codes](http://dev.commercetools.com/http-api-projects-discountCodes.html) CSV data to JSON
+
+https://commercetools.github.io/nodejs/cli/csv-parser-discount-code.html
```
hisabimbola: Can we add something like, _more info here_

packages/csv-parser-discount-code/src/cli.js
```diff
@@ -0,0 +1,157 @@
...
+  .option('help', {
+    alias: 'h',
+  })
+  .help('help', 'Show help text.')
+
+  .option('version', {
+    alias: 'v',
+    type: 'boolean',
+  })
+  .version('version', 'Show version number.', version)
```
hisabimbola: we can just refactor this to `.version()` and yargs will automatically fetch the version number for us from `package.json`

packages/csv-parser-discount-code/src/cli.js
```diff
@@ -0,0 +1,157 @@
...
+  })
+
+  .option('output', {
+    alias: 'o',
+    default: 'stdout',
+    describe: 'Path to output JSON file.',
+  })
+  .coerce('output', (arg) => {
+    if (arg === 'stdout') {
+      npmlog.stream = fs.createWriteStream('discountCodeGenerator.log')
```
hisabimbola: I think the file name is wrong...

packages/csv-parser-discount-code/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+
+    this._rowIndex = 0
+  }
+
+  // Remove fields with empty values from the code objects
+  static _removeEmptyFields (item: Object) {
+    return _.omitBy(item, val => val === '')
+  }
+
+  parse (input: stream$Readable, output: stream$Writable) {
```
hisabimbola: I guess _stream$Readable_ is a flow thing....🤔

packages/csv-parser-discount-code/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+  ParserSummary,
+ } from 'types/discountCodes'
+import _ from 'lodash'
+import csv from 'csv-parser'
+import JSONStream from 'JSONStream'
+import highland from 'highland'
+import npmlog from 'npmlog'
+import { unflatten } from 'flat'
+
+export default class CsvParser {
```
hisabimbola: let's make this more descriptive. Can be confusing with the `csv-parser` module above.

packages/csv-parser-discount-code/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+      })
+      .pipe(JSONStream.stringify())
+      .pipe(output)
+  }
+
+  // Convert the cartDiscounts field to an array of references to commercetools
+  // cartDiscounts
+  _cartDiscountsToArray (item: Object) {
+    if (item.cartDiscounts) {
+      // eslint-disable-next-line no-param-reassign
```
hisabimbola: Why can't be write this a pure function and not mutate the arguments, a bit messy that we have to do this everytime.

packages/csv-parser-discount-code/test/main.spec.js
```diff
@@ -0,0 +1,118 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
hisabimbola: Do we still need this in our tests, I thought we fixed it

Cc @junajan

packages/csv-parser-discount-code/test/main.spec.js
```diff
@@ -0,0 +1,118 @@
...
+import streamtest from 'streamtest'
+import fs from 'fs'
+import path from 'path'
+import CsvParser from '../src/main'
+
+describe('CsvParser', () => {
```
hisabimbola: let's change the name also

packages/csv-parser-discount-code/readme.md
wizzy25: This is the same template used in **all** modules in the monorepo. Should we also update all others?

packages/csv-parser-discount-code/src/main.js
wizzy25: Yeah... `ReadableStream` and `WritableStream` classes do not have the `.emit()` (according to flow), so this was the only way

docs/cli/csv-parser-discount-code.md
```diff
@@ -0,0 +1,170 @@
+# CSV Parser Discount Code
+
+Convert [commercetools discount codes](http://dev.commercetools.com/http-api-projects-discountCodes.html) CSV data to JSON. See example below for CSV format, sample response and usage
```
junajan: Missing `.` at the end of the sentence?

packages/csv-parser-discount-code/readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Install
+
+```bash
+npm install --save @commercetools/csv-parser-discount-code
```
junajan: In the main documentation, you have also a `--global` flag - should we add it also here?

packages/csv-parser-discount-code/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+      .through(csv({
+        separator: this.delimiter,
+        strict: true,
+      }))
+      .map(CsvParser._removeEmptyFields)
+      .map(unflatten)
+      .map(this._cartDiscountsToArray)
+      .errors(this._handleErrors) // <- Pass errors to errorHandler
+      .stopOnError((error) => { // <- Emit error and close stream if needed
+        // this.logger.error(error.message)
```
junajan: This is not needed anymore, right?

packages/csv-parser-discount-code/test/main.spec.js
junajan: Yes, it got fixed by merging csv-parser-order package to master. Please try to update this branch and check if still needed now.

packages/csv-parser-discount-code/readme.md
wizzy25: I think we should leave it as it is. Generally, the use-case will determine how it will be installed.
And it's so in other packages

packages/csv-parser-discount-code/test/main.spec.js
wizzy25: Not fixed here after updating the branch

junajan: I will try it

junajan: It looks like the `import/no-extraneous-dependencies` eslint  plugin does not work with parent package.json - it only checks the nearest one. I tried to move `streamtest` devDependency to `/packages/csv-parser-discount-code/package.json` and it stopped complaining. But when the devDep is in the main package.json, it complains with the message:
```eslint
  packages/csv-parser-discount-code/test/main.spec.js:1:1
  ✖    1:1   streamtest should be listed in the project's dependencies. Run npm i -S streamtest to add it  import/no-extraneous-dependencies
```
Also, here is the issue on the plugins github page:
https://github.com/benmosher/eslint-plugin-import/issues/458

packages/csv-parser-discount-code/readme.md
hisabimbola: yeah, and we can start the change from here...

packages/csv-parser-discount-code/readme.md
hisabimbola: I lean more towards adding a `--global` flag, this is a CLI flag so makes sense to add `--global` but no hard opinion on this, I am fine with as is

wizzy25: Well then, we're going global!!

PR #232, opened 2017-07-11
Code exporter

wizzy25:
#### Summary

Add a module to export commercetools discount codes from the CTP as csv or json

#### Description

This is the first part and PR of the work. The documentation as well as the integration tests will be implemented in a separate PR


packages/discount-code-exporter/package.json
```diff
@@ -20,7 +20,25 @@
     "Williams Omayuku <williams.omayuku@commercetools.com>"
   ],
   "main": "lib/index.js",
+  "bin": {
+    "discount-code-exporter": "bin/discount-code-exporter.js"
```
hisabimbola: Can we make the executable command one word

`discountcodeexporter` or shorter if possible

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+    const logger = {
+      error: npmlog.error.bind(this, ''),
+      warn: npmlog.warn.bind(this, ''),
+      info: npmlog.info.bind(this, ''),
+      verbose: npmlog.verbose.bind(this, ''),
+    }
+    return new DiscountCodeExport(constructorOptions, logger)
+  })
+  .then(discountCodeExport => discountCodeExport.run(args.output))
```
hisabimbola: Let's add a `.catch`

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+
+    const defaultOptions = {
+      accessToken: '',
+      batchSize: 500,
+      delimiter: ',',
+      exportFormat: 'json',
+      multiValueDelimiter: ';',
+      predicate: '',
+    }
+    Object.assign(this, defaultOptions, options)
```
hisabimbola: let's use the object spread format here

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+      data.body.results.forEach((codeObj) => {
+        output.write(codeObj)
+      })
+      return Promise.resolve()
+    }, { accumulate: false })
+  }
+
+  _buildRequest () {
+    const service = this._createService()
+    const uri = service.perPage(this.batchSize)
```
hisabimbola: little nitpick about the variable namings. what is stored here is a _method_ not an _uri_ 🤔

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+      method: 'GET',
+    }
+    if (this.accessToken)
+      request.headers = {
+        Authorization: `Bearer ${this.accessToken}`,
+      }
+    return request
+  }
+
+  _createService () {
```
hisabimbola: any benefit of having this method separately?

looks okay though

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+    return createRequestBuilder({
+      projectKey: this.apiConfig.projectKey,
+    }).discountCodes
+  }
+
+  _processCode (data: CodeData) {
+    // Use this function to make the `cartDiscounts`
```
hisabimbola: according to you, comments should be outside the function (if possible) 😜

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+    // field compatible with  the importer
+    const cartDiscounts = data.cartDiscounts.reduce((acc, discount) => {
+      if (!acc)
+        return discount.id
+      return `${acc}${this.multiValueDelimiter}${discount.id}`
+    }, '')
+    const newCodeObj = Object.assign({}, data, { cartDiscounts })
```
hisabimbola: object spread here too. This is for consistency.

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
-import findSum from '../src/main'
+// eslint-disable-next-line import/no-extraneous-dependencies
```
hisabimbola: adding the dependency to the module's `package.json` fixes this

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+        cartDiscounts: [{ id: 'cart-discount-1' }, { id: 'cart-discount-2' }],
+      }
+      const spy = jest
+        .spyOn(codeExport, '_fetchCodes')
+        .mockImplementation((jsonStream) => {
+          jsonStream.write(sampleCode)
+          return Promise.resolve()
+        })
+      const outputStream = streamtest['v2'].toText((error, result) => {
+        const expectedResult = [{ ...sampleCode }]
```
hisabimbola: yep, `object spread` is sexy 👯

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+      codeExport.run(outputStream)
+    })
+
+    it('should fetch codes and output json to stream by default', (done) => {
+      const sampleCode = {
+        code: 'discount-code',
+        name: { en: 'some-discount-name' },
+        cartDiscounts: [{ id: 'cart-discount-1' }, { id: 'cart-discount-2' }],
+      }
+      const spy = jest
```
hisabimbola: actually instead of `spyOn` you can just replace the method.

I used `spyOn` on methods that on the `class` not an instance of the class.

hisabimbola: Same goes for other implementations

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+  describe('::_processCode', () => {
+    const sampleCodeObj = {
+      name: { en: 'English', de: 'German' },
+      cartDiscounts: [{
+        typeId: 'cart-discount',
+        id: 'discount-id-1',
+      }],
+    }
+
+    it('should flatten object and return `cartDiscounts` as id string', () => {
```
hisabimbola: _as id string_? meant _a string_?

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+      }
+      const actual = codeExport._buildRequest()
+      expect(actual).toEqual(expected)
+    })
   })
 
-  it('should return the sum of two numbers', () => {
-    expect(findSum(3, 5)).toBe(8)
+  describe('::_processCode', () => {
+    const sampleCodeObj = {
```
hisabimbola: since you are reusing this variable, can we wrap it in a `beforeEach` to prevent shared state between tests...

packages/discount-code-exporter/package.json
adnasa: in terms of command, I recommend that @wizzy25 keeps this

packages/discount-code-exporter/src/main.js
adnasa: should we really mutate the constructor this way?
I would have really preferred that we explicit set it one option at a time instead of assign the whole object.

With this approach, it would be possible for users to do the following

```
const discountCodeExport = new DiscountCodeExport({ run: () => {} /* overrides the instance run method */ })
```

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+    const request = this._buildRequest()
+    return this.client.process(request, (data) => {
+      if (data.statusCode !== 200)
+        return Promise.reject(data)
+      this.logger.verbose(`Successfully exported ${data.body.count} codes`)
+      data.body.results.forEach((codeObj) => {
+        output.write(codeObj)
+      })
+      return Promise.resolve()
+    }, { accumulate: false })
```
adnasa: formatting...

```
process(
  request,
  (data) => {},
  {
    accumulate: false,
  }
)
```

packages/discount-code-exporter/src/main.js
adnasa: this was for performance reasons, right? 😉 😜

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,158 @@
...
+  _buildRequest () {
+    const service = this._createService()
+    const uri = service.perPage(this.batchSize)
+    if (this.predicate)
+      uri.where(this.predicate)
+    const request: Object = {
+      uri: uri.build(),
+      method: 'GET',
+    }
+    if (this.accessToken)
```
adnasa: Doesn't it make sense to _require_ `accessToken` already on instantiation?
then this if-statement would no longer be needed

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+    warn: () => {},
+    info: () => {},
+    verbose: () => {},
+  }
+
+  let codeExport
+  beforeEach(() => {
+    codeExport = new DiscountCodeExport({
+      apiConfig: {
+        projectKey: 'asafaelhn',
```
adnasa: I'd rather see a meaningful `projectKey` than junk even if it doesn't make any actual request

`projectKey: 'random-project-key'`

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -1,11 +1,220 @@
...
+    }, logger)
+  })
+
+  describe('::constructor', () => {
+    it('should be a function', () => {
+      expect(typeof DiscountCodeExport).toBe('function')
+    })
+
+    it('should set default properties', () => {
+      expect(codeExport.apiConfig).toBeDefined()
```
adnasa: can we please be explicit with our assertions?
`toBeDefined` is ambiguous

packages/discount-code-exporter/test/main.spec.js
adnasa: why do you need to spread here?
`expectedResult = [sampleCode]` should be enough?

packages/discount-code-exporter/test/main.spec.js
wizzy25: Yeah, but isn't that anti-pattern?

packages/discount-code-exporter/test/main.spec.js
wizzy25: No. It should return only the id of the `cartDiscount` as a string, I don't know how to write it more succinct without `max-len` error

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+if (args.output === process.stdout)
+  npmlog.stream = fs.createWriteStream(args.logFile)
+
+// Register error listener
+args.output.on('error', errorHandler)
+
+resolveCredentials(args)
+  .then((credentials) => {
+    const apiConfig = {
+      host: args.authUrl,
```
wizzy25: Yes. This works :-)

packages/discount-code-exporter/src/main.js
wizzy25: The accessToken is not necessary if `clientID` and `clientSecret` are passed

packages/discount-code-exporter/package.json
wizzy25: hyphenating makes it more readable IMO

packages/discount-code-exporter/src/main.js
wizzy25: It looked quite clunky and difficult to read when it was all together

packages/discount-code-exporter/src/cli.js
adnasa: but is `authUrl`really `host`?
when we talk about host in CTP, we mean `region` (EU or US). if this "host" as in the hostUrl, maybe it should be renamed accordingly?

WDYT?

packages/discount-code-exporter/package.json
adnasa: #userfriendly... I think it is important that we make this decision __once__ so we don't deviate on future packages

packages/discount-code-exporter/src/cli.js
wizzy25: The `host` here is the authentication url. You can see it [here](https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html#named-arguments-options)
Renaming it would be a breaking change on the sdk which we wouldn't want to do

packages/discount-code-exporter/test/main.spec.js
wizzy25: Not in every case, but sometimes the ambiguity is a necessary evil 😃

packages/discount-code-exporter/src/cli.js
hisabimbola: On the contrary, we should *always* want to break change if needed.

wizzy25: I would disagree with `always` here, especially as I don't think this is `needed`

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+
+  .option('projectKey', {
+    alias: 'p',
+    describe: 'API project key.',
+    demand: true,
+  })
+
+  .option('where', {
+    alias: 'w',
+    describe: 'specify where predicate',
```
junajan: `specify where predicate` -> `Specify where predicate.`

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+  })
+
+  .option('exportFormat', {
+    alias: 'f',
+    describe: 'Format for export',
```
junajan: `Format for export.`

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+    choices: ['csv', 'json'],
+    default: 'json',
+  })
+
+  .option('batchSize', {
+    alias: 'b',
+    describe: 'Number of codes to exports in a chunk',
```
junajan: Number of codes to exports in a chunk.

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+  })
+
+  .option('exportFormat', {
+    alias: 'f',
+    describe: 'Format for export',
+    choices: ['csv', 'json'],
+    default: 'json',
+  })
+
+  .option('batchSize', {
```
junajan: Is this parameter needed?

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,165 @@
...
+        createAuthMiddlewareForClientCredentialsFlow(this.apiConfig),
+        createUserAgentMiddleware({
+          libraryName: pkg.name,
+          libraryVersion: pkg.version,
+        }),
+        createHttpMiddleware({ host: this.apiConfig.apiUrl }),
+      ],
+    })
+
+    this.accessToken = options.accessToken || ''
```
junajan: Use defaults here?

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,165 @@
...
+      warn: () => {},
+      info: () => {},
+      verbose: () => {},
+      ...logger,
+    }
+
+    this._processCode = this._processCode.bind(this)
+  }
+
+  run (outputStream: stream$Writable) {
```
junajan: Could you please refactor and maybe also break this method to smaller ones so it will reuse the common code for CSV and JSON export types? This part looks very similar on both types:
```js
jsonStream.pipe(outputStream)
      this._fetchCodes(jsonStream)
      .then(() => {
        this.logger.info('Operation completed successfully')
        if (outputStream !== process.stdout)
          jsonStream.end()
      })
      .catch((e: Error) => {
        outputStream.emit('error', e)
      })
```

packages/discount-code-exporter/src/cli.js
wizzy25: Not really "needed", but it's a nice to have and can simply be ignored

packages/discount-code-exporter/src/main.js
wizzy25: I'd refer to a comment from @adnasa:
>should we really mutate the constructor this way?
I would have really preferred that we explicit set it one option at a time instead of assign the whole object.
With this approach, it would be possible for users to do the following
```
const discountCodeExport = new DiscountCodeExport({ run: () => {} /* overrides the instance run method */ })
```

hisabimbola: is having an empty string default for accessToken really necessary?

packages/discount-code-exporter/src/main.js
```diff
@@ -1,3 +1,165 @@
...
+        createUserAgentMiddleware({
+          libraryName: pkg.name,
+          libraryVersion: pkg.version,
+        }),
+        createHttpMiddleware({ host: this.apiConfig.apiUrl }),
+      ],
+    })
+
+    this.accessToken = options.accessToken || ''
+    this.batchSize = options.batchSize || 500
```
hisabimbola: A more elegant solution is to store all this in an object and you can spread on them

packages/discount-code-exporter/src/cli.js
```diff
@@ -0,0 +1,163 @@
...
+
+const resolveCredentials = (_args) => {
+  if (_args.accessToken)
+    return Promise.resolve({})
+  return getCredentials(_args.projectKey)
+}
+
+// If the stdout is used for a data output, save all logs to a log file.
+if (args.output === process.stdout)
+  npmlog.stream = fs.createWriteStream(args.logFile)
```
hisabimbola: we can actually move this here: https://github.com/commercetools/nodejs/pull/232/files#diff-67566d93ac5d91fd846de48cd0c99a00R34

packages/discount-code-exporter/src/main.js
wizzy25: No not really. Any falsy value will do. Just to keep it consistent because it's expecting a string (and to keep flow happy)

packages/discount-code-exporter/src/cli.js
wizzy25: Added here: 6986c4af8bb1b851093c080ee3d1f358d49b4f4c

packages/discount-code-exporter/src/main.js
wizzy25: How would you write this? I can't seem to get it (The logger in your reference is one object which is different in this case)

packages/discount-code-exporter/src/cli.js
wizzy25: We should allow the args to be parsed first because the logFile is also part of the arguments

packages/discount-code-exporter/src/main.js
wizzy25: Done here: d7c18993f77056341a4773e45906a54f17396c7d

packages/csv-parser-discount-code/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools/csv-parser-discount-code",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
hisabimbola: why are we manually changing this?

wizzy25: Wasn't changed manually, but here 9b294a5c7ef3dda5d34c79a81aa5c7d9b5242bbb. Probably shows so due to the rebasing

hisabimbola: Okay, I feel the commit history might be messed up but let's see how this goes

PR #237, opened 2017-07-12
fix(cli-modules): fix error handler

### LINKED ISSUE #235: [cli-modules] fix error handler
### Description

Fix buggy error handling implementation where errors are not handled if output is a file

wizzy25:
affects: @commercetools/csv-parser-orders, @commercetools/csv-parser-price

resolves #235


packages/csv-parser-orders/src/cli.js
```diff
@@ -104,11 +104,10 @@ const logError = (error) => {
 const errorHandler = (errors) => {
   // print errors to stderr if we use stdout for data output
```
hisabimbola: the comments are not needed anymore

wizzy25:
@adnasa we (sadly) don't have unit tests for the cli files 😭😭😭

wizzy25:
@hisabimbola setting `npmlog` by default to `stdout` means **all** logs are written to the stdout, even error logs (instead of `stderr`) 
I personally don't like this idea... or what do you think? 
Cc @junajan

junajan:
I think too that error logs do not belong to stdout .. but stderr instead.

wizzy25:
@junajan Which is quite frustrating, as `npmlog` pushes all logs to a single stream, it's kind of difficult to push info logs to `stdout` and error logs to `stderr` this logger (AFAIK)

hisabimbola:
I do not understand what you mean. The errors are manually logged to `stderr`. Sample [here ](https://github.com/commercetools/nodejs/blob/master/packages/inventories-exporter/src/cli.js#L106) and [here](https://github.com/commercetools/nodejs/blob/master/packages/inventories-exporter/src/cli.js#L108) and [here](https://github.com/commercetools/nodejs/blob/master/packages/discount-code-exporter/src/cli.js#L108)

wizzy25:
@hisabimbola They are apparently also written by the logger [here](https://github.com/commercetools/nodejs/blob/master/packages/csv-parser-price/src/main.js#L67), [here](https://github.com/commercetools/nodejs/blob/master/packages/csv-parser-price/src/main.js#L82), [here](https://github.com/commercetools/nodejs/blob/master/packages/csv-parser-orders/src/parsers/abstract-parser.js#L35) and [here](https://github.com/commercetools/nodejs/blob/master/packages/csv-parser-orders/src/parsers/abstract-parser.js#L46)

PR #242, opened 2017-07-17
240 importer token

wizzy25:
#### Summary

add support for access token

BREAKING CHANGE: on successful import, the returned promise does not resolve to the summary. The `summaryReport()` must be called directly if the summary is required.

### DIFF
--- a/docs/cli/discount-code-importer.md
+++ b/docs/cli/discount-code-importer.md
@@ -8,7 +8,7 @@ This package is built to be used in conjunction with [sphere-node-cli](https://g
 The constructor accepts two arguments:
 - A required object containing the following values:
   - `apiConfig` (Object): `AuthMiddleware` options for authentication on the commercetools platform. (Required. See [here](https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html#named-arguments-options))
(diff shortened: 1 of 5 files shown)


docs/cli/discount-code-importer.md
```diff
@@ -184,6 +186,7 @@ const discountCodeImport = new DiscountCodeImport(options)
 
 discountCodeImport.run(codes)
   .then(() => {
+  discountCodeImport.summaryReport()
```
junajan: Could you please add here how does the `summaryReport` looks like?

packages/discount-code-importer/src/main.js
```diff
@@ -98,15 +100,13 @@ export default class DiscountCodeImport {
...
-        method: 'GET',
-      })
+      const req = this._buildRequest(uri, 'GET')
+      return this.client.execute(req)
         .then(({ body: { results: existingCodes } }: Object) => (
           this._createOrUpdate(codeObjects, existingCodes)
         ))
     })
-      .then(() => Promise.resolve(this.summaryReport()))
+      .then(() => Promise.resolve())
```
junajan: This line can be omitted, right?

wizzy25: o. it actually has to be there so the CLI knows when to call the summaryReport

PR #245, opened 2017-07-18
fix(csv-parser-discount-code): properly cast types

### LINKED ISSUE #244: [csv-parser-discount-code] fix json format
### Description

Properly cast boolean and number types, as they are currently cast to strings

wizzy25:
affects: @commercetools/csv-parser-discount-code

resolves #244

### DIFF
--- a/packages/csv-parser-discount-code/test/utils.spec.js
+++ b/packages/csv-parser-discount-code/test/utils.spec.js
@@ -1,0 +1,58 @@
+import castTypes from '../src/utils'
+
+describe('castTypes', () => {
+  test('should properly cast object types', () => {
+    const sample = {
+      name: {
+        de: 'Valerian',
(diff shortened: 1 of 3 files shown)


packages/csv-parser-discount-code/test/utils.spec.js
```diff
@@ -0,0 +1,58 @@
+import castTypes from '../src/utils'
+
+describe('castTypes', () => {
+  test('should properly cast object types', () => {
```
hisabimbola: why not using `it`?

wizzy25: Because they mean (and do) the same thing... And you could have as well asked me the reverse question if I used `it`... (Besides, that's more like Integration test 😜)

PR #250, opened 2017-07-20
Code exporter Integration test

wizzy25:
#### Summary

Add integration tests for the discount code importer package
Refactor code to fix errors noticed during integration testing

### DIFF
--- a/integration-tests/cli/discount-codes.it.js
+++ b/integration-tests/cli/discount-codes.it.js
@@ -1,15 +1,12 @@
 import fs from 'fs'
+import tmp from 'tmp'
 import path from 'path'
-import Promise from 'bluebird'
-import {
-  createAuthMiddlewareForClientCredentialsFlow,
-} from '@commercetools/sdk-middleware-auth'
-import { createClient } from '@commercetools/sdk-client'
-import { createRequestBuilder } from '@commercetools/api-request-builder'
-import { createHttpMiddleware } from '@commercetools/sdk-middleware-http'
+import isuuid from 'isuuid'
+import csv from 'csvtojson'
+import { oneLine } from 'common-tags'
 import { exec } from 'child_process'
 import { getCredentials } from '@commercetools/get-credentials'
-import { clearData } from './helpers/utils'
+import { createData, clearData } from './helpers/utils'
 import DiscountCodeImport from '../../packages/discount-code-importer/src/main'
 
 let projectKey
@@ -20,8 +17,6 @@ else
 
 describe('DiscountCode tests', () => {
   let apiConfig
-  let service
-  let client
   let cartDiscount
 
   const logger = {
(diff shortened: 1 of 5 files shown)


integration-tests/cli/discount-codes.it.js
```diff
@@ -40,77 +35,38 @@ describe('DiscountCode tests', () => {
...
-        body: cartDiscountDraft,
-      }
-      // Create cart-discount
-      return client.execute(req)
+      // Wrap in an array because the util function expects an array
+      return createData(apiConfig, 'cartDiscounts', [cartDiscountDraft])
     })
     .then((data) => {
-      cartDiscount = data.body
+      // console.error(data);
```
hisabimbola: thing you want to remove it?

integration-tests/cli/discount-codes.it.js
```diff
@@ -40,77 +35,37 @@ describe('DiscountCode tests', () => {
...
-        }
-        return client.execute(deleteCartRequest)
-      })
-      .catch(process.stderr.write)
-  })
+  // Delete Discount codes
+  afterAll(() => clearData(apiConfig, 'discountCodes')
+    // Delete cart discounts
+    .then(() => clearData(apiConfig, 'cartDiscounts'))
+    .catch(process.stderr.write),
```
junajan: Is it ok to swallow errors here? The tests will pass even if there is something that cannot be deleted? Btw nice refactor 👍

packages/discount-code-exporter/src/main.js
```diff
@@ -153,17 +153,30 @@ export default class DiscountCodeExport {
   }
 
   // Use this method to make the `cartDiscounts`
-  // field compatible with  the importer
+  // field compatible with the importer
   _processCode (data: CodeData): Object {
-    const cartDiscounts = data.cartDiscounts.reduce((
+    const newCodeObj = data
```
junajan: Why do you create a new variable `newCodeObj `?
`_processCode (newCodeObj: CodeData): Object {` Should be enought?

integration-tests/cli/discount-codes.it.js
```diff
@@ -277,4 +247,116 @@ describe('DiscountCode tests', () => {
...
+          fs.readFile(csvFilePath, { encoding: 'utf8' }, (error, data) => {
+            expect(error).toBeFalsy()
+            csv().fromString(data)
+              .on('json', (jsonObj) => {
+                expect(jsonObj).toMatchObject(expected)
+                expect(isuuid(jsonObj.id)).toBe(true)
+                expect(jsonObj.createdAt).toMatch(UTCDateTimeRegex)
+                expect(jsonObj.lastModifiedAt).toMatch(UTCDateTimeRegex)
+              })
+              .on('done', () => done())
```
junajan: Some error handler would be lovely here: `.on('error', done)`

integration-tests/cli/discount-codes.it.js
wizzy25: @junajan This is after all the test cases have run. And knowing how jest handles rejected promises, not catching it here will be a big problem

packages/discount-code-exporter/src/main.js
wizzy25: just avoiding `eslint-no-parameter-reassign`

integration-tests/cli/discount-codes.it.js
wizzy25: This will make failing tests pass, as the handler would simply call done on errors and the assertions would just be ignored, passing the tests :-)

junajan: So if the error will be raised as an exception then it is ok I guess..

packages/discount-code-exporter/src/main.js
junajan: Should we disable that rule instead of evading it?

integration-tests/cli/discount-codes.it.js
wizzy25: It will throw out and the test case will fail, if there is no error listener. Jest generally handles asynchronous test suites not very graciously

packages/discount-code-exporter/src/main.js
wizzy25: Well, that could be a matter of convenience... cc @hisabimbola

PR #266, opened 2017-08-02
252 price export

### LINKED ISSUE #252: [price export] add new package for price export
### Description

Add a new package to export price data from the CTP.

### Expected Behavior

New package should support custom fields

wizzy25:
#### Summary

Add a new module to export price data from the commercetools platform as CSV or JSON

resolves #252


integration-tests/cli/price-export.it.js
```diff
@@ -0,0 +1,210 @@
...
+        credentials,
+      }
+      return clearData(apiConfig, 'products')
+    })
+    .then(() => clearData(apiConfig, 'types'))
+    .then(() => clearData(apiConfig, 'channels'))
+    .then(() => clearData(apiConfig, 'customerGroups'))
+    .then(() => clearData(apiConfig, 'productTypes'))
+    .then(() => createData(apiConfig, 'productTypes', [sampleProductType]))
+    .then((data) => {
```
adnasa: lots of `clearData` here ☝🏽
Is this really needed as a `before`?
Would it make it worse, if we put it on `afterAll`?

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+import { description } from '../package.json'
+
+process.title = 'price-exporter'
+
+const args = yargs
+  .usage(
+    `
+Usage: $0 [options]
+${description}`,
+  )
```
adnasa: Weird formatting.. anything that `common-tags` can help out with?

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+    process.stderr.write(`ERR: ${errorFormatter.render(error)}`)
+  else
+    process.stderr.write(`ERR: ${error.message || error}`)
+}
+
+// print errors to stderr if we use stdout for data output
+// if we save data to output file errors are already logged by npmlog
+const errorHandler = (errors) => {
+  if (Array.isArray(errors))
+    errors.forEach(logError)
```
adnasa: I don't like this approach due to misleading bindings..

suggested approach
```js
errors.forEach((e) => { logError(e) })
```

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+    logError(errors)
+
+  process.exitCode = 1
+}
+
+const getHeaders = _args => (
+  new Promise((resolve, reject) => {
+    if (!_args.input)
+      resolve()
+    const rl = readline.createInterface({
```
adnasa: minifying original source code won't help in the future.
can we find a better name, please?

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+    })
+    rl.on('error', reject)
+    rl.on('line', (line) => {
+      rl.close()
+      resolve(line.split(_args.delimiter))
+    })
+  })
+)
+
+const resolveCredentials = (_args) => {
```
adnasa: `_args` is missing a representation on what the argument really is and should be.
A proper naming atleast helps the next developer understand the representation of this.

maybe `credentials` would suffice?

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+    return resolveCredentials(args)
+  })
+  .then((credentials) => {
+    const apiConfig = {
+      host: args.authUrl,
+      apiUrl: args.apiUrl,
+      projectKey: args.projectKey,
+      credentials,
+    }
+    const constructorOptions = {
```
adnasa: This name is too generic. I would prefer `priceExportOptions`, which would make the following line completely transparent to the representation

```
new PriceExporter(priceExporterOptions...)
```

integration-tests/cli/price-export.it.js
wizzy25: It's in before AND after. Ensures everything is clean before it starts. Jest Integration tests don't fail gracefully and on failure, the `afterAll` block is not always executed (learned this the hard way from the discount-code-exporter)

packages/price-exporter/src/cli.js
wizzy25: How is it misleading?

packages/price-exporter/src/cli.js
wizzy25: Hmmmm... I could, but this (variable name ) is also on the official [node documentation](https://nodejs.org/dist/latest-v6.x/docs/api/readline.html)

wizzy25: `_args` is just a representation of `args` from `yargs` 😌

packages/price-exporter/src/cli.js
wizzy25: [here](https://github.com/commercetools/nodejs/pull/266#discussion_r130964262)

packages/price-exporter/bin/price-exporter.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+/* eslint-disable */
```
adnasa: what do we want to disable here?
if it is about the max-len on L5, this can be easily fixed. There is no need to disable eslint, however trivial the need

packages/price-exporter/package.json
```diff
@@ -21,8 +21,31 @@
...
+  ],
+  "dependencies": {
+    "@commercetools/api-request-builder": "^3.1.1",
+    "@commercetools/get-credentials": "^2.1.0",
+    "@commercetools/sdk-client": "^1.4.1",
+    "@commercetools/sdk-middleware-auth": "^3.1.0",
+    "@commercetools/sdk-middleware-http": "^2.1.1",
+    "@commercetools/sdk-middleware-user-agent": "^1.2.0",
+    "JSONStream": "^1.3.1",
+    "bluebird": "^3.5.0",
```
adnasa: oh, we're still using bluebird? I didn't know that 😄 

---

while we are here, I'm not entirely sure that we should continue with ranged version on `dependencies`

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+        serviceType = 'types'
+        break
+      case 'customerGroup':
+        serviceType = 'customerGroups'
+        break
+      case 'channel':
+        serviceType = 'channels'
+        break
+      default:
+        serviceType = type
```
adnasa: what if the `type` is not something we should support?

packages/price-exporter/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+
+import { flattenDeep } from 'lodash'
+
+// This method is necessary because of the custom fields, it is not possible
+// to know the actual headers on the fly so we would have to prepare the
+// headers before we start writing to CSV file
+export function _getHeaders (prices: Array<Object>) {
+  const headers = new Set()
+  prices.forEach((price) => {
+    if (price)
```
adnasa: looking to the signature `prices: Array<Object>`, to my understanding this would never be falsy. 
What's the reason for the if-statement?

packages/price-exporter/src/main.js
wizzy25: This is a private method, called by another private method with well-specified types

packages/price-exporter/package.json
wizzy25: Well, yes we are:
1. Presently, bluebird's `Promise` implementation is faster than the native `Promise` implementation
2. There's still no native `Promise.map`

wizzy25: This topic on range dependencies again 😂😂 cc @hisabimbola

packages/price-exporter/bin/price-exporter.js
wizzy25: semicolons, max length, we wouldn't want to lint our hashbangs.......

packages/price-exporter/src/utils.js
wizzy25: Wrong type definition. The price could be null sometimes. Thanks for this

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+    }
+    this._cache = {}
+
+    this._resolveReferences = this._resolveReferences.bind(this)
+  }
+
+  run (outputStream: stream$Writable) {
+    this.logger.info('Starting Export')
+    const request = this._buildRequest('productProjections')
+    this.client.process(request, (data) => {
```
junajan: This can be shortened with destructor.. { body: { results: products } }

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+      const products = data.body.results
+      this.logger.verbose(`Fetched ${products.length} products`)
+      let processedBatch = []
+      return Promise.map(products, product => utils._getPrices(product))
+      .then((pricesArray) => {
```
junajan: .then(pricesArray => this._handlePrices(pricesArray))

integration-tests/cli/price-export.it.js
wizzy25: @junajan Only in before is also a problem: `price-exporter` could not clear data before because `inventory-export` did not clear the inventory when done and `customTypes`/`channels` cannot be cleared whiled linked to an inventory... Had to refactor that part of the int test

packages/price-exporter/src/cli.js
wizzy25: @junajan I don't get

integration-tests/cli/inventories-exporter.it.js
```diff
@@ -36,6 +36,10 @@ describe('StockExporter CLI', () => {
     .then(() => createData(apiConfig, 'inventory', inventories))
   , 10000)
 
+  afterAll(() => clearData(apiConfig, 'inventory')
```
wizzy25: No. Inventories have to be cleared first before the types linked to them

integration-tests/cli/price-export.it.js
```diff
@@ -0,0 +1,210 @@
...
+else
+  projectKey = process.env.npm_config_projectkey
+
+describe('Price Exporter', () => {
+  let apiConfig
+  const productType = { typeId: 'product-type' }
+  const customerGroup = { typeId: 'customer-group' }
+  const channel = { typeId: 'channel' }
+  const customType = { typeId: 'type' }
+  const bin = './integration-tests/node_modules/.bin/price-exporter'
```
wizzy25: This has worked so far. It doesn't pull from npm. The module isn't even on npm and as you can see, Travis isn't complaining 😂

integration-tests/cli/price-export.it.js
```diff
@@ -0,0 +1,210 @@
...
+        projectKey,
+        credentials,
+      }
+      return clearData(apiConfig, 'products')
+    })
+    .then(() => clearData(apiConfig, 'types'))
+    .then(() => clearData(apiConfig, 'channels'))
+    .then(() => clearData(apiConfig, 'customerGroups'))
+    .then(() => clearData(apiConfig, 'productTypes'))
+    .then(() => createData(apiConfig, 'productTypes', [sampleProductType]))
```
wizzy25: Well, `products` must be created last and cleared first. I'll try to "parallelized" the others

integration-tests/cli/price-export.it.js
```diff
@@ -0,0 +1,210 @@
...
+      return createData(apiConfig, 'products', sampleProducts)
+    })
+    .catch(process.stderr.write)
+  , 10000)
+
+  afterAll(() => clearData(apiConfig, 'products')
+    .then(() => clearData(apiConfig, 'types'))
+    .then(() => clearData(apiConfig, 'channels'))
+    .then(() => clearData(apiConfig, 'customerGroups'))
+    .then(() => clearData(apiConfig, 'productTypes'))
```
wizzy25: See [here](https://github.com/commercetools/nodejs/pull/266#discussion_r131217668)

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+const errorHandler = (errors) => {
+  if (Array.isArray(errors))
+    errors.forEach(logError)
+  else
+    logError(errors)
+
+  process.exitCode = 1
+}
+
+const getHeaders = _args => (
```
wizzy25: It all depends on your POV 😜😜

integration-tests/cli/price-export.it.js
junajan: looks weird

docs/cli/price-exporter.md
```diff
@@ -0,0 +1,103 @@
...
+  - Only the first line is read and subsequent lines (if present) will be ignored
+  - The delimiter must match the delimiter passed in by `--delimiter` (or the default delimiter)
+- The `--output` flag specifies where to output/save the exported prices.
+  - If the file specified already exists, it will be overwritten.
+  - The default location for status report logging is the standard output.
+  - If no output path is specified, the exported prices will be logged to the standard output.
+- The `--delimiter` flag specifies the delimiter used in the output file if CSV. Defaults to `','` if omitted.
+- The `where` flag specifies an optional (where) query predicate to be included in the request. This predicate is on the products containing the prices and not on the prices themselves. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](http://dev.commercetools.com/http-api.html#predicates)
+- The `--staged` flag specifies the projection of the products from which the prices should be fetched.
+  - If passed `true`, prices from published and unplished products are retrieved
```
wizzy25: No no no.... some users just don't get our `staged` and `current` vernacular (trust me)

integration-tests/cli/price-export.it.js
wizzy25: That's how all the int tests work. It's a symlink to the `packages` folder

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+  _buildRequest (
+    type: string,
+    price?: Object,
+  ): ClientRequest {
+    const priceObject = { ...price }
+    let serviceType
+    let uri
+    switch (type) {
+      case 'custom':
+        serviceType = 'types'
```
junajan: Would it work if we rename "custom" type to "types" and then just do this?
`serviceType = type`

Because this switch case looks complicated and maybe it is not needed in the end.

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+    .then((response: Object): Promise<Object> => {
+      // Save resolve to cache
+      if (type === 'customerGroup') {
+        resolvedPrice[type].groupName = response.body.name
+        this._cache[price[type].id] = response.body.name
+      } else {
+        this._cache[price[type].id] = response.body.key
+        resolvedPrice[type].key = response.body.key
+      }
+      return Promise.resolve(resolvedPrice)
```
junajan: You are already in Promise so you do not need to add another `Promise.resolve` here.

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+      }
+
+      // If not in cache, resolve from API
+      const request = this._buildRequest(type, price)
+      return this.client.execute(request)
+      .then((response: Object): Promise<Object> => {
+        // Save resolved id to cache
+        this._cache[price[type].type.id] = response.body.key
+        resolvedPrice.customType = response.body.key
+        return Promise.resolve(resolvedPrice)
```
junajan: You are already in a promise so no need for another `Promise.resolve` here

integration-tests/cli/price-export.it.js
hisabimbola: The best solution is for the API to expose a method that allows creating a project on the fly. Makes testing easier.

packages/price-exporter/src/cli.js
hisabimbola: Lost 😕

docs/cli/price-exporter.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Info on flags
+- The `--input` flag specifies the path to an optional CSV template file.
+  - Only the first line is read and subsequent lines (if present) will be ignored
+  - The delimiter must match the delimiter passed in by `--delimiter` (or the default delimiter)
+- The `--output` flag specifies where to output/save the exported prices.
+  - If the file specified already exists, it will be overwritten.
+  - The default location for status report logging is the standard output.
+  - If no output path is specified, the exported prices will be logged to the standard output.
+- The `--delimiter` flag specifies the delimiter used in the output file if CSV. Defaults to `','` if omitted.
```
hisabimbola: I think this delimiter works both for output file and input file. We should specify that.

integration-tests/cli/price-export.it.js
wizzy25: Well, for now that isn't available, so we have to make do with what we have 😞

docs/cli/price-exporter.md
```diff
@@ -0,0 +1,103 @@
...
+- The `where` flag specifies an optional (where) query predicate to be included in the request. This predicate is on the products containing the prices and not on the prices themselves. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](http://dev.commercetools.com/http-api.html#predicates)
```
hisabimbola: Since we are using `product-projections` endpoint. We should specify it also

docs/cli/price-exporter.md
```diff
@@ -0,0 +1,103 @@
...
+  info: console.log,
+  verbose: console.debug,
+}
+
+const priceExporter = new PriceExporter(options, logger)
+
+// Register error listener
+outputStream.on('error', errorHandler)
+
+outputStream.on('finish', () => console.log('done with export'))
```
hisabimbola: let's use `process.stdout` instead of `console.log`

packages/price-exporter/package.json
hisabimbola: What's your thought about _ranged version on dependencies_ @adnasa

packages/price-exporter/readme.md
```diff
@@ -7,5 +7,5 @@ more info here: https://commercetools.github.io/nodejs/cli/price-exporter.html
 ## Install
 
 ```bash
-npm install --save @commercetools/price-exporter
+npm install @commercetools/price-exporter --global
```
hisabimbola: 👍

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,192 @@
...
+if (args.output === process.stdout)
+  npmlog.stream = fs.createWriteStream(args.logFile)
+else
+  npmlog.stream = process.stdout
+
+// Register error listener
+args.output.on('error', errorHandler)
+
+let csvHeaders
+getHeaders(args)
```
hisabimbola: can we document we get are getting headers in the code.

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+      ],
+    })
+
+    const defaultOptions = {
+      delimiter: ',',
+      exportFormat: 'json',
+      staged: false,
+    }
+
+    this.config = { ...defaultOptions, ...options }
```
hisabimbola: 💯

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+  }
+
+  run (outputStream: stream$Writable) {
+    this.logger.info('Starting Export')
+    const request = this._buildRequest('productProjections')
+    this.client.process(request, (data) => {
+      const products = data.body.results
+      this.logger.verbose(`Fetched ${products.length} products`)
+      let processedBatch = []
+      return Promise.map(products, product => utils._getPrices(product))
```
hisabimbola: Why did you decide to use an _utils_ approach, and not add it as a method of this class.

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+    this._resolveReferences = this._resolveReferences.bind(this)
+  }
+
+  run (outputStream: stream$Writable) {
+    this.logger.info('Starting Export')
+    const request = this._buildRequest('productProjections')
+    this.client.process(request, (data) => {
+      const products = data.body.results
+      this.logger.verbose(`Fetched ${products.length} products`)
+      let processedBatch = []
```
hisabimbola: I am not sure there is a need to declare this variable here and as a _let_

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+      return Promise.map(products, product => utils._getPrices(product))
+      .then((pricesArray) => {
+        processedBatch = this._handlePrices(pricesArray)
+        return Promise.resolve(processedBatch)
+      })
+    }, { accumulate: true })
```
hisabimbola: we should not accumulate all prices

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,270 @@
...
+    .then((allPricesBatches) => {
+      this._handleOutput(allPricesBatches, outputStream)
+    })
+    .catch((e) => {
+      outputStream.emit('error', e)
+    })
+  }
+
+  _handlePrices (productPricesArray: Array<UnprocessedPriceObject>) {
```
hisabimbola: though I understand how hand it can be to come up with method names, this method names are really very vague and does not describe what is being done in the methods.

I think we should strip all _.._handle_. that might help in coming up with better method names.

packages/price-exporter/src/cli.js
wizzy25: `common-tags` removes the blank line at the beginning, even with `\n` or an actual blank line, so I think it's better to live with this formatting

packages/price-exporter/src/cli.js
adnasa: > Lost

which IS EXACTLY what I mean with misleading bindings.
having `.forEach(logError)` leaves room for unnecessary bindings when not needed

adnasa: TLDR; __use closure__
yes: more scoped runtimes, but highly readable and eliminates __implicity__

packages/price-exporter/src/main.js
wizzy25: Renaming makes it a lot more complicated than it already is because there's a field named custom in the object

packages/price-exporter/src/main.js
wizzy25: No not really, wouldn't work if it's unwrapped because the return value is chained

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,302 @@
...
+      if (outputStream !== process.stdout)
+        pipeStream.end()
+      this.logger.info('Export operation completed successfully')
+    })
+    .catch((e) => {
+      outputStream.emit('error', e)
+    })
+  }
+
+  _flattenPrices (productPricesArray: Array<UnprocessedPriceObject>) {
```
hisabimbola: Let's check this method, not sure it's needed anymore.

packages/price-exporter/src/main.js
```diff
@@ -1,3 +1,302 @@
...
+      return Promise.resolve(this._resolveReferences(flatPrices))
+    return Promise.resolve(flatPrices)
+  }
+
+  _writePrices (
+    retrievedPrices: Array<any>,
+    pipeStream: stream$Writable,
+  ) {
+    if (this.config.exportFormat === 'csv') {
+      const flatCSVPrices = flattenDeep(retrievedPrices)
```
hisabimbola: No need for this again.

packages/price-exporter/package.json
adnasa: My opinion is that it really doesn't make any sense to range your deps if you already are relying on greenkeeper to keep them up to date.
One issue we suffered with range, is that sometimes deps break with some breaking change sneaking in

hisabimbola: Yeah, the fault is that we trust people to really know semver, but I guess we have to trust anyway

Also greenkeeper does not support the monorepo structure yet so it only works for the base package.json

packages/price-exporter/src/cli.js
hisabimbola: @adnasa can you break this down, too many grammer TBH

I am not sure what you are saying really applies in this scenario

packages/price-exporter/src/cli.js
```diff
@@ -0,0 +1,197 @@
...
+const errorHandler = (errors) => {
+  if (Array.isArray(errors))
+    errors.forEach(logError)
+  else
+    logError(errors)
+
+  process.exitCode = 1
+}
+
+// Retrieve the headers from the input file
```
hisabimbola: we should update this comment to reflect the main reason(s) why we need getHeaders method.

packages/price-exporter/package.json
adnasa: your call, buddy.
I just want to point out that ranging just kind of beats the purpose of using `greenkeeper`.

---

anyways, this is a really a discussion for another time. @wizzy25 you don't have to address this in this PR at all 😛

Issue #193, opened 2017-06-06
Commit package-lock.json


hisabimbola:
### Description

Since npm officially now uses [package-lock.json](https://docs.npmjs.com/files/package-lock.json) and it's recommended that we commit that, I think it's best we commit it to our packages.

wizzy25:
I'm not sure if we should. As we run CI tests with Node v6 and v4 (with npm v3 and v2 respectively), and personally, I develop on Node v6 and npm v3 to maintain cmpatibility...

hisabimbola:
I am not sure there will be any compatibility issue, if you do not use npm5 it will simply ignore

wizzy25:
As long as there're no compatibility issues, 😃

wizzy25:
This issue has been resolved in #286

PR #286, opened 2017-08-24
Add cart discounts sync actions

damarna85:
#### Summary
Adds a new feature in the sync actions package in order to build sync actions for the cart discounts.

#### Description
This PR also has a commit adding `package.lock` files, which we are not sure  (based on this [discussion](https://github.com/commercetools/nodejs/issues/193)) wether they should be committed or not. In case we don't need it, we can easily get rid of this [commit](https://github.com/commercetools/nodejs/commit/5c17dd3f88b1f8967a5efeafc82bb2ca25168360)

#### Todo
- Tests
    - [x] Unit


damarna85:
done @emmenko thanks!

packages/sync-actions/test/cart-discounts-sync.spec.js
```diff
@@ -0,0 +1,289 @@
+import cartDiscountsSyncFn, {
+  actionGroups,
+} from '../src/cart-discounts'
+import { baseActionsList } from '../src/cart-discounts-actions'
+
+describe('Exports', () => {
```
hisabimbola: Can we prefix this with `Cart Discounts` so tests suites are well described?

Also [here](https://github.com/commercetools/nodejs/pull/286/files#diff-ad3fd9838b0f01608285f49c706da4d7R27)

packages/sync-actions/test/cart-discounts-sync.spec.js
```diff
@@ -0,0 +1,289 @@
...
+      { action: 'setValidUntil', key: 'validUntil' },
+    ])
+  })
+})
+
+describe('Actions', () => {
+  describe('changeIsActive action', () => {
+    let actual
+    let expected
+    beforeEach(() => {
```
hisabimbola: I noticed the `beforeEach` in all the tests, and the declaring of `actual` and `expected` variables above.

While it might look _good and beneficial_. It's a bit anti-pattern to the way the tests are written in this repo.

It's good if the pattern in each project is considered when contributing, I think that's the *good thing* about patterns after all.

damarna85:
@emmenko / @hisabimbola I addressed your comments. Can you take a second look at them?

Thanks for your feedback!

PR #295, opened 2017-08-30
feat(sdk-client): limit returned items with processFn

### LINKED ISSUE #294: [sdk-client] ensure limit in processFn
### Description

Implement limit on items returned with the processFn

### Expected Behavior

Should return maximum of a set limit from the CTP

### Current Behavior

Returns limit on each request, but still sends multiple requests so the returned items are n*limit

wizzy25:
### This is a breaking change
BREAKING CHANGES: Calling `process` with a limit query now returns the `limit` number of items total and not the `limit` amount for **_"n"_** requests

resolves #294


wizzy25:
#### \*Edit\*

This is no longer a breaking change

PR #304, opened 2017-09-14
feat(sync-actions): add `changeMasterVariant` action generation

tdeekens:
#### Summary

This changes the internal workings of sync-actions to allow for automatic generation of a `changeMasterVariant` action.

#### Description

The sync-actions package treated the `masterVariant` separately when calculating update actions. This comes with a couple of down sides:

1. Code duplication across a `if (masterVariant)` `forEach(variants)`
2. It restricts the module in detecting changing the master variant 

The latter is the issue this tries to solve while also enabling the the generation of a `changeMasterVariant` update action. Previously, whenever the `masterVariant` changed the library would assume the whole variant would need creation and would create a update actions accordingly. While also creating update actions to delete the variant which actually became the new master variant.

The problem with this approach is that it is a bug and a feature at the same time. However, more of a bug than a feature in my opinion.

This feels unnecessary as it should just cover the following cases

1. `masterVariant` was changed but none of the attributes changed and the variant previously existed
2. `masterVariant` was changed some attributes changed and the variant existed previsouly
3. The `masterVariant` did not exist (must be new product) and all attributes are new

To achieve this and to minimize the amount of needed book keeping in the process I move the `masterVariant` into the `variants` array at the first position. 

1. This removes the code duplication
2. Allowing tracking the identify of the variant (existence essentially) by `id`, `name`, etc (sku and key don't seem implemented yet)
3. Easier checking if the id of the old vs new master variant actually changed 

#### Todo

- Tests
    - [x] Unit


packages/sync-actions/src/product-actions.js
```diff
@@ -61,10 +61,15 @@ export function actionsMapVariants (diff, oldObj, newObj) {
       ...newObject,
       action: 'addVariant',
     }),
-    [REMOVE_ACTIONS]: objectToRemove => ({
-      action: 'removeVariant',
-      id: objectToRemove.id,
-    }),
+    [REMOVE_ACTIONS]: ({ id }, index) => {
+      // The master variant can not be removed
```
tdeekens: We can't remove the master variant. This was done before by treating it as in the `masterVariant` property which would non run through the array detection process.

packages/sync-actions/src/product-actions.js
```diff
@@ -196,16 +180,7 @@ export function actionsMapAttributes (
...
-
-  if (masterVariant) {
-    const mActions = _buildVariantImagesAction(
-      masterVariant.images,
-      oldObj.masterVariant,
-      newObj.masterVariant,
-    )
-    actions = actions.concat(mActions)
-  }
+  const { variants } = diff
```
tdeekens: ...all this duplication of handling changes goes away.

packages/sync-actions/src/product-actions.js
```diff
@@ -256,6 +220,30 @@ export function actionsMapPrices (diff, oldObj, newObj) {
     .concat(addPriceActions)
 }
 
+export function generateChangeMasterVariantAction (oldObj, newObj) {
```
tdeekens: Specific handler to create the action.

packages/sync-actions/src/product-actions.js
```diff
@@ -256,6 +220,30 @@ export function actionsMapPrices (diff, oldObj, newObj) {
...
+  const createChangeMasterVariantAction = variantId => ({
+    action: 'changeMasterVariant',
+    variantId,
+  })
+  const extractMasterVariantId = (fromObj) => {
+    const variants = Array.isArray(fromObj.variants) ? fromObj.variants : []
+
+    return variants[0] ? variants[0].id : undefined
```
tdeekens: This feels a bit wonky but eases the code further down. Generally caused by variants not having to be set and hence the `id` not necessarily having to be present.

packages/sync-actions/src/products.js
```diff
@@ -58,10 +62,32 @@ function createProductMapActions (mapActionGroup) {
   }
 }
 
+function moveMasterVariantsIntoVariants (before, now) {
```
tdeekens: This is the hook we pass in to process data before it hits the differ.

packages/sync-actions/src/products.js
```diff
@@ -58,10 +62,32 @@ function createProductMapActions (mapActionGroup) {
...
+
 export default (config: Array<ActionGroup>): SyncAction => {
   const mapActionGroup = createMapActionGroup(config)
   const doMapActions = createProductMapActions(mapActionGroup)
-  const buildActions = createBuildActions(diffpatcher.diff, doMapActions)
+
+  const buildActions = createBuildActions(
+    diffpatcher.diff,
+    doMapActions,
+    moveMasterVariantsIntoVariants,
```
tdeekens: ...third argument of the action builder.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,13 +1,23 @@
-export default function createBuildActions (diff, doMapActions) {
+function applyOnBeforeDiff (before, now, fn) {
+  return fn && typeof fn === 'function' ? fn(before, now) : [before, now]
```
tdeekens: Eases code on call site.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,13 +1,23 @@
...
+  differ, doMapActions, onBeforeDiff,
+) {
   return function buildActions (now, before, options = {}) {
     if (!now || !before)
       throw new Error('Missing either `newObj` or `oldObj` ' +
         'in order to build update actions')
 
-    // diff 'em
-    const diffed = diff(before, now)
+    const [
```
tdeekens: This function feels generally flaky to me. It shuffles the arguments of `new` and `before` around quite a bit and I wonder if that's the intention or actually a bug which we then work around when consuming the diff.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,13 +1,23 @@
...
+      proccedBefore, processedNow,
+    ] = applyOnBeforeDiff(before, now, onBeforeDiff)
+
+    const diffed = differ(proccedBefore, processedNow)
```
tdeekens: I renamed the arg. The library seems to call a `diff` the same as the `differ` creating a `diff` which confused me quite a bit with this argument shuffling.

packages/sync-actions/src/utils/create-build-array-actions.js
```diff
@@ -23,26 +23,42 @@ export default function createBuildArrayActions (key, config) {
...
     if (diff[key]) {
       const arrayDelta = diff[key]
+
       Object.keys(arrayDelta).forEach((index) => {
-        if (config[ADD_ACTIONS] && isCreateAction(arrayDelta, index))
-          addActions.push(
-            // When adding a new element you don't need the oldObj
-            config[ADD_ACTIONS](newObj[key][index]),
+        if (config[ADD_ACTIONS] && isCreateAction(arrayDelta, index)) {
+          const actionGenerator = config[ADD_ACTIONS]
```
tdeekens: Just general simplications in naming things here. The addition is to pass the index to the callback so it has the information.

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -270,7 +270,7 @@ describe('Actions', () => {
     const now = {
       id: '123',
       masterVariant: {
-        sku: 'v1', key: 'v2', attributes: [{ name: 'foo', value: 'new value' }],
+        id: 1, sku: 'v1', key: 'v2', attributes: [{ name: 'foo', value: 'new value' }],
```
tdeekens: The test below felt like it didn't want to test for the creation of the masterVariant. However if there is no id it should be created.

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -291,42 +291,186 @@ describe('Actions', () => {
...
-        sku: 'v1',
-        attributes: [{ name: 'foo', value: 'bar' }],
-      },
-      variants: [
-        { id: 2, sku: 'v2', key: 'v2', attributes: [{ name: 'foo', value: 'qux' }] },
-        { id: 3, sku: 'v3', key: 'v3', attributes: [{ name: 'foo', value: 'baz' }] },
-      ],
-    }
+  describe('without master variant in `now`', () => {
+    describe('with master variant in `before`', () => {
```
tdeekens: This tries to BDD down the tests

tdeekens:
The test pass. Is the `yarn it` failing due to me?

screen shot 2017-09-14 at 12 20 30

packages/sync-actions/src/product-actions.js
```diff
@@ -256,6 +220,30 @@ export function actionsMapPrices (diff, oldObj, newObj) {
...
+  }
+
+  const newMasterVariantId = extractMasterVariantId(newObj)
+  const oldMasterVariantId = extractMasterVariantId(oldObj)
+
+  // Previosuly no master variant existed
+  if (!oldMasterVariantId && newMasterVariantId)
```
wizzy25: I don't understand this part; there will always be a [previous] master variant

packages/sync-actions/src/utils/create-build-actions.js
wizzy25: Just a nitpick: maybe you should rename the variable to `processedBefore`, and also on line [21](https://github.com/commercetools/nodejs/pull/304/files#diff-fb58f097aa83a7bd481dab8d7c44cf2aR21)

packages/sync-actions/src/product-actions.js
tdeekens: Yeah I wasn't 100% about a. the cases here and b. how to guard against unexpected things. Moreover, the tests had cases where no `masterVariant` in `before` existed if I am not mistaken. I guess we could

1. Both have an id and in `now` the id changed?
2. Only `now` has an id (as `else if`-branch)

tdeekens: However, then the loads of tests fail and I assume it's due some "flawed" fixture data maybe.

wizzy25: The oldObj is gotten from the API, so it must contain a masterVariant at least. Although it's good to guard against unexpected situations. I like your implementation as it is.

PS. You should also bump the test coverage 😉

tdeekens: Voila, a completely untested module now tested. I am begging for the % to increase enough. It was the only "red" one in the coverage report ac4b76a

tdeekens:
I verified with the platform that a `masterVariant` will always be auto generated. If you create a project without any variants a `masterVariant` will be implicitly added. As a result all test cases with e.g.

```js
masterVariant: {
  //No id
  name: 'foo',
  prices: {},
}
```

are "flawed" as they won't happen. The `now` and its `masterVariant` you diff will always have an `id`. The `id` is just a sequential value which can't be changed afterwards.

tdeekens:
So from my side this is ready.

packages/sync-actions/src/product-actions.js
```diff
@@ -234,9 +234,6 @@ export function generateChangeMasterVariantAction (oldObj, newObj) {
   const newMasterVariantId = extractMasterVariantId(newObj)
   const oldMasterVariantId = extractMasterVariantId(oldObj)
 
-  // Previosuly no master variant existed
-  if (!oldMasterVariantId && newMasterVariantId)
```
tdeekens: Referring to the comment above and your statement too: this won't happen due to the workings of our API and the implicit generation of a `masterVariant`.

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -291,42 +291,186 @@ describe('Actions', () => {
...
+          variants: [],
+        }
+
+        it('should generate `changeMasterVariant` and `addVariant` action',
+        () => {
+          const actions = productsSync.buildActions(now, before)
+
+          expect(actions).toEqual([
+            { action: 'addVariant', attributes: [{ name: 'foo', value: 'bar' }], id: 2, sku: 'v1' },
+            { action: 'changeMasterVariant', variantId: 2 },
```
yanns: It should also delete the variant with id `1`, shouldn't it?

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -291,42 +291,186 @@ describe('Actions', () => {
...
+            }],
+          }
+
+          it('should generate `changeMasterVariant` and `setAttribute` actions',
+          () => {
+            const actions = productsSync.buildActions(now, before)
+
+            expect(actions).toEqual([
+              { action: 'changeMasterVariant', variantId: 1 },
+              { action: 'setAttribute', name: 'foo-2', value: 'bar-3', variantId: 2 },
```
yanns: 👍

packages/sync-actions/test/product-sync-variants.spec.js
tdeekens: This is one of the semantics I wasn't 100% sure about. It felt safer not to remove than to just throw it away. I'll see what other reviewers say.

yanns: if the goal is that `now` = `before` + all update actions, then we have to remove the variant. Otherwise we have one variant too much in `now`.

tdeekens: Yeap. You're right. I am thinking how we can bake it in.

packages/sync-actions/src/products.js
```diff
@@ -58,10 +62,32 @@ function createProductMapActions (mapActionGroup) {
...
+    variants: [
+      obj.masterVariant,
+      ...obj.variants || [],
+    ],
+  })
+  const hasMasterVariant = obj => obj && obj.masterVariant
+
+  return [
+    hasMasterVariant(before) ? move(before) : before,
+    hasMasterVariant(now) ? move(now) : now,
```
hisabimbola: I am not sure how often we will have this scenario but what if the `masterVariant` object is not passed in the `now` object.

packages/sync-actions/test/product-sync-variants.spec.js
hisabimbola: Personally I think the test description is not explanatory enough

A sample below
```
Actions › without master variant in `now` › with master variant in `before` › should not generate update action to remove master variant
```

compared to something like this

```
Actions › should build `addVariant` action
```

No hard opinion on this though. Usable for me

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -291,42 +291,186 @@ describe('Actions', () => {
...
+        variants: [],
+      }
+
+      const now = {
+        id: '123',
+        // <-- no masterVariant
+        variants: [],
+      }
+
+      it('should not generate update action to remove master variant',
```
hisabimbola: Building from comment above, this test fails

```
describe('no mastervariant in `now` obj', () => {
  const before = {
    id: '123',
    version: 1,
    masterVariant: {
      id: 1,
      sku: 'v1',
      attributes: [{ name: 'foo', value: 'bar' }],
    },
    variants: [],
  }

  const now = {
    id: '123',
    // <-- no masterVariant
    variants: [
      {
        id: 2,
        sku: 'v1',
        attributes: [{ name: 'foo', value: 'bar' }],
      },
    ],
  }

  it('should generate a addVariant action', () => {
    const actions = productsSync.buildActions(now, before)
    expect(actions).toEqual([
      { action: 'addVariant', sku: 'v1', attributes: [{ name: 'foo', value: 'bar' }] },
    ])
  })
})
```

packages/sync-actions/test/product-sync-variants.spec.js
hisabimbola: Yeap, I agree with us removing the update actions.

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -291,42 +291,186 @@ describe('Actions', () => {
...
+        })
+      })
+
+      describe('with existing variant', () => {
+        describe('without changes to attributes', () => {
+          const before = {
+            id: '123',
+            version: 1,
+            masterVariant: {
+              id: 2,
```
hisabimbola: `masterVariant` cannot have a `variantId` of 2. I think it's always 1

Cc @yanns

packages/sync-actions/test/utils/create-map-action-group.spec.js
```diff
@@ -0,0 +1,89 @@
+import createMapActionGroup from '../../src/utils/create-map-action-group'
+
+describe('', () => {
```
hisabimbola: I am not sure why this is empty string.

using `createMapActionGroup` might work since that's the method we are testing.

hisabimbola:
I also think we should add *more* comments in the code to explain a couple of decisions and why.

I find the `sync-actions` a bit hard to read/understand, and mostly because of the custom rules of the `jsondiffpatch` and lack of comments explaining _irregular_ decisions done in the code.

packages/sync-actions/src/products.js
tdeekens: This is what can not happen in theory. Whenever a project has been created implicitly a master variant will be added. Theoretically we could remove these checks but our test data is sometimes a bit messy. E.g. not having a masterVariant in `now` of `before`.

packages/sync-actions/test/product-sync-variants.spec.js
tdeekens: It can have an id of 2 otherwise you'd never be able to move the master variant. The id is immutable and sequentially assigned.

packages/sync-actions/test/utils/create-map-action-group.spec.js
tdeekens: Ups.

tdeekens:
> I find the sync-actions a bit hard to read/understand, and mostly because of the custom rules of the jsondiffpatch

You should have seen my face yesterday. I got a bold spot on my head from scratching 😄

packages/sync-actions/src/products.js
hisabimbola: I mean from the user end, I think what you describing is the `before`(what already exists in the CTP)

If we understand each other though 🤔

packages/sync-actions/test/product-sync-variants.spec.js
hisabimbola: Oh, okay, good to know that `variantId` does not change.

yanns: I confirm when @tdeekens says.
Master variant id is not always 1. The variant id is the invariant :rofl:

packages/sync-actions/src/product-actions.js
```diff
@@ -256,6 +220,27 @@ export function actionsMapPrices (diff, oldObj, newObj) {
...
+  }
+
+  const newMasterVariantId = extractMasterVariantId(newObj)
+  const oldMasterVariantId = extractMasterVariantId(oldObj)
+
+  // Old and new master master variant differ and a new master variant id exists
+  if (newMasterVariantId && oldMasterVariantId !== newMasterVariantId)
+    return createChangeMasterVariantAction(newMasterVariantId)
+
+  return null
```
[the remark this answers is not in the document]
tdeekens: We can I didn't as it will and should always just return one action so the array there felt odd but I'll change it for consistency.

packages/sync-actions/src/products.js
tdeekens: Hm, so I think there are a couple of invariants in this discussions

1. The `before` can be assumed to always have a master variant
2. The `now` should also always have a master variant
   2.1 If it doesn't we can assume that the first variant ought to be the new master variant

If non the the cases hold be should either throw or do nothing which breaks things?

tdeekens:
Regarding the BDD on a comment above I can't comment on

`Actions › without master variant in `now` › with master variant in `before` › should not generate update action to remove master variant`

This gives context as in Given, When Then. Which I think is nicer to follow along as it sets up the context towards what the then is. Otherwise you have to peel the Given and When off the data which needs more knowledge.

packages/sync-actions/test/product-sync-variants.spec.js
tdeekens: Yeah, cause that test lives in the grey zone mentioned above. What do we assume to be intended here? You had a master variant and now you don't want one. Which basically is breaking with our invariants. 

Moreover, the test fails as the `id` 1 doesn't exist anymore. 

1. So should it be removed? (`1`)?
2. Should `2` be implicitly be the new master?
    Did use user intend to do that?
3. Should be keep both?

I think with `now` you declare where you want to add up at. This test is flawed cause the state the user wants to end up in is not valid and we can only assume what he meant.

tdeekens:
@hisabimbola I went over things a little. We now remove the variant if if doesn't exist anymore in variants. I also changed the test structure a bit to hopefully be nicer to read along.

tdeekens:
After talking to @hisabimbola we decided not to have validation on this PR. Changes it might be a breaking change to loads of consumers of the library. Furthermore, our test data does not always contain all required fields. 

This is ready then for a last round of review.

@hisabimbola you want to publish it on the `@latest` or `@next` tag?

packages/sync-actions/src/products.js
```diff
@@ -38,6 +38,11 @@ function createProductMapActions (mapActionGroup) {
     allActions.push(mapActionGroup('variants', () =>
       productActions.actionsMapVariants(diff, oldObj, newObj)))
 
+    const changeMasterVariantActions =
+      productActions.actionsMapMasterVariant(oldObj, newObj)
+    if (changeMasterVariantActions.length > 0)
```
hisabimbola: I think we don't need the condition again, if it's empty array, it's fine for us to push it.

tdeekens: Ah ok. Thought better to be safe than sorry. Here you go f8a6616

PR #310, opened 2017-10-05
fix(sync-action): to only pass reference for reference actions

tdeekens:
#### Summary

Actions based on references only need to send a reference of `typeId` and `id` in addition with the reference itself. Before `sync-actions`, whenever `now` contained the expanded object would also attach those to the action. 

```json
[{
	"action": "transitionState",
	"state": {
		"typeId": "state",
		"id": "fb77d7dc-94c5-45df-95bc-b8499286e4d5",
		"obj": {
			"id": "fb77d7dc-94c5-45df-95bc-b8499286e4d5",
			"version": 2,
			"key": "new",
			"type": "ProductState",
			"roles": [],
			"name": {
				"de": "Neu",
				"en": "New"
			},
			"builtIn": false,
			"transitions": [{
				"typeId": "state",
				"id": "1436898f-6080-4a18-99df-134e05a80f73"
			}],
			"initial": true,
			"createdAt": "2016-08-11T16:07:33.676Z",
			"lastModifiedAt": "2016-08-11T16:07:34.888Z"
		}
	}
}]
```

Whereas the update action should be

```json
[{
	"action": "transitionState",
	"state": {
		"typeId": "state",
		"id": "fb77d7dc-94c5-45df-95bc-b8499286e4d5",
	}
}]
```

This intends to fix that behaviour which will apply to taxCategories and states at least.

### DIFF
--- a/packages/sync-actions/src/utils/common-actions.js
+++ b/packages/sync-actions/src/utils/common-actions.js
@@ -84,7 +84,13 @@ export function buildReferenceActions ({
 
         return {
           action,
-          [key]: newValue,
+          // We only need to pass a reference to the object.
+          // This prevents accidentally sending the expanded (`obj`)
+          // over the wire.
+          [key]: {
+            typeId: newValue.typeId,
+            id: newValue.id,
+          },
         }
       }
 
--- a/packages/sync-actions/test/utils/common-actions.spec.js
+++ b/packages/sync-actions/test/utils/common-actions.spec.js
@@ -81,48 +81,179 @@ describe('Common actions', () => {
       { action: 'transitionState', key: 'state' },
     ]
 
-    it('should build reference actions', () => {
-      const before = {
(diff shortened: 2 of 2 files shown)


tdeekens:
There weren't any. On what level should they, if, be added? `common-actions.spec.js` or on a syncing level?

hisabimbola:
Yeah, at `common-actions.spec.js` level, I think modifying this test [here](https://github.com/commercetools/nodejs/blob/5b417f80b6c16f2790c64b585227a4881e1956a2/packages/sync-actions/test/utils/common-actions.spec.js#L90) might assert this behaviour

tdeekens:
Alright you tickled me. I restructured the test so we declaratively test with and without expansions and then one expectation for each type of action. Hope that makes sense.

tdeekens:
You‘re welcome. Id love to help out on some tooling improvements on the project when I have/get time.

PR #324, opened 2017-10-27
🛤  Add `prettier`, change build tooling to `rollup` and migrate `eslint` 🏄🏻

tdeekens:
#### Summary

This PR contains a bit of what I promised myself to give to this project. It changes quite a few things to be fair and could have been split but just getting such major changes in at once might also reduce the impact on other PRs.

#### Description

Generally the commits describe my journey though aligning the tooling with that of https://github.com/tdeekens/flopflip/pull/49 as discussed with @emmenko.

What this project gains from this

1. __Prettier__
    - Automatic code formatting
2. __ESLint__ update
    - Migrate the base to airbnb to be consistent on that within the company
    - Integrate ESLint with prettier to that rules do not contradict and developers are "forced" to 
format
3. Remove webpack in favor of __rollup__
    - Rollup is generally considered nicer for libraries
    - All CLI tools are still only bundled with babel as we don't need rollup there
    - Other libraries are not bundled up for ESM, CJS and UMD
    - All packages have main entries for module, browser and main for each platform
    - The target folder for all builds is now `dist`
     - All packages have a build script called `build` (API kind of)
4. Update __lerna__
    - This allows us to use the _`--parallel`_ option.
    - Paired with each package having a `build` run script the bit strange webpack/babel interop goes away
5. Remove pre-build for tests
    - Generally this seems not really useful as building happens as postinstall already
6. Simplifies and removes duplication in package.json
7. Adds some redundancy but clear purpose to babelrc and the environments we build for
8. Renames all `test.config.js` to `jestrc.json`
9. Fixes up a whole lot of tests and code due to newer and stricter linting rules (airbnb base)
10. Also all dependencies have been updated hence the closes tags  below

#### What's left

Some 👁  on this and making sure that all packages contain non breaking changes while being built correctly.

- Closes #321
- Closes #319
- Closes #318
- Closes #317
- Closes #316
- Closes #315
- Closes #313
- Closes #309
- Closes #216

screen shot 2017-10-27 at 17 46 18


tdeekens:
Will have to understand the failing build some other day.

emmenko:
> Will have to understand the failing build some other day.

Easy: as far as I can see the main entry point of the package is `./lib/index.js` which does not seem to exist anymore (the babel step got removed). Hence the import simply fails.

Maybe this helps you further 😉

tdeekens:
No hurry as long as master doesn't run away. Thanks, fresh eyes after hours of `package.json` numbing are always a time saver 🐱 . I'll check later. We also have to dig the output of rollup - if there are any rightful "complaints". I wanted all packages to build to `dist` as some before did `dist` others `lib`. Do you have a preference here?

emmenko:
> No hurry as long as master doesn't run away. Thanks, fresh eyes after hours of package.json numbing are always a time saver

Of course, no worries 😄

> I wanted all packages to build to dist as some before did dist others lib. Do you have a preference here?

I usually do it like this:
* `src`: the source files
* `lib`: the transpiled files (e.g. with babel)
* `dist`: UMD files (for browser usage)

tdeekens:
Ok makes sense. I don't mind my inner urge was to homogenise but I'll resort to your pattern. Also reduces "churn" by maybe e.g. tools such as `unpkg` or other CDNs which take the path into account.

emmenko:
Hmm, I'm getting the same error locally.

tdeekens:
I still don't see Travis using the latest, freshly installed `yarn` version which is `1.2.1`. Something else to look into. The version used on now doesn't "even" support workspaces.

emmenko:
> I still don't see Travis using the latest, freshly installed yarn version which is 1.2.1. Something else to look into. The version used on now doesn't "even" support workspaces.

Uh, that sucks 😑

emmenko:
@tdeekens ok, fixed the lint problem https://github.com/commercetools/nodejs/pull/324/commits/a984ec680a857c2858189a5a56a17c969aab91d8 😅

tdeekens:
Thanks. Package main blindness is what I call that.

hisabimbola:
202 files changes 😱 🏃 🏃 

Thanks for the PR @tdeekens

tdeekens:
I maybe touched 20 something files. Fixing up ESLint warnings or package.json things. Anything else is caused by applying prettier and formatting changes. Those are furthermore encapsulated in one commit which can we skipped for review.

tdeekens:
The integration tests seem to fail. I or someone else would have to check if thats caused the this changeset any by what.

wizzy25:
Nice job with this massive PR 💯💯💯
Will look through this more in-depth in the coming days.
But first of all, do we want to start using semicolons?

tdeekens:
Semicolons are one of the few things prettier is configurable for. Just flip a switch, reformat and off you go. Probably a change to do last to not muddy history too much.

tdeekens:
### What's left to do

> Might be missing things, just to have transparency and we can pick things whenever we find time.

- [ ] Ensure that Travis runs with latest `yarn` (1.2.1 not 0.x.x)
   - It's part of the the `before_install` and linked to the path but still doesn't really work
- [ ] Fix integration tests
   - Build setup or some ESLint fixes might have caused breakage 
   - I am not fully aware how/what they do as I never looked into them so far
- [ ] Check what release bumps we need for this
   - Mostly it's a fix but some Promise rejections with an error might be breaking (strictly speaking)

tdeekens:
⛅️ 

yarn

emmenko:
@tdeekens slightly unrelated question but since we're talking about toolings here we can briefly discuss releases. At the moment we are using semantic releases, it's a bit tricky to trigger it correctly though and if you miss something it might not get triggered.
I'm actually also fine with removing this and trigger it using the lerna publish command. What's your experience with this and what would you recommend?

tdeekens:
> trigger it using the lerna publish command

I use lerna publish on flopflip and aliased it to `yarn release` which works super super nice. It also nicely allows for canary releases and next-tags via `lerna publish --npm-tag=next`. It puts the effort on the dev's machine however. I don't know if that's ok for maintainers here. For automation int hat regard, I think you can have something like `--cd-version` to skip the "questions" towards what version should be published. However, I didn't really check out the release process here and if there is a newer/simpler way of doing it.

emmenko:
We wanted to try semantic releases to have less manual steps but I wasn't that much convinced about it from the beginning. I think at this point we can also give a try at lerna publish and then make a decision. I haven't done much releases here lately so the guys can definitely give a better feedback here.

Regarding the pre-packaged dependencies, I actually took the idea from jest (they do a similar thing). It has a couple of benefits (e.g. you make sure that the "released" package actually works as opposed to only "link" the dependencies). But this is also up for discussion/improvement of course 👍

tdeekens:
## Regarding integration tests

Thanks for the clarification. Will then maybe have a look at how Jest these days goes about it. However, to my knowledge the `main` fields will even during "unit" test phrase make the resolving import a `cjs` build after all - not the src files. At least I think. Can be verified with pruning build assets and then running Jest I guess.

## Regarding releasing

I honestly never really worked a lot with semantic release. There kind of always seemed to be a quite high cost for the last "20%" of automation for releasing. At least that's what it felt to me. But I really have to check out the setup and talk to the maintainers about it.

wizzy25:
#### With regards to the failing integration test

> The integration tests seem to fail. I or someone else would have to check if thats caused the this changeset any by what.

One of the projects required for the integration tests is expired. This should be fixed soon from support. So it's not any fault of yours (yet😃)

tdeekens:
Ok, talked to support. They extended the test project. Retriggering the build which should hopefully show a difference.

tdeekens:
🎉  Alright, things are running fine again. To recap a bit what's open

- Check what release bumps we need for this
     - Mostly it's a fix but some Promise rejections with an error might be breaking (strictly speaking)
     - Personally I vote for a fix release as properties on errors are still there as before ref https://github.com/commercetools/nodejs/pull/324/commits/5831975ddcb541a3ec6bc0cc3bdf9448b8730aca#diff-46771d9292e4888bba30a424e8d59071R23
- Should we work on the release process?
     - Personally I'd like to keep this as is as it's already quite a bunch
- Semicolons or no - the question never ends
     - Just let me know what your prefer @hisabimbola and @wizzy25 and I'll reconfigure and reformat if needed

wizzy25:
I think most packages would be fixes, if at all, but the `Discount-code-importer` broke and should be released as such.

For releases, I personally haven't worked with `lerna` and can't say which is better between the both of them. (And this PR is already pretty large)

On the semicolons: say No to semicolons 😄😄

tdeekens:
Ok, I'll change it to no after this has been reviewed as a last chore on the sources. 😈

.babelrc
```diff
@@ -1,25 +1,123 @@
...
+          {
+            "cwd": "./",
+            "alias": {
+              "types": "./types"
+            }
+          }
+        ]
+      ]
+    },
+    "production": {
```
hisabimbola: Do we need this?

Or just some good defaults to have

package.json
```diff
@@ -10,16 +10,15 @@
...
+    "build": "lerna run --parallel build",
     "lint": "flow check && ./scripts/lint packages",
-    "test": "npm run build && jest --config test-config.json",
-    "test:package": "npm run build && ./scripts/test-package",
-    "it": "npm run it:install && npm run it:run",
-    "it:run": "jest --config integration-tests/config.json",
-    "it:install": "npm run build && ./scripts/install-npm-packages-for-integration-tests",
-    "coverage": "npm run build && jest --config test-config.json --no-watchman --coverage",
-    "coverage:ci": "jest --config test-config.json --no-watchman --coverage && codecov",
+    "test": "cross-env NODE_ENV=test jest --config .jestrc.json",
```
hisabimbola: do we need to set the _NODE_ENV_ using `cross-env`, we already set it in the jest config file.

.babelrc
tdeekens: Maybe development and production can be the same. It depends on other repos we reduce the amount of transforms on development to e.g. just the latest browser to produce smaller sourcemaps and snappier code.

In regards to defaulting or merging: it's been a bit of a myth to me and I'd like to be explicit in the babel config as a result.

package.json
```diff
@@ -36,99 +35,63 @@
...
-    "webpack": "3.6.0"
+    "validate-commit-msg": "2.14.0"
   },
   "config": {
     "commitizen": {
       "path": "./node_modules/cz-lerna-changelog"
     }
-  },
-  "eslintConfig": {
-    "extends": "commercetools",
```
hisabimbola: I noticed that this is not being used anymore in merchant center hence the difference coding convention in different teams, but now that we are moving more towards smaller teams and hopefully smaller repos. I think we need to maintain this so we have a consistent coding style, across the company.

AFAIK, we are the only team using this config and if we remove it now then the config itself will be useless. I think we should all agree on a coding style and maintain it across the company.

Cc @emmenko

package.json
tdeekens: Theoretically not. However I would prefer having envs always set in the package.json and the run scripts for explicitness.

tdeekens:
@hisabimbola in regards to what we talked about

1. The readmes are updated to point to the umd build for browsers
2. Prettier is configured to not insert semicolons
3. I merged with master to incorporate @wizzy25 changes regarding the chunk size

tdeekens:
Yoyo. @hisabimbola what's the review status of this? Do you have anything @emmenko?

CONTRIBUTING.md
```diff
@@ -31,12 +31,24 @@ accurate comments, etc.) and any other requirements (such as test coverage).
 
 5. Integration tests are separated out in another folder "/integration-tests". To run the integration test, you need to create a env file as specified [here](https://commercetools.github.io/nodejs/sdk/api/getCredentials.html). Then run integration test with
   ```
-  npm run it -- --projectkey=testing-project --runInBand
+  npm run test:integration -- --projectkey=testing-project --runInBand
   ```
   replace "testing-project" with your project
 
 6. Linting and static checks are done by `npm run lint`. We follow lint rules defined in our [eslint-config](https://github.com/commercetools/eslint-config) which is based on [Airbnb eslint config](https://www.npmjs.com/package/eslint-config-airbnb). Static checks are done using [Flow](https://flowtype.org/) and can be included / adopted incrementally. Commiting also runs a git hook to lint the changed files.
```
wizzy25: Is this line still necessary? Now that the `.eslintrc` file is explicit in this repo?

tdeekens: True. I will remove this in a second.

tdeekens:
Alright here comes some updates

1. The mention of our custom ESLint in the docs is gone /cc @wizzy25 
2. I updated all dependencies again so we don't get old things to master
   - This includes prettier which today released 1.8
3. I fixed up conflicts with master
4. I removed npm-run-all which didn't seem used and with parallel building in rollup feels "pointless"

- Closes #334
- Closes #333
- Closes #331

tdeekens:
Alright, that's enough approvals to get this in. After coordinating with @hisabimbola I will merge this as a fix for all packages.

tdeekens:
Alright the last things improve on a couple of things

1. Be more "silent" on the output during build by using the `global` and `buildins` plugin for rollup
   - As we bundle with `isomorphic-fetch` in some parts we receive warnings whenever we import node APIs
2. Use lodash-es as this better supports ESM and static analysis of rollup
  - All `lodash.flatten` etc packages export CJS builds
  - This easies things for rollup and lets tree shaking do its magic

So should we get this in or wait for #337? /cc @dferber90

tdeekens:
The merge will be postponed until the beginning of next week to have hands on deck instead we encounter issues.

emmenko:
🚀 🙌 🎉

Issue #325, opened 2017-10-30
Implement a complete package for parsing JSON products to CSV


wizzy25:
### Description

Implement a module to parse commercetools products from JSON format to CSV

(Part of a series implementing the rewrite of the [product-csv-sync](https://github.com/sphereio/sphere-node-product-csv-sync) module)

#292 is getting too large so will be broken down into smaller issues and smaller PRs, with this issue encompassing all the issues

### TODO

- [x] Implement core functionality(#292)

- [x] Implement CLI wrapper (#332)

- [x] Add integration tests (#332)

- [x] Add documentation

PR #326, opened 2017-10-30
292 json2csv

### LINKED ISSUE #292: Implement core of JSON to CSV parser for products
### Description

Implement core parsing functionality

Part of #325

wizzy25:
#### Summary

This PR implements the core parsing functionality. Other functionalities (e.g a CLI wrapper) would be added in subsequent PRs. (See [here](https://github.com/commercetools/nodejs/issues/325#issue-269572271) for checklist)

resolves #292


packages/json-parser-products/test/mappings.spec.js
```diff
@@ -0,0 +1,620 @@
+import { oneLineTrim } from 'common-tags'
+import ProductMapping from '../src/mappings'
+
+describe('::ProductMapping', () => {
+  let productMapping
+  beforeEach(() => {
+    productMapping = new ProductMapping()
+  })
+  it('should be defined', () => {
+    expect(ProductMapping).toBeDefined()
```
junajan: Is this test needed? If ProductMapping is not defined, the `beforeEach` will fail first

packages/json-parser-products/test/mappings.spec.js
```diff
@@ -0,0 +1,620 @@
...
+        key: 'cat-keysss',
+      }]
+      const args = ['namedPath', ';', 'en']
+      const expected = oneLineTrim`
+        greatgrandparent>grandparent>parent>cat-in-en
+        ;another-parent>cat-foo-en`
+      expect(ProductMapping._mapCategories(sampleCat, ...args)).toBe(expected)
+    })
+
+    describe('::retrieveNamedPath', () => {
```
junajan: Isn't this test similar to the previous one? Can it be merged somehow?

junajan:
Also is the `json-parser-products` a good name for this tool? Because it converts products JSON to CSV so I would expect something like `product-json-to-csv` instead.

packages/json-parser-products/src/writer.js
```diff
@@ -0,0 +1,100 @@
...
+export function writeToZipFile (productStream, output, logger) {
+  const tmpDir = tmp.dirSync({ unsafeCleanup: true }).name
+  let currentProductType
+  let fileName
+  let filePath
+  let fileStream
+  let headers
+  const headersCache = {}
+  const streamCache = {}
+  const myset = new Set()
```
hisabimbola: what is `mySet` 🤔

packages/json-parser-products/src/writer.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Accept a highland stream and write the output to multiple files per
+// product type, then compress all files to a zip file
+export function writeToZipFile (productStream, output, logger) {
+  const tmpDir = tmp.dirSync({ unsafeCleanup: true }).name
+  let currentProductType
+  let fileName
+  let filePath
+  let fileStream
```
hisabimbola: No need to make `fileStream`, `filePath`, `fileName` global or defined with `let`

packages/json-parser-products/src/main.js
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  parse (input: stream$Readable, output: stream$Writable) {
+    this.logger.info('Starting conversion')
+    let productCount = 0
+    input.setEncoding('utf8') // TODO: Move to CLI
+
+    return highland(input)
+      // parse chunk and split into JSON object strings
+      .splitBy('\n')
```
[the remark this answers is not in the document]
wizzy25: @junajan @hisabimbola The corresponding exporter separates each product by a new line character, that's why the parser also uses this. As this is still a work in progress, the separator can be passed through the CLI, with this newLine character being the default

packages/json-parser-products/src/mappings.js
```diff
@@ -0,0 +1,172 @@
...
+    multiValueDelimiter = ';',
+  }: Object = {}) {
+    this.fillAllRows = fillAllRows
+    this.categoryBy = categoryBy
+    this.lang = lang
+    this.multiValDel = multiValueDelimiter
+  }
+
+  run (product: ResolvedProdProj) {
+    const mergedVarProducts = ProductMapping._mergeVariants(product)
```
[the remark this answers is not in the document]
wizzy25: The variable namings tend to let me down sometimes 😏

packages/json-parser-products/src/mappings.js
```diff
@@ -0,0 +1,172 @@
...
+        case 'productType': {
+          acc[property] = value.name
+          break
+        }
+        case 'state': {
+          acc[property] = value.key
+          break
+        }
+        case 'taxCategory': {
+          acc[property] = value.key || value.name
```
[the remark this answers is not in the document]
wizzy25: key is unique but not required, so not all categories will have key. If a category doesn't then the name will be used.
On a second thought, the current importer resolves only `name` for `taxCategory` so I'll remove the `key` possibility

packages/json-parser-products/src/mappings.js
```diff
@@ -0,0 +1,172 @@
...
+        case 'categories': {
+          acc[property] = ProductMapping._mapCategories(
+            value, this.categoryBy, this.multiValDel, this.lang)
+          break
+        }
+        case 'categoryOrderHints': {
+          if (!isEmpty(value))
+            acc[property] = Object.keys(value).map((key: string): string => (
+              `${key}:${value[key]}`
+            )).join(this.multiValDel)
```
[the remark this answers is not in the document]
wizzy25: Why is that?

packages/json-parser-products/src/mappings.js
```diff
@@ -0,0 +1,172 @@
...
+        case 'version': {
+          // Remove version number
+          break
+        }
+        case 'variant': {
+          if (!isEmpty(value.attributes)) {
+            acc.attr = {}
+
+            value.attributes.forEach((attribute: Object) => {
+              acc.attr[attribute.name] = attribute.value.key || attribute.value
```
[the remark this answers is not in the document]
wizzy25: All attributes have a `value` property regardless of the type. If the `value` property is an object, then it always has a unique `key`

packages/json-parser-products/src/writer.js
```diff
@@ -0,0 +1,100 @@
+import archiver from 'archiver'
+import EmitOnce from 'single-emit'
+import fs from 'fs'
+import json2csv from 'json2csv'
+import path from 'path'
+import tmp from 'tmp'
+import slugify from './utils'
```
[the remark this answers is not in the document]
wizzy25: Will be removed with the move to [slugify](https://www.npmjs.com/package/slugify)

packages/json-parser-products/src/writer.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+        currentProductType = product.productType
+        fileName = `${slugify(product.productType)}.csv`
+        filePath = path.join(tmpDir, fileName)
+        // use a stream cache for switching back and forth between file
+        // streams and reusing the same write stream
+        if (streamCache[currentProductType])
+          fileStream = streamCache[currentProductType]
+        else {
+          fileStream = fs.createWriteStream(filePath, { flags: 'a' })
```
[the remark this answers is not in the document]
wizzy25: Yes it is, as the file writes are performed per product type, and we cannot guarantee that the products will be sorted by product type. Without this, each new write will overwrite the old one

packages/json-parser-products/test/main.spec.js
```diff
@@ -0,0 +1,558 @@
...
+    const logger = {
+      error: () => {},
+      warn: () => {},
+      info: () => {},
+      verbose: () => {},
+    }
+    const parserConfig = {
+      staged: true,
+      batch: 5,
+      predicate: 'foo=bar',
```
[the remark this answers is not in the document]
wizzy25: Oops. wrong config

packages/json-parser-products/test/main.spec.js
```diff
@@ -0,0 +1,558 @@
...
+    it('should write data to single `csv` file if headers', () => {
+      jsonParserProduct.parse = jest.fn(() => 'foo')
+      jsonParserProduct.parserConfig.headers = []
+
+      jsonParserProduct.run()
+      expect(csvSpy).toBeCalled()
+      expect(zipSpy).not.toBeCalled()
+    })
+
+    it('should write data to `zip` file if no headers', () => {
```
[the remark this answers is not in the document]
wizzy25: Trying to stay within accepted line length... but something better will be done

packages/json-parser-products/test/mappings.spec.js
wizzy25: TDD... Will be removed

packages/json-parser-products/test/mappings.spec.js
wizzy25: It can, but they're testing different methods

packages/json-parser-products/src/main.js
```diff
@@ -0,0 +1,307 @@
...
+    this.categoriesCache = {}
+    this.accessToken = accessToken
+
+    const mappingParams = {
+      fillAllRows: this.parserConfig.fillAllRows,
+      categoryBy: this.parserConfig.categoryBy,
+      lang: this.parserConfig.lang,
+      multiValueDelimiter: this.parserConfig.multiValueDelimiter,
+    }
+    this._productMapping = new ProductMapping(mappingParams)
```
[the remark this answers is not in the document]
wizzy25: I think it's better to stay explicit for readability

packages/json-parser-products/src/main.js
```diff
@@ -0,0 +1,307 @@
...
+    if (headers)
+      writeToSingleCsvFile(productStream, output, this.logger, headers)
+    else
+      writeToZipFile(productStream, output, this.logger)
+  }
+
+  parse (input: stream$Readable, output: stream$Writable) {
+    this.logger.info('Starting conversion')
+    let productCount = 0
+    input.setEncoding('utf8') // TODO: Move to CLI
```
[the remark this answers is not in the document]
wizzy25: Will be implemented in the CLI PR... But yes, it's about time😄

packages/json-parser-products/src/main.js
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  _resolveProductType (productTypeReference: TypeReference): Object {
+    if (!productTypeReference)
+      return {}
+
+    const productTypeService = this._createService('productTypes')
+    const uri = productTypeService.byId(productTypeReference.id).build()
+    return this.fetchReferences(uri)
+      .then(({ body }: SuccessResult): { productType: ProductType } => (
```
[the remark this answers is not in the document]
wizzy25: As it's defined [here](https://github.com/commercetools/nodejs/blob/master/types/sdk.js#L35) I thought it wise to go with it. Each resolution will have different types and this was the most suited to cover them all. And since we have just one method handling all fetched data, we would run into flow issues if we are more specific

packages/json-parser-products/src/mappings.js
```diff
@@ -0,0 +1,172 @@
...
+          break
+        }
+        case 'categoryOrderHints': {
+          if (!isEmpty(value))
+            acc[property] = Object.keys(value).map((key: string): string => (
+              `${key}:${value[key]}`
+            )).join(this.multiValDel)
+          break
+        }
+        case 'searchKeywords': {
```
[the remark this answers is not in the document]
wizzy25: If we do not add it, then we'd get the `[object Object]` in the csv file.

packages/json-parser-products/src/writer.js
wizzy25: Implementing caching strategies 😄

packages/json-parser-products/src/mappings.js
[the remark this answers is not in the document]
junajan: Should we somehow prevent the user of setting `:` as a multivalue delimiter?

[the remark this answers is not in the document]
wizzy25: Wait. I think I read your first comment wrong (didn't see the **not**). Yes `this.multiValDel` should not be `:`. It can be advised in the CLI to not use `:`

packages/json-parser-products/src/mappings.js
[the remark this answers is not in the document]
junajan: @wizzy25 I checked few projects and I found, for example, this `number` attribute:
```json
{
    "name": "attr-number",
   "value": 8300138
}
```
Which would cause an error when retrieving `attribute.value.key` (same as boolean or a simple text would do)
Then there is, for example, this `ltext` object without the key:
```json
{
  "name": "attr-ltext",
  "value": {
    "de": "PCE"
  }
}
```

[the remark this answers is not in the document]
wizzy25: @junajan If a field has no `key` value, like the examples you sighted, undefined will be returned, and the attribute will be set to `value` instead of `value.key` (It will not throw)
`attribute.value.key || attribute.value`
In the latter case, on flattening we will simply get a localised field in our csv after flattening, which is what we want.
You can see one of the tests for it [here](https://github.com/commercetools/nodejs/pull/326/files/4ea77c1801075e182193f1f9cf0171eea052a1a2#diff-f675f797a8217a80c1646e0340b26671R168) and [here](https://github.com/commercetools/nodejs/pull/326/files/4ea77c1801075e182193f1f9cf0171eea052a1a2#diff-f675f797a8217a80c1646e0340b26671R210)

packages/json-parser-products/src/writer.js
wizzy25: `fileStream` is needed with the let

wizzy25:
Kindly review @PhilippSpo @tdeekens 
Thanks

packages/json-parser-products/src/index.js
```diff
@@ -0,0 +1 @@
+export default from './main'
```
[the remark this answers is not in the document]
PhilippSpo: Not sure I can follow. I would have the same argument as @junajan .

tdeekens:
I can review code but I have zero context from a functionality perspective on this one. Can you elaborate a bit more what the feature does from a product perspective and how it fits into things?

packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -0,0 +1,195 @@
...
+    const productWithVariants: Array<Object> = product.variant.map(
+      (eachVariant: Variant): SingleVarPerProduct => ({
+        variant: eachVariant,
+      })
+    )
+    productWithVariants[0] = { ...product, ...productWithVariants[0] }
+    return productWithVariants
+  }
+
+  _mapProperties(product: SingleVarPerProduct): MappedProduct {
```
tdeekens: We do so mich data wrangling here and I cant really follow why. Can we add a comment for what the input abd output shape of this is and why we need to do this?

packages/product-json-to-csv/src/writer.js
```diff
@@ -0,0 +1,101 @@
...
+  const columnNames = mapHeaders(trimmedHeaders)
+  productStream
+    .each(product => {
+      const csvData = json2csv({
+        data: product,
+        fields: columnNames,
+        hasCSVColumnTitle: false,
+        quotes: '',
+      })
+      output.write(`${csvData}\n`)
```
tdeekens: How does it know when its done without returning anything?

packages/product-json-to-csv/src/writer.js
```diff
@@ -0,0 +1,101 @@
...
+        // get product columnNames from cache or masterVariant.
+        // variants may not have all header fields so should not be relied on
+        if (columnNamesCache[product.productType])
+          columnNames = columnNamesCache[product.productType]
+        else {
+          hasCSVColumnTitle = true
+          columnNames = mapHeaders(Object.keys(product))
+          columnNamesCache[product.productType] = columnNames
+        }
+        currentProductType = product.productType
```
tdeekens: Is this like grouping them by product type? What if they dont come in sorted?

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+import ProductJsonToCsv from '../src/main'
+import * as writer from '../src/writer'
+
+const streamTest = StreamTest.v2
+
+describe('ProductJsonToCsv', () => {
+  let productJsonToCsv
+  beforeEach(() => {
+    const logger = {
+      error: () => {},
```
tdeekens: Have you tried out Jests auto mocking? You import what you want to mock. 

import logger from utils/logger

to then mock

jest.mock(utils/logger)

All functions on the logger are now jest.fn().

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+      expect(productJsonToCsv.client).toBeDefined()
+      expect(productJsonToCsv.parserConfig).toEqual(defaultConfig)
+    })
+  })
+
+  describe('::run', () => {
+    let zipSpy
+    let csvSpy
+    beforeEach(() => {
+      zipSpy = jest.spyOn(writer, 'writeToZipFile').mockImplementation(() => {})
```
tdeekens: Automocking applies here to and eases things. This should have an example. https://github.com/tdeekens/flopflip/blob/master/packages/splitio-adapter/modules/adapter/adapter.spec.js

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+      csvSpy = jest
+        .spyOn(writer, 'writeToSingleCsvFile')
+        .mockImplementation(() => {})
+    })
+
+    it('should write data to single `csv` file if headers are set', () => {
+      productJsonToCsv.parse = jest.fn(() => 'foo')
+      productJsonToCsv.parserConfig.headers = []
```
tdeekens: Makes me shiffer a bit cause its basicslly s private property.

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+      const productStream = productJsonToCsv.parse(inputStream)
+      const expected = [
+        {
+          id: 'product-1-id',
+          'slug.en': 'my-slug-1',
+          'variant.id': 'mv-id',
+          'variant.key': 'mv-key',
+        },
+      ]
+      productStream.toArray(actual => {
```
tdeekens: If these would be returning promises we could just run these in a beforeEach and return them. 

In its we can also then use toResolve etc.

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+
+    it('should log and emit error if error occurs', done => {
+      const fakeError = new Error('fake error')
+      productJsonToCsv._resolveReferences = jest.fn(() =>
+        Promise.reject(fakeError)
+      )
+      const outputStream = {
+        emit: jest.fn(),
+      }
+      const productStream = productJsonToCsv.parse(inputStream, outputStream)
```
tdeekens: Maybe just flavour but we put setup into its. Where BDD advocates that to be more in describes and a beforeEach  

describe: when erros
   beforeEach
   it should log

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+        const categoryOrderHints = {
+          'res-cat-name-1': '0.015',
+          'res-cat-name-2': '0.987',
+        }
+
+        productJsonToCsv._resolveProductType = jest.fn(() => ({ productType }))
+        productJsonToCsv._resolveTaxCategory = jest.fn(() => ({ taxCategory }))
+        productJsonToCsv._resolveState = jest.fn(() => ({ state }))
+        productJsonToCsv._resolveCategories = jest.fn(() => ({ categories }))
+        productJsonToCsv._resolveCategoryOrderHints = jest.fn(() => ({
```
tdeekens: Oh this reaching in feels fragile. To me its a bit of an indication of the module/class not following the single responsibilty pattern. If we extract this resolving things into modules themselves we can mock them without doing the reaching in.

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+          categories: [
+            { id: 'cat-id-1', name: { en: 'res-cat-name-1' } },
+            { id: 'cat-id-2', name: { en: 'res-cat-name-2' } },
+          ],
+          categoryOrderHints: {
+            'res-cat-name-1': '0.015',
+            'res-cat-name-2': '0.987',
+          },
+        }
+        const actual = await productJsonToCsv._resolveReferences(sampleProduct)
```
tdeekens: We can up our game here with dealing with promises. An example would be

await expect(user.getUserName(5)).resolves.toEqual(bar)

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -0,0 +1,625 @@
...
+        expect(actual).toEqual(expected)
+      })
+    })
+
+    describe('::_resolveProductType', () => {
```
tdeekens: Generally I would consider this an internal which should be tested through the public API. These fine grained unit tests often turn out as very fragile.

packages/product-json-to-csv/test/map-headers.spec.js
```diff
@@ -0,0 +1,181 @@
...
+        label: 'attribute',
+        value: 'attribute',
+      },
+      {
+        label: 'customAttribute',
+        value: 'customAttribute',
+      },
+    ]
+
+    expect(mapHeaders(headers)).toEqual(expected)
```
tdeekens: These should likely really be split up. Such masive asserts are hard to debug, maintain and extend. Tricks up our sleeve are expect.arrayContaining and that in multiple its.

packages/product-json-to-csv/test/map-product-data.spec.js
```diff
@@ -0,0 +1,622 @@
...
+  describe('::mergeVariants', () => {
+    it('merge all variants in a product into one array', () => {
+      const sampleProduct = {
+        masterVariant: { id: 1 },
+        variants: [{ id: 2 }, { id: 3 }],
+      }
+      const expected = {
+        variant: [{ id: 1 }, { id: 2 }, { id: 3 }],
+      }
+      const actual = ProductMapping._mergeVariants(sampleProduct)
```
tdeekens: Looking at the spec I think the implementation is somewhat overly complex.

packages/product-json-to-csv/test/map-product-data.spec.js
```diff
@@ -0,0 +1,622 @@
...
+        {
+          name: {
+            en: 'cat-foo-en',
+            de: 'cat-foo-de',
+          },
+          externalId: 'cat-foo-id',
+        },
+      ]
+      const expected = 'cat-ext-id;cat-foo-id'
+      const args = ['externalId', ';']
```
tdeekens: These are not really args. They are something additional we spread into the function applied. Should we rename these accordingly?

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+} from 'types/product'
+import type { Client, ClientRequest, SuccessResult } from 'types/sdk'
+
+import { createClient } from '@commercetools/sdk-client'
+import { createRequestBuilder } from '@commercetools/api-request-builder'
+import { createHttpMiddleware } from '@commercetools/sdk-middleware-http'
+import { createAuthMiddlewareForClientCredentialsFlow } from '@commercetools/sdk-middleware-auth'
+import { createUserAgentMiddleware } from '@commercetools/sdk-middleware-user-agent'
+import highland from 'highland'
+import Promise from 'bluebird'
```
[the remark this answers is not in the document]
wizzy25: Bluebird promises are [significantly faster than the native implementation](https://kyrylkov.com/2017/04/25/native-promises-async-functions-nodejs-8-performance/) while also exposing nicer APIs (No `Promise.map` on native promise implementation)

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+      categoryOrderHintBy: 'name', // key, externalId or name supported
+      delimiter: ',',
+      fillAllRows: false,
+      language: 'en',
+      multiValueDelimiter: ';',
+      productSeparator: '\n',
+    }
+
+    this.parserConfig = { ...defaultConfig, ...parserConfig }
+    this.logger = {
```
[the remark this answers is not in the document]
wizzy25: The logger is passed in from [L74](https://github.com/commercetools/nodejs/pull/326/files/3ecd2c2f940eb11a575140e0e2c267aa46757765#diff-e0a3628cb03c02182cfe6b4c1295da09R74). It is done this way so that in the event that a custom logger is passed in without any of the functions, it will not throw while trying to log

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+    ]).then(
+      (
+        [
+          productType,
+          taxCategory,
+          state,
+          categories,
+          categoryOrderHints,
+        ]: Array<Object>
+      ): ResolvedProdProj => ({
```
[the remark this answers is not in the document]
wizzy25: Can do that now (`rollup` should still make it readable)

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+
+    const catIdentifier = this.parserConfig.categoryOrderHintBy
+    const lang = this.parserConfig.language
+    const categoryIds = Object.keys(categoryOrderHintsReference)
+    return this._getCategories(categoryIds).then(
+      (resolvedCategories: Array<Category>): Object => {
+        const categoryOrderHints = {}
+        resolvedCategories.forEach((category: Category) => {
+          const catRef =
+            catIdentifier === 'name'
```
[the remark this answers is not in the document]
wizzy25: One of three options can be passed in `name`, `key` or `externalId`
This check is to know what to resolve

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+    })
+    if (!notCachedIds.length) return Promise.resolve(cachedCategories)
+
+    const predicate = `id in ("${notCachedIds.join('", "')}")`
+    const categoriesService = this._createService('categories')
+    const uri = categoriesService.where(predicate).build()
+    return this.fetchReferences(uri).then(
+      ({ body: { results } }: SuccessResult): Array<Category> => {
+        results.forEach((result: Category) => {
+          cachedCategories.push(result)
```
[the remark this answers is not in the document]
wizzy25: I don't get this... Could you expatiate?

packages/product-json-to-csv/src/map-headers.js
```diff
@@ -0,0 +1,50 @@
+/* @flow */
+
+type HeaderObj = {
+  label: string,
+  value: string,
+}
+
+export default function mapHeaders(headers: Array<string>): Array<HeaderObj> {
```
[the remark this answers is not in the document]
wizzy25: More like mapping the csv headers to the object keys. This part requires more context so we'll talk about this

packages/product-json-to-csv/src/map-headers.js
```diff
@@ -0,0 +1,50 @@
...
+  return headers.map((header: string): HeaderObj => {
+    if (header === '_published' || header === 'publish')
+      return {
```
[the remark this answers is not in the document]
wizzy25: ^^^

packages/product-json-to-csv/src/map-product-data.js
wizzy25: This part is context dependent, and will be explained in a wiki or face to face

packages/product-json-to-csv/src/writer.js
wizzy25: One of the beauties of `highland`
`output.end()` is called in the `.done()` method of the highland stream which closes it

packages/product-json-to-csv/test/main.spec.js
wizzy25: Haven't tried it, but I'd love to

packages/product-json-to-csv/src/writer.js
wizzy25: It almost definitely wouldn't come sorted. [L62 to L66](https://github.com/commercetools/nodejs/pull/326/files/3ecd2c2f940eb11a575140e0e2c267aa46757765#diff-7cc529106f4be69ec3e79b24c55b1fa7R62) uses a cache to decide to which file to write to

packages/product-json-to-csv/test/map-headers.spec.js
wizzy25: I agree with this. Why I didn't split it up is that in this case the order of the array is also important

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+  _resolveReferences: Function
+  _productMapping: Object
+
+  constructor(
+    apiConfig: ApiConfigOptions,
+    parserConfig: ParserConfigOptions,
+    logger: LoggerOptions,
+    accessToken: string
+  ) {
+    this.apiConfig = apiConfig
```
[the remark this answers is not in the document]
wizzy25: How will this work in this case?

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+    const mappingParams = {
+      fillAllRows: this.parserConfig.fillAllRows,
+      categoryBy: this.parserConfig.categoryBy,
+      lang: this.parserConfig.lang,
+      multiValueDelimiter: this.parserConfig.multiValueDelimiter,
+    }
+    this._productMapping = new ProductMapping(mappingParams)
+  }
+
+  run(input: stream$Readable, output: stream$Writable) {
```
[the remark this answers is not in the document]
wizzy25: Some of the cli tools use `run` and none currently uses `execute`. May cause misunderstanding of the modules, but I have a loose opinion on this

packages/product-json-to-csv/src/main.js
```diff
@@ -0,0 +1,292 @@
...
+        ...taxCategory,
+        ...state,
+        ...categories,
+        ...categoryOrderHints,
+      })
+    )
+  }
+
+  _resolveProductType(productTypeReference: TypeReference): Object {
+    if (!productTypeReference) return {}
```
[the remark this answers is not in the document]
wizzy25: From [L135](https://github.com/commercetools/nodejs/pull/326/files/3ecd2c2f940eb11a575140e0e2c267aa46757765#diff-e0a3628cb03c02182cfe6b4c1295da09R135), we're already in a promise so the empty object gets wrapped in the promise in any case

packages/product-json-to-csv/src/main.js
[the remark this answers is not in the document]
wizzy25: A key-value cache is defined [here](https://github.com/commercetools/nodejs/pull/326/files/3ecd2c2f940eb11a575140e0e2c267aa46757765#diff-e0a3628cb03c02182cfe6b4c1295da09R76). This array is used because a product normally has an array of categories which should be resolved simultaneously. This array is passed to the function to resolve the categories

packages/product-json-to-csv/src/map-headers.js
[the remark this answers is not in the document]
wizzy25: `.endsWith()` will not work here as some fields have different endings from the passed in headers and the property name (`tax` = `taxCategory`)
`.includes()` works for most fields, but with fields having "variant" (`variantId`, `variantKey`) this wouldn't also work. For simplicity and concurrence (and speed), it's to match the exact expected header fields

packages/product-json-to-csv/test/main.spec.js
wizzy25: While it makes sense to test through the public APIs, I also believe each method should be entitled to it's own tests... sort of

packages/product-json-to-csv/test/main.spec.js
```diff
@@ -3,16 +3,18 @@ import { oneLineTrim } from 'common-tags'
...
   beforeEach(() => {
     const logger = {
-      error: () => {},
-      warn: () => {},
-      info: () => {},
-      verbose: () => {},
+      error: jest.fn(),
+      warn: jest.fn(),
+      info: jest.fn(),
+      verbose: jest.fn(),
```
tdeekens: Can't you just

```js
import logger from '../whereever/logger/is'

jest.mock('logger')
```

which is essentially the same?

packages/product-json-to-csv/test/map-headers.spec.js
```diff
@@ -1,181 +1,275 @@
...
+      expect(mapHeaders(headerFields)[4]).toEqual({
         label: 'variantKey',
         value: 'variant.key',
-      },
-      {
+      })
+    })
+
+    it('contains `id`', () => {
+      expect(mapHeaders(headerFields)[5]).toEqual({
```
tdeekens: In case you want this to me more resiliant why not go for a 

```js
expect(mapHeaders(headerFields)).toEqual(expect.arrayContaining([{ label: 'id', value: 'id' }])
```

or does the position matter so much? I guess not...

wizzy25: The position matters. We had issues before that users complained the order of the fields in the templates didn't match the order in the exported file and this isn't the desired functionality

wizzy25:
Review will contInue on the next PR

PR #330, opened 2017-11-01
fix(sync-actions): update image update action

### LINKED ISSUE #322: [sync-actions] update "changeImageLabel" update action
### Description

The CTP changed the "changeImageLabel" update action to "setImageLabel". We should update it also in the `sync-actions`

wizzy25:

change the `changeImageLabel` update action to `setImageLabel`

resolves #322


wizzy25:
@hisabimbola There's currently no integration test for specific update actions. Doing that would be an overkill IMO. Or what do you think?

PR #337, opened 2017-11-07
feat(api-request-builder): add parse option

dferber90:

More mature alternative to #335 

#### Summary
- adds chained `parse` command  to `service` returned from `api-request-builder`
- `parse` will throw errors on invalid operations as it uses the existing setters behind the scenes
- `parse` will throw an error when an unknown key exists in the provided `params` object (not sure we want to keep this)
- the `parse` works just like every other setter method and can be chained as such
  - `service.parse({ page: 5 }).perPage(10).build()`
  - `service.perPage(10).parse({ page: 5 }).build()`

#### Description
At the moment there is only an imperative way to use `api-request-builder`:

```js
createService(options, projectKey)
  .byId('123')
  .expand('channel')
  .build()
```

On the side of people using the RequestBuilder this makes testing the usage of the package hard, as we either:
- have to mock the complete API or 
- break test-isolation by asserting the returned string from `service.build()` (we really only care about what we give to it, not what it returns; the conversion of the arguments to a string should be tested in `api-request-builder` itself - not in every test using the request builder)

An easier, more declarative API is to allow passing an object which describes what we want to fetch:

```js
const params = {
  id: 'some-id',
  version: 3,
  expand: ['foo=true']
}

createService(options, projectKey)
  .parse(params)
  .build()
```

Now the application-level tests only need to ensure that `params` has the correct format and values. Tests stay isolated (we don't assert the output string) and there is no need for complicated mocking.

#### Todo

- Tests
    - [ ] Unit
    - [x] Integration
    - [ ] Acceptance
- [x] Documentation


packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
+    )
+
+  // query
+  if (params.where)
+    params.where.forEach((predicate: string) => { this.where(predicate) })
+  if (hasKey(params, 'whereOperator')) this.whereOperator(params.whereOperator)
+
+  // version
+  if (hasKey(params, 'version')) this.withVersion(params.version)
+}
```
dferber90: <details>
<summary>I also considered doing this approach (expand me)</summary>

```js
/**
 * Set the supplied parameters given the current service object.
 *
 * @return {void}
 */
export function setParams (params: ServiceBuilderDefaultParams) {
  Object.entries(params).forEach(([paramKey, paramValue]: [string, mixed]) => {
    switch (paramKey) {
      // query-expand
      case 'expand':
        if (!Array.isArray(paramValue)) throw new Error('Expected array')
        paramValue.forEach((expansion: string) => { this.expand(expansion) })
        break

      // query-id
      case 'id':
        this.byId(paramValue)
        break
      case 'key':
        this.byKey(paramValue)
        break
      case 'customerId':
        this.byCustomerId(paramValue)
        break

      // query-page
      case 'sort':
        paramValue.forEach(
          (sortDesc: { by: string, direction: 'asc' | 'desc' }) => {
            this.sort(sortDesc.by, sortDesc.direction === 'asc')
          },
        )
        break
      case 'page':
        this.page(paramValue)
        break
      case 'perPage':
        this.perPage(paramValue)
        break

      // query-projection
      case 'staged':
        this.staged(paramValue)
        break
      case 'priceCurrency':
        this.priceCurrency(paramValue)
        break
      case 'priceCountry':
        this.priceCountry(paramValue)
        break
      case 'priceCustomerGroup':
        this.priceCustomerGroup(paramValue)
        break
      case 'priceChannel':
        this.priceChannel(paramValue)
        break

      // query-search
      case 'text':
        this.text(paramValue.value, paramValue.language)
        break
      case 'fuzzy':
        if (paramValue) this.fuzzy()
        break
      case 'fuzzyLevel':
        this.fuzzyLevel(paramValue)
        break
      case 'markMatchingVariants':
        if (paramValue) this.markMatchingVariants()
        break
      case 'facet':
        paramValue.forEach((facet: string) => { this.facet(facet) })
        break
      case 'filter':
        paramValue.forEach((filter: string) => { this.filter(filter) })
        break
      case 'filterByQuery':
        paramValue.forEach((query: string) => { this.filterByQuery(query) })
        break
      case 'filterByFacets':
        paramValue.forEach((facet: string) => { this.filterByFacets(facet) })
        break

      // query-suggest
      case 'searchKeywords':
        paramValue.forEach(
          (searchKeyword: {value: string, language: string }) => {
            this.searchKeywords(searchKeyword.value, searchKeyword.language)
          },
        )
        break

      // query
      case 'where':
        paramValue.forEach((predicate: string) => { this.where(predicate) })
        break
      case 'whereOperator':
        this.whereOperator(paramValue)
        break

      // version
      case 'version':
        this.withVersion(paramValue)
        break

      default: throw new Error(`Unknown key "${paramKey}"`)
    }
  })
}
```

</details>

--- 


That `switch` approach
- makes the `knownKeys` array unnecessary
- makes it hard in case we want to have nested keys (we'd need nested `switch`-statements)

Not too sure which is better. Both work.

packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
     this.params.pagination = getDefaultSearchParams().pagination
     this.params.search = getDefaultSearchParams().search
   }
 
-  if (this.features.indexOf(features.projection) >= 0)
+  if (this.features.includes(features.projection))
     this.params.staged = true
 
-  if (this.features.indexOf(features.suggest) >= 0)
+  if (this.features.includes(features.suggest))
```
dferber90: Those `includes` are unrelated changes.

docs/sdk/api/apiRequestBuilder.md
```diff
@@ -93,3 +93,78 @@ client.execute(channelsRequest)
...
+  searchKeywords?: Array<{language: string; value: string;}>;
+
+  // query
+  where?: Array<string>;
+  whereOperator?: 'and' | 'or';
+
+  // version
+  version?: string;
+}
+```
```
dferber90: I fear that this might get outdated fast. Should we rather link to the type definition?

packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
+  // verify params
+  const knownKeys = [
+    'expand', 'id', 'key', 'customerId', 'sort', 'page', 'perPage', 'staged',
+    'priceCurrency', 'priceCountry', 'priceCustomerGroup', 'priceChannel',
+    'text', 'fuzzy', 'fuzzyLevel', 'markMatchingVariants', 'facet', 'filter',
+    'filterByQuery', 'filterByFacets', 'searchKeywords', 'where', 'version',
+  ]
+  Object.keys(params).forEach((key: string) => {
+    if (!knownKeys.includes(key)) throw new Error(`Unknown key "${key}"`)
+  })
```
dferber90: Not sure whether we want to warn on unknown keys or not. But it's in for now as it can catch possible errors early. Otherwise we'd fail silently.

Please advise on which is better here. As far as I could tell we're failing loudly in other cases, e.g. when page is provided with `0` (it must be at least `1`).

packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
+  if (params.expand)
+    params.expand.forEach((expansion: string) => { this.expand(expansion) })
+
+  // query-id
+  if (hasKey(params, 'id')) this.byId(params.id)
+  if (hasKey(params, 'key')) this.byKey(params.key)
+  if (hasKey(params, 'customerId')) this.byCustomerId(params.customerId)
+
+  // query-page
+  if (params.sort) params.sort.forEach(
```
dferber90: The internal representation nests these objects as `pagination: { page, perPage, sort }`. The public API has no nesting. Not sure which is better.

packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
+    )
+  if (params.filterByFacets)
+    params.filterByFacets.forEach(
+      (facet: string) => { this.filterByFacets(facet) },
+    )
+
+  // query-suggest
+  if (params.searchKeywords)
+    params.searchKeywords.forEach(
+      (searchKeyword: {value: string, language: string }) => {
```
dferber90: So far this library has been using `lang` when referring to variables internally.

I chose `language` deliberately, as
- it's coming from the application 
- it would be  weird to have `language` shortened to `lang`, but `value` not shortened to `val`

packages/api-request-builder/test/create-service.spec.js
```diff
@@ -88,61 +89,283 @@ describe('createService', () => {
...
+
+    describe('features of parsed object', () => {
+      // query-expand
+      it('should support single `expand`', () => {
+        expect(service.parse({ expand: ['bar'] }).build())
+          .toBe('/my-project1/foo?expand=bar')
+      })
+      it('should support mutliple `expand`s', () => {
+        expect(service.parse({ expand: ['bar', 'baz'] }).build())
+          .toBe('/my-project1/foo?expand=bar&expand=baz')
```
dferber90: The tests I added were written against the public API exclusively.

I got this from Dan

> **For libraries, tests should ideally be written against the public APIs.** This goes against the common mantra of unit testing, but in our experience this helps both ensure that the right thing is being tested, and that it is easy to replace the underlying implementation. It does, however, make debugging test failures harder. - [source](https://hashnode.com/post/what-have-you-learned-after-working-at-facebook-for-almost-two-years-have-you-grown-as-a-developer-and-what-are-some-of-the-key-takeaways-cj7q3gkjx019xkhwujchsrtho)

packages/api-request-builder/test/create-service.spec.js
```diff
@@ -88,61 +89,283 @@ describe('createService', () => {
...
+        endpoint: '/foo',
+        features: [
+          defaultFeatures.query,
+          defaultFeatures.queryOne,
+          defaultFeatures.queryExpand,
+          defaultFeatures.projection,
+        ],
+      }
+      service = createService(options, projectKey)
+    })
```
dferber90: I extracted the service creation into `beforeEach` so that we don't have to repeat it.

docs/sdk/api/apiRequestBuilder.md
adnasa: I think that makes sense, because the reference to these keys are much more clear and that we have a SSOT.
If there is any deviation to what the original documentation, as in if there is a GET query we don't support on the requestBuilder, we should state the as well

packages/api-request-builder/src/create-service.js
```diff
@@ -116,13 +118,22 @@ export default function createService (
       const uri =
         (withProjectKey ? `/${options}` : '') +
         endpoint +
+        // TODO this can lead to invalid URIs as getIdOrKey can return
+        //   "/?customerId", so there can be multiple questionmarks,
+        // same for when `queryParams` and `version` are present
         getIdOrKey(this.params) +
```
adnasa: does `getIdOrKey` throw an error at least if both `id` and `key` are undefined or null?

packages/api-request-builder/src/default-params.js
adnasa: I think in the case of

- where a query is invalid => throw an error
    - e.g `id`(customer-id, product-id) are strings in the API, but the dev can potentially pass in a number. This can eg. stop unnecessary requests to the API.
     - we add a validation step before setting them
- where the query key is not supported
    - I think throwing an Error can be a bit excessive. I would've taken the approach of just omitting the key from the params to construct, and along with it emit a warning via `logger` to inform the consumer about the unsupported key.

packages/api-request-builder/src/create-service.js
```diff
@@ -116,13 +118,22 @@ export default function createService (
       const uri =
         (withProjectKey ? `/${options}` : '') +
         endpoint +
+        // TODO this can lead to invalid URIs as getIdOrKey can return
+        //   "/?customerId", so there can be multiple questionmarks,
```
emmenko: This was introduced in #269, I'm actually not sure why it was implemented like this 🤔

packages/api-request-builder/src/default-params.js
emmenko: I would also simply warn if a key is unknown, it's not really an error.

packages/api-request-builder/src/default-params.js
```diff
@@ -62,23 +63,104 @@ export function getDefaultSearchParams (): ServiceBuilderDefaultParams {
...
+
+const hasKey = (obj: {}, key: string): boolean =>
+  Object.prototype.hasOwnProperty.call(obj, key)
+
+/**
+ * Set the supplied parameters given the current service object.
+ *
+ * @return {void}
+ */
+export function setParams (params: ServiceBuilderParams) {
```
emmenko: Be careful ⚠️

Depending on how the service was configured with the "features" you might have those internal methods or not (`this.xxx`).
This means that not all params can be parsed. I don't think it's a problem, because if you use the service for e.g. products you won't use e.g. `byCustomerId` so it does not make sense to pass it and parse it.

Nevertheless, we need to check that the setter function is defined before calling it

packages/api-request-builder/src/default-params.js
dferber90: @emmenko I think we should throw an error here, because:

The users task is to pass configuration to `RequestBuilder`. He can either do that imperatively, or declaratively (using `parse`).

When doing it imperatively, the library will throw on unknown setters (e.g `service.perPag(4)` will throw as the setter is called `perPage`, not `perPag`). When using the declarative way the same rules should apply: `service.parse({ perPag: 4 }).build()`  should throw as well. I don't see how it would be helpful to allow passing unknown keys. It is very likely to be a bug if that happens. And it should be caught early.

packages/api-request-builder/src/create-service.js
dferber90: No it does not.

```js
function getIdOrKey (params: Object): string {
  if (params.id)
    return `/${params.id}`
  else if (params.key)
    return `/key=${params.key}`
  else if (params.customerId)
    return `/?customerId=${params.customerId}`
  return ''
}
```

I opened https://github.com/commercetools/nodejs/pull/338 to keep track of this.

packages/api-request-builder/src/default-params.js
dferber90: @adnasa 

> where a query is invalid => throw an error

This already happens naturally, as the `parse` function uses the setters behind the scenes. These check the passed values.

> where the query key is not supported

See the answer above.

emmenko: Ah I see what you mean. Makes sense yes 👍

docs/sdk/api/apiRequestBuilder.md
dferber90: @emmenko The `params` are an argument for that one specific `parse` function. They are not used anywhere else. I don't think it should go into the Glossary.

@adnasa I do not understand your comment 😕

packages/api-request-builder/src/default-params.js
wizzy25: Thanks a lot for this PR (loving it). I think the `switch` approach will be better for this and is in my opinion, more concise, and also favours the flat key approach (I think page, `perPage` and `sort` should not be nested)

packages/api-request-builder/src/default-params.js
dferber90: I'm aware of that and deliberately set it up like that.

As I pointed out in https://github.com/commercetools/nodejs/pull/337#discussion_r149602263 the `parse` is the declarative counterpart to the imperative API. So they should behave the same.

When using the imperative API and trying to use `byCustomerId` on a service that does not support it, the user is calling a non-existent `byCustomerId` function and an error will be thrown: `service.byCustomerId(2).build()`.

At the moment the same thing will happen when the user uses the declarative API: `service.parse({ customerId: 2 }).build()`.

So with this implementation they already have the same behaviour. So I would not change anything about it.

> Nevertheless, we need to check that the setter function is defined before calling it

That would be like adding `noop` alternatives to the imperative API in case the a feature is not enabled. So it would fail silently, which is pretty bad. The same applies for the declarative API, which is why I would not add these checks.

dferber90:
Ok, so from my side the only open change request is about the documentation of the `params` argument for `parse`. I do not know what to do about this as the comments were unclear.

Thanks for the reviews so far :)

packages/api-request-builder/src/default-params.js
dferber90: I forgot one downside of the `switch` approach:
- we loose control over the order in which the setters are called as the calls will happen in the order of the keys of the object. `Object.entries` traverses the objects in an arbitrary order, so the application using the library doesn't gain control either. But we probably don't need the control anyways. In case we do, we could always execute some before/after the `switch` statement.

wizzy25: True. And the besides the `id`, `key` and `customerId`, the order in which the setters are called does not really matter. But then, a user should not pass in parameters having more than one of these properties.

wizzy25: At least for now. We can go with the `if` statements, in case future developments make the order of execution necessary

docs/sdk/api/apiRequestBuilder.md
wizzy25: I think the documentation of the params should be in the documentation as it is. The imperative usage is already there and would make it more succinct

adnasa: what I was saying was __yes__ link to the docs please. :)

dferber90: @adnasa bbbbbbbbut these are the docs? I'm even more confused now 😕

dferber90: Anyways, @wizzy25 and I seem to agree to keep it as-is.

dferber90:
Seems like all feedback was addressed. I consider this ready to merge now

docs/sdk/api/apiRequestBuilder.md
emmenko: I'm talking about this page https://commercetools.github.io/nodejs/sdk/Glossary.html

packages/api-request-builder/src/create-service.js
emmenko: It should not throw, those are optional params but the function is always called.

packages/api-request-builder/src/default-params.js
emmenko: I guess this relates to the other "throwing" part. Makes sense yes, thanks for the explanation 👍

emmenko:
@dferber90 please remember how to properly release it when you merge it, thanks! 🙏

https://github.com/commercetools/nodejs/blob/master/CONTRIBUTING.md#releases

docs/sdk/api/apiRequestBuilder.md
dferber90: @emmenko I know. But why would we add a glossary definition for the parameters of a single function of a single module to the global `sdk` glossary?

From the description:

> This is a glossary of the core terms around SDK packages, along with their type signatures

I don't consider the `params` argument we accept in `parse` to be a core term. So it should not go in there unless I'm misunderstanding something.

emmenko: I'm fine also keeping it here for now.

dferber90:
@emmenko I'm not authorized to merge this PR (unless it needs to have two reviews before I can merge it). Not sure yet, we'll see after the other approval.

dferber90:
@wizzy25 @emmenko Since I don't have the rights, can one of you get this in? I'm blocked by this.

tdeekens:
Sorry. @dferber90 I can go ahead and ship this.

emmenko:
@dferber90 that's weird, we should be able to. I guess because of the pending "status checks"??

@tdeekens thanks for taking over!

tdeekens:
No it's not due to status checks, I also had to ask for those rights when working on my first PR here.

emmenko:
Hmm but you have write permissions. Do you need to be an admin instead?

dferber90:
@emmenko Seems like it. But @tdeekens is handling this anyways, because the build tooling was broken as well.

dferber90:
Finally 😍

PR #338, opened 2017-11-07
Bug: building request string contains multiple question marks

dferber90:

#### Summary
Adds failing test case to reproduce error.

As shown by the additional test cases the logic in the `build` function can fail in multiple ways.

#### Description
```
 FAIL  test/create-service.spec.js
  createService
    ✓ should create a full service (4ms)
    ✓ should throw if project key is missing
    ✓ should throw if definition is missing
    ✓ should throw if definition prop (type) is missing (1ms)
    ✓ should throw if definition prop (endpoint) is missing
    ✓ should throw if definition prop (features) is missing
    ✓ should throw if definition prop (features) is not an array
    ✓ should throw if definition prop (features) is an empty array (1ms)
    build
      ✓ include projectkey in uri by default
      ✓ exclude projectkey from uri using flag
      ✓ only base endpoint
      ✓ endpoint with id (1ms)
      ✓ endpoint with customer id
      ✓ endpoint with key
      ✓ endpoint with query params
      ✓ include version in uri
      ✓ full endpoint (1ms)
      with additional services
        ✕ should mix customerId and queryParams (1ms)
        ✕ should mix customerId and version
        ✕ should mix queryParams and version (1ms)

  ● createService › build › with additional services › should mix customerId and queryParams

    expect(received).toBe(expected)

    Expected value to be (using ===):
      "/my-project1/test/?customerId=foo&expand=baz"
    Received:
      "/my-project1/test/?customerId=foo?expand=baz"

      at Object.<anonymous> (packages/api-request-builder/test/create-service.spec.js:109:9)
          at Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (packages/api-request-builder/internal/process/next_tick.js:188:7)

  ● createService › build › with additional services › should mix customerId and version

    expect(received).toBe(expected)

    Expected value to be (using ===):
      "/my-project1/test/?customerId=foo&version=3"
    Received:
      "/my-project1/test/?customerId=foo?version=3"

      at Object.<anonymous> (packages/api-request-builder/test/create-service.spec.js:114:9)
          at Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (packages/api-request-builder/internal/process/next_tick.js:188:7)

  ● createService › build › with additional services › should mix queryParams and version

    expect(received).toBe(expected)

    Expected value to be (using ===):
      "/my-project1/test?expand=baz&version=3"
    Received:
      "/my-project1/test?expand=baz?version=3"

      at Object.<anonymous> (packages/api-request-builder/test/create-service.spec.js:119:9)
          at Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (packages/api-request-builder/internal/process/next_tick.js:188:7)

Test Suites: 1 failed, 1 total
Tests:       3 failed, 17 passed, 20 total
Snapshots:   0 total
Time:        0.152s, estimated 1s
```

I discovered this while working on #337.


dferber90:
To state the issue explicitly:

The `build` function produces invalid URIs as they contain multiple `?`.


For example, it produces `/my-project1/test/?customerId=foo?expand=baz`  when one would expect `/my-project1/test/?customerId=foo&expand=baz`.

emmenko:
Good catch, I guess this was introduced here https://github.com/commercetools/nodejs/pull/269

dferber90:
To avoid any confusion: I am not working on resolving this bug myself. Somebody else should take over.

wizzy25:
@dferber90 Thanks for catching this bug. Would help a great deal if you create issues first before opening pull requests per the [contribution  guidelines](/commercetools/nodejs/blob/master/CONTRIBUTING.md) to make our lives (and issue tracking) easier 😉

dferber90:
I can close this and copy the PR description to an issue if you want that.

But then you have to create a new branch and copy the failing test cases from the PR description back into the test files. Basically ending up where you are right now.

With an open PR of failing test cases you can just start working on the solution right away (and you already have a branch you can work from). This is a pretty standard procedure in OSS development, so I'm pretty surprised by that response, honestly. I tried to make it as easy for you as it gets.

In case the reason is that this leads to problems identifying open bugs: Not sure you're aware, but issues and PRs can be mixed into one view (https://github.com/commercetools/nodejs/issues?utf8=%E2%9C%93&q=label%3A%22Type%3A%20Bug%22%20), so the overview still exists.

I don't want to step on anyones toes here, though. I you want an issue, I can copy everything over to one.

wizzy25:
@dferber90 Not a problem. Your setup does make our lives easy as well, and maybe you misunderstood my intention. It just goes contrary to our workflow. No toes have been stepped on 😊

packages/api-request-builder/src/create-service.js
```diff
@@ -109,29 +109,26 @@ export default function createService (
...
+      if (this.params.id)
+        uri += `${endpoint}/${this.params.id}` +
+          `${queryParams ? `?${queryParams}` : ''}`
+
+      else if (this.params.key)
+        uri += `${endpoint}/key=${this.params.key}` +
+          `${queryParams ? `?${queryParams}` : ''}`
+
+      else if (this.params.customerId)
+        uri += `${endpoint}/?customerId=${this.params.customerId}` +
```
emmenko: Can't you simply put `customerId` also in the `buildQueryString` function like for `version`?

I would also keep the `getIdOrKey` function, makes this thing easier to read. Thanks 🙏

wizzy25: I think that on first glance, it could make the code less understandable because a URI can only have one of `id`, `key` or `customerId` so it makes sense for them to be in the same place

emmenko: The `customerId` is a query string parameter, `id` and `key` are URI parts.

packages/api-request-builder/src/create-service.js
```diff
@@ -132,6 +130,6 @@ function getIdOrKey (params: Object): string {
   else if (params.key)
     return `/key=${params.key}`
   else if (params.customerId)
```
emmenko: This can also be removed, the trailing slash is not important

PR #340, opened 2017-11-09
feat(api-request-builder): add restrictResult feature with onlyIds flag

dferber90:

#### Summary
- adds `restrictResult` feature
- adds `query-restrict-result` which only implements `onlyIds`

#### Description
`product-projection-search` now supports returning ids only instead of full products. This is now represented in `nodejs` by adding a "restrictResult" feature (which currently only supports "onlyIds").

#### Exampe of `onlyIds`

```
## Product Projection Search (onlyIds)
# Fetches a list of products
curl "https://api.commercetools.com/test-with-big-data-44/product-projections/search?onlyIds=true" \
     -H 'Authorization: Bearer xxxx'
```

will return

```json
{
  "limit": 20,
  "offset": 0,
  "count": 20,
  "total": 33,
  "results": [
    {
      "id": "758185f8-1747-4496-a77d-87e3171d3096"
    },
    {
      "id": "34185700-69b2-43c7-a0bd-b3ce0924df84"
    },
    {
      "id": "32a24c0f-99c0-40cf-8c50-59ab7dde4d22"
    },
    // ....
  ],
  "facets": {}
}
```

#### Todo

- Tests
    - [x] Unit
    - [x] Integration
    - [x] Acceptance
- [x] Documentation


dferber90:
The feature proposed here should be added to #337 as well. 

But as we're unsure which PR get's merged first it's not feasible to add it yet. We'll probably create a follow-up PR.

dferber90:
The build keeps on failing on node v4 only with

```
npm ERR! Linux 4.9.6-040906-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v4.8.6/bin/node" "/home/travis/.nvm/versions/node/v4.8.6/bin/npm" "install"
npm ERR! node v4.8.6
npm ERR! npm  v2.15.11
npm ERR! code EPEERINVALID
npm ERR! peerinvalid The package eslint-plugin-jsx-a11y@5.1.0 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer eslint-config-commercetools@6.0.0 wants eslint-plugin-jsx-a11y@>=3
npm ERR! peerinvalid Peer eslint-config-airbnb@15.1.0 wants eslint-plugin-jsx-a11y@^5.1.1
npm ERR! Please include the following file with any support request:
npm ERR!     /home/travis/build/commercetools/nodejs/npm-debug.log
The command "eval npm install  " failed. Retrying, 2 of 3.
```

Not sure it is related to my changes or an issue in general.

dferber90:
@emmenko It's only available in `product-project-search` specifically. That's why it is configured as its own thing, instead of in `querySearch`. Otherwise every service supporting `querySearch` would also support `onlyIds`. But that's not what we want.

emmenko:
As far as I know, only products have `search` at the moment. I don't have a strong opinion on this, maybe just making it a bit more clear that it's only about the search. I'll leave it up to you.

dferber90:
```js
  productProjectionsSearch: {
    type: 'product-projections-search',
    endpoint: '/product-projections/search',
    features: [
      features.search,
      features.queryOne,
      features.queryExpand,
      features.projection,
      features.restrictResult,
    ],
  },
  productProjectionsSuggest: {
    type: 'product-projections-suggest',
    endpoint: '/product-projections/suggest',
    features: [
      features.search,
      features.suggest,
      features.queryOne,
      features.projection,
    ],
  },
```

This one also uses search. It seems that that feature doesn't even return something with `ids`. So it would break there. Having it as its own thing gives more control here.

> maybe just making it a bit more clear that it's only about the search

Do you mean by adding a code comment? Doesn't it only being set on `product-projection-search` express this? No other service explains itself, so I'd rather not comment it.

dferber90:
Added `onlyIds` to `service.parse` now. This is ready to be merged now (from my side).

packages/api-request-builder/src/create-service.js
```diff
@@ -103,6 +104,12 @@ export default function createService(
           ...queryProjection,
         }
 
+      if (feature === defaultFeatures.restrictResult)
+        return {
+          ...acc,
```
tdeekens: So not part of this PR but every time I see `acc` I shudder and think in this case it's just `feature`.

dferber90: But `feature` is already taken and a string :( Shouldn't it be `service` anyways, since that's the object we're putting together here?

tdeekens: Yeah makes even more sense.

dferber90:
It's not clear anymore whether this should actually go into the SDK as it's an internal feature.
And we have a way to "work around" this in the MC.

@wizzy25  @emmenko You can hold off reviewing for now.

I might close the PR after checking whether the workaround is good enough later. If a review is required I'll ping you again.

emmenko:
What do you mean it’s an internal feature? From the API?

dferber90:
@emmenko Internal to CT. It is not documented, so it is not "public". Support for it could be dropped when the MC does not need it anymore.

emmenko:
I see, thanks for the info

dferber90:
@wizzy25 Why was this merged? I should have made the "do not merge" more prominent.

I talked to @yanns and this should not have been merged as it's an internal feature only. We should remove it again before anybody starts using it.

tdeekens:
Lets just revert this feature merge before the next release. @wizzy25 can do that early next week.

PR #341, opened 2017-11-13
chore(travis): upgrade to latest yarn version

tdeekens:
#### Summary

This installs the latest yarn version on TravisCI's Trusty images. You can verify that it's installed in the logs during a CI runs.

This fixes master which currently fails as dependencies on node@4 are falsely installed with npm@2.


tdeekens:
This is waiting for the new token to publish things. `buffer-shim` is installed with the latest version of `yarn` which seemed to be the reason for moving to `npm`. After we have the new token, we will merge this, to then release anything pending, to then merge the declarative parsing and then the tooling refactoring.

PR #344, opened 2017-11-13
chore(travis): fix release with npm install on after_success

tdeekens:
#### Summary

Installing dependencies before releasing with npm.


wizzy25:
A solution I think of is to add the buffer-shims to the `package.json`. That way, it will not be unbuilt with `npm prune`

Issue #223, opened 2017-07-03
Duplicate publishing of modules


hisabimbola:
### Description

A couple of times we have encountered an issue where a module is bumped twice wrongfully. This happens mostly when a module was released locally because [lerna-semantic-release](https://github.com/atlassian/lerna-semantic-release) failed to do its job in CI. The sad thing is upon another release or merging of a PR, semantic release goes through the commits again and publishes the module again.

After several days of debugging, I discovered that `lerna-semantic-release` depends on the `gitHead` from npm as a base of searching for relevant commits to use for publishing, if the `gitHead` is missing, it analyses all the commits in history and publishes accordingly.

Since we now know that missing `gitHead` was the cause of `lerna-semantic-release` error. I went further to check why npm decides not to include the `gitHead` when publishing and I discovered here that npm *always* assumes that we are running in the `npm publish` command in the [root of the repository](https://github.com/npm/npm/blob/3c3341b41a52c692fde53a126e9227f12ffed22a/node_modules/read-package-json/read-json.js#L293) (i.e repo where the `.git` folder lies). Because we are using a monorepo structure, to publish manually we `cd` into the module's subfolder in the [packages](/commercetools/nodejs/tree/c986b80e39fe0a6540a7c446bad3abe7efe6cc80/packages) folder and since that is not where the `.git` folder resides, npm publishes without a `gitHead` causing `lerna-semantic-release` not to be able to figure out that a `publish` had occur before.

### The fix
The fix(hack) was to symlink the `.git` folder to the package folder and run `npm publish` then npm is able to use the appropriate `gitHead` and this fixes the issue. This hack is also what `lerna-semantic-release` uses in it's [perform command](https://github.com/atlassian/lerna-semantic-release/blob/c48c50db5844cc9e1f5a1d2752cdadff84a8d426/packages/lerna-semantic-release-perform/index.js#L60).

To publish a module now, the `npm run manual-release` command should be use and this runs the `publish-manually` [script](/commercetools/nodejs/tree/c986b80e39fe0a6540a7c446bad3abe7efe6cc80/scripts/publish-manually) in the [scripts](/commercetools/nodejs/tree/c986b80e39fe0a6540a7c446bad3abe7efe6cc80/scripts) folder

PR #346, opened 2017-11-14
Migrate from TravisCI to CircleCI

tdeekens:
#### Summary

We are planning to have most of our projects running on CircleCI. It allows for more flexible configuration and speedier builds at times. This is an attempt to migrate out setup with TravisCI over.

I will try to elaborate on decisions/attempts in the code. _Note_: this is quite an early stage PR but I'd like to chat about the configuration and setup a bit and have some 👁  on it.

##### Todos

- [ ] Migrate environment variables over
    - Their are hidden in TravisCI's UI for me. Either we regenerate or somebody with more rights should move them
- [ ] ~Verify and debug parallelism of build (lint, test etc)~
- [x] Test on multiple environments (node 4 - 8)


.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
+version: 2
+defaults: &defaults
```
tdeekens: Found this pattern after investigating CircleCI configs on other projects. Reduces the repetition and noise in the environment setup of jobs.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+  working_directory: ~/nodejs
+
+jobs:
+  install:
+    <<: *defaults
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - v1-yarn-cache-{{ checksum "yarn.lock" }}
```
tdeekens: We had "issues" with the cache before. By just appending a version we can easily "bust" the cache.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+          command: yarn install --pure-lockfile
+      - save_cache:
+          key: v1-yarn-cache-{{ checksum "yarn.lock" }}
+          paths:
+            - node_modules
+      - run:
+          name: Update greenkeeper-lockfile
+          command: ~/.yarn/bin/greenkeeper-lockfile-update
+      - run:
+          name: Upload greenkeeper-lockfile
```
tdeekens: I don't think this has to run at the complete end of a chain as mentioned in their docs as dependencies won't change after the install.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+  lint:
+    <<: *defaults
+      - run:
+          name: Run linters
+          command: yarn lint
+  test_unit:
+    <<: *defaults
+      - run:
+          name: Run tests
+          command: yarn test:coverage:ci --maxWorkers=10
```
tdeekens: maxWorkers is a copy of our setup with our frontend. Otherwise Jest will spawn too many workers making the build allocate too many resources.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+  test_integration:
+    <<: *defaults
+      - run:
+          name: Run integration tests
+          command: yarn test:integration
+  prepare_release:
+    <<: *defaults
+      - run:
+          name: Run prepare release
+          command: ./.circleci/prepare-release-packages.sh
```
tdeekens: This scripts are simplified now as we can integrate with the `filters` of CircleCI instead of switching over env vars.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+  perform_release:
+    <<: *defaults
+      - run:
+          name: Run perform release
+          command: ./.circleci/perform-release-packages.sh
+
+workflows:
+  version: 2
+  linting:
```
tdeekens: I felt like using progressive tense (hope that's correct :smile:) to separate naming of job from workflows. I often got lost in other configs and felt this might help. Maybe it makes sense. I saw the job as a blueprint to be instantiated/run in a job.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+    jobs:
+      - install
+      - lint
+  unit_testing:
+    jobs:
+      - test_unit
+          requires:
+            - install
+  integration_testing:
```
tdeekens: The idea is to fan out on install into linting and unit testing to then fan in again on integration testing which is followed by releasing.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+              only:
+                - master
+          requires:
+            - install
+            - lint
+            - test_unit
+  releasing:
+    jobs:
+      - prepare_release
+          context: org-global
```
tdeekens: This should allow us to have access the env variables if I read the docs correct.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+          requires:
+            - integration_testing
+          filters:
+            branches:
+              only:
+                - master
```
tdeekens: This replaces what we had in the scripts themselves before basically ensuring that we only release from `master`.

package.json
```diff
@@ -17,10 +17,10 @@
     "pretest:integration": "./scripts/install-npm-packages-for-integration-tests",
     "test:integration": "cross-env NODE_ENV=test jest --config integration-tests/.jestrc.json",
     "format": "prettier --write '{packages,integration-tests,types}/**/*.js'",
-    "coverage": "npm run build && npm test -- --no-watchman --coverage",
-    "coverage:ci": "npm test -- --no-watchman --coverage && codecov",
-    "semantic-release": "lerna-semantic-release pre && lerna-semantic-release post && lerna-semantic-release perform",
-    "manual-release": "./scripts/publish-manually",
+    "test:coverage": "npm run build && npm test -- --no-watchman --coverage",
```
tdeekens: Renaming these makes sense to me as they're nicely grouped then when e.g. running an `npm run`.

package.json
```diff
@@ -17,10 +17,10 @@
...
+    "test:coverage:ci": "npm test -- --no-watchman --coverage && codecov",
+    "release:semantic": "lerna-semantic-release pre && lerna-semantic-release post && lerna-semantic-release perform",
+    "release:manual": "./scripts/publish-manually",
```
tdeekens: Same for these.

.circleci/config.yml
```diff
@@ -0,0 +1,94 @@
...
+    <<: *defaults
+      - run:
+          name: Run perform release
+          command: ./.circleci/perform-release-packages.sh
+
+workflows:
+  version: 2
+  linting:
+    jobs:
+      - install
```
tdeekens: Note to self: these could probably be split into `installing` and `linting` requiring `installing`.

tdeekens:
One of the "funniest" tasks will be - if we want to migrate - how to run our tests on multiple node versions. Seems like separate jobs with different images could do the trick. Maybe we can have more defaults we can then "spread" into the jobs themselves such as

```yaml
node4: &node4
  docker:
    - image: circleci/node:4
  working_directory: ~/nodejs

node8: &node8
  docker:
    - image: circleci/node:8
  working_directory: ~/nodejs

jobs:
  install:
    <<: *node4
    steps:
```

Even cooler would be if we could "parametrize" jobs in a workflow.

tdeekens:
After talking to @wizzy25 we agreed that

1. Running on different node versions is needed
    - Obviously still up for debate if others have opinions about it
2. We should investigate the performance and of having concurrent linting/testing over a joint step

.circleci/config.yml
emmenko: Have you tried this already? I fail to see how you can have dependencies between different workflows. As far as I know you can do that within the same workflow, not across different ones.

What is the idea of doing it like this? 🤔

.circleci/perform-release-packages.sh
```diff
@@ -0,0 +1,16 @@
...
+echo "https://$RELEASE_GH_TOKEN:@github.com/$TRAVIS_REPO_SLUG.git" > .git/credentials
+
+git config --global user.email "npmjs@commercetools.com"
+git config --global user.name "CT Release Bot"
+git config --global push.default simple
+
+npm config set //registry.npmjs.org/:_authToken=${NPM_TOKEN} -q
+
+echo "Running release:semantic"
+yarn release:semantic
```
emmenko: I would honestly try to get rid of this and simply use `lerna publish`. To be discussed...

emmenko: And I don’t mean as an automatic step but as a manual step

.circleci/config.yml
emmenko: Not really. The “org-global” gives you access to env variables defined globally at the organization level, so each project can use the same variables.
However you can define env vars for specific projects, and that’s what we most likely need in this case.

emmenko:
> Running on different node versions is needed

Yes we should still support older node versions. Maybe we should document this and have the last 2 LTS versions supported plus the previous one deprecated until a certain date (eg in this case v4)

> Persisting the built assets to the workspace to share them across builds (if that makes sense)

What assets do you mean?

.circleci/perform-release-packages.sh
emmenko: See also #223, which didn’t really fix the problem. We have the tool there, we should just use it

tdeekens:
Thanks for the early feedback.

1. I am really just getting my feet wet with CircleCI configs for 2 hours now. I'll fix up things. Thanks for some heads up
2. I am adding testing node 4,6,8 for now using aliases and default configs makes it quite ok to work with

> What assets do you mean?

The build `dist` assets. I wonder if we can gain some speed for generating them once and reusing them whenever needed. But also for another step once we're there.

> Just use lerna publish

I'd like to do this last. Learned that painfully with the rollup PR that these "unrelated" changes pile up quite easily.

tdeekens:
@emmenko this is my next "try" I validated the config using CircleCI's CLI.

.circleci/perform-release-packages.sh
tdeekens: Lets have a ScreenHero chat about this. How to proceed etc, I can then chat with @wizzy25 or he joins in.

CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ To ensure the lock file is up-to-date with the new versions, it's recommended to
 We use [semantic release](https://github.com/semantic-release/semantic-release) to automatically do releases based on the commit message.
 Since we are using [lerna](https://github.com/lerna/lerna) we need to specify in the commit description which packages might be affected by the release.
 
-If for some reason you have to release a module manually, please run `npm run manual-release` to publish the module. The reason for this are well stated [here](https://github.com/commercetools/nodejs/issues/223)
+If for some reason you have to release a module manually, please run `npm run release:manual` to publish the module. The reason for this are well stated [here](https://github.com/commercetools/nodejs/issues/223)
```
luhmann: Couldn't this be a yarn command too?

package.json
luhmann: Same here: shouldn't we use `yarn` consistently?

.circleci/config.yml
```diff
@@ -0,0 +1,165 @@
...
+  docker:
+    - image: circleci/node:6
+  working_directory: ~/nodejs
+node8Defaults: &node8Defaults
+  docker:
+    - image: circleci/node:8
+  working_directory: ~/nodejs
+
+aliases:
+  - &restore_yarn_cache
```
luhmann: nice

CONTRIBUTING.md
tdeekens: Yeah, that's another thing so fix up for me.

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+            - lint
+      - test_unit_node_8:
+          requires:
+            - lint
+      - test_integration:
+          context: org-global
+          filters:
+            branches:
+              only:
+                - master
```
wizzy25: This means the integration tests will only run on `Master` ?

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+  test_unit_node_8:
+    <<: *node8Defaults
+    steps:
+      - checkout
+      - restore-cache: *restore_yarn_cache
+      - run: *yarn_install
+      - save_cache: *save_yarn_cache
+      - run: *unit_test_with_coverage
+  test_integration:
+    <<: *node8Defaults
```
wizzy25: I still think running the integration tests in multiple environments will be a good idea

tdeekens: Yeah maybe up for debate. I don't see what vector of bugs it avoids running it on multiple envs after doing that for the unit tests.

.circleci/config.yml
tdeekens: Yeah, we had this before https://github.com/commercetools/nodejs/blob/master/.travis/scripts.sh#L13. But I think that was only disabling them for PRs coming from forks as Travis does not expose env vars onto them. I can enable them once we migrate the env vars over.

wizzy25: okay. Makes sense

.circleci/config.yml
wizzy25: I actually haven't experienced such scenario :1st_place_medal:

tdeekens:
Ok, this starts to get somewhere. 

screen shot 2017-11-14 at 15 10 50

For some reason we fail on node 4.

wizzy25:
circle now passes on node v4 💯

tdeekens:
Neato, I like seeing what job actually fails on GitHub. Plus, a run and all node versions takes around 4-5 minutes in total. A build on `master` with TravisCI is currently at around 25-30 minutes.

CONTRIBUTING.md
tdeekens: So chatted to @wizzy25 about this and we agreed not to change that for now. This is a public repository and we should not be as prescriptive to force people using `yarn`. In whatever environment we control we can use it: e.g. on CircleCI but otherwise we felt it's a big harsh on a public project.

tdeekens:
## Todo

- [x] Investigate performance and parallelism of build
   - It's one workflow now with unit testing running in parallel
   - Running this takes about 5 minutes which is good for now
   - Caching the build assets makes little sense as the tasks run in parallel and we actually want to ensure that we build on all node envs
   - The yarn cache is updated on and used where ever possible
- [x] Remove the building before testing a package (as per @wizzy25 request)
- ~[ ] Rename all `npm run` to `yarn`~
   - Not to be done here as it's a public repo and we want to not enforce `yarn` on anyone
   - For environments we control it's fine however
- [ ] Migrate over env variables from TravisCI to CircleCI 
   - This would be one for @emmenko 
   - After we should remove the `branches only master` restriction on the integration tests and have them run on branch builds
- [ ] ~Manual releasing using `lerna publish`~
   - I would like to do that separately if that's fine with everyone to keep this a focussed migration

tdeekens:
The last commits remove unneeded building e.g. during linting so we only build once per node environment. Let's see if that works.

tdeekens:
Will rebase after review to retrigger the build to get the updated status checks here.

README.md
```diff
@@ -6,8 +6,8 @@
 </p>
 
 <p align="center">
-  <a href="https://travis-ci.org/commercetools/nodejs">
-    <img alt="Travis CI Status" src="https://img.shields.io/travis/commercetools/nodejs/master.svg?style=flat-square&label=travis">
+  <a href="https://circleci.com/gh/commercetools/nodejs">
+    <img alt="Travis CI Status" src="hhttps://circleci.com/gh/commercetools/nodejs.svg?style=shield&circle-token=abc">
```
wizzy25: Typo: ~~hhttps...~~ https...

.circleci/config.yml
```diff
@@ -0,0 +1,176 @@
...
+      - run: *yarn_install
+      - run: *yarn_bootstrap
+      - run: *yarn_build
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Performing release
+          command: ./.circleci/perform-release-packages.sh
+
+workflows:
+  version: 2
```
wizzy25: The version was already specified on [L51](https://github.com/commercetools/nodejs/pull/346/files#diff-1d37e48f9ceff6d8030570cd36286a61R51). Does it need to be specified here again?

README.md
wizzy25: And also change the image alternate is still "_Travis_ CI Status"

.circleci/config.yml
tdeekens: Yeah, I saw that on other configs. You specify the version for the build config itself and the workflows.

README.md
tdeekens: de431f1

PR #348, opened 2017-11-15
feat(csv-parser-orders): Support parcel.items in deliveries CSV

### LINKED ISSUE #347: [csv-parser-orders] Add parcel.items to deliveries parser
### Description
Add possibility to specify parcel items in CSV file with the following column format:
```csv
parcel.items
itemId:quantity;itemId2:quantity2
```

junajan:
This PR adds a support for parcel.items in deliveries CSV
Resolves #347

affects: @commercetools/csv-parser-orders

#### Summary

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [x] Documentation


packages/csv-parser-orders/src/parsers/deliveries.js
```diff
@@ -252,9 +259,24 @@ export default class DeliveriesParser extends AbstractParser {
...
       objectPath.set(parcel, transitionMap[fieldName], fieldValue)
     })
 
     return parcel
   }
+
+  static _parseParcelItems(itemsString) {
+    if (!itemsString) return []
+
+    return itemsString.split(';').map(itemString => {
```
wizzy25: Maybe destructuring...
```js
    return itemsString.split(';').map(itemString => {
      const [id, quantity] = itemString.split(':')
      return { id, quantity }
    })
```

junajan: looks nice :-)

tdeekens: We could try to give things more meaningful names

```js
static _parseParcelItems(parcelItemsAsString) {
    if (!parcelItemsAsString) return []

    return parcelItemsAsString.split(';').map(parcelItemAsString => {
      const [id, quantity] = parcelItemAsString.split(':')
      return { id, quantity }
    })
}
```

junajan: @tdeekens - I would not repeat long variable names - the method is short and it already sais that it parses parcel item so it is not so hard to discover, that the input variable `itemString` is a parcel item.

junajan: Also with longer variable names, you will hit another problem like max line length and that there will be a lot of text which can be confusing and hard to read so I would vote for shorting wherever it is possible and meaningful.

junajan: @wizzy25 - I refactored code 868d6ac .. please let us know what do you think about that variableName.

tdeekens: BTW with the recent addition of prettier to the codebase just install e.g. atom-prettier or run `yarn format` and all code will be formatted for you.

junajan: don't know this magic .. will try it, thanks

tdeekens: I really recommend an IDE integration to whatever you use: VCCode, Atom etc. It will format everything on save. Once you been down that path you never want back 😄 .

wizzy25: I'm all for explicit names, and `yarn format` makes me no longer worry about formatting and line length... Thanks for that @tdeekens

junajan: Variable names updated here b934293

PR #351, opened 2017-11-15
chore(circle): require unit tests before releasing

tdeekens:
#### Summary

Makes sure unit tests on all platforms pass before performing an release after integration tests are green.


tdeekens:
Waiting with this until we hopefully made sure that our release token on CircleCI has sufficient rights.

PR #352, opened 2017-11-15
chore(package): remove semantic-release

tdeekens:
#### Summary

Semantic release in accordance with @emmenko should be removed. It provides quite a few benefits but with a low count of contributors comes with a high cost of setup and automation.

Currently, we can not get CircleCI to be allowed to push back into the repository without having to add an custom SSH key which feels like reason enough now to make the release process manual.

This means that contributors release manually with lerna which helps to

1. Determine which packages changed
2. Determine a good guess for a version bump (with a cli flag)


tdeekens:
What do we do about `yarn docs:publish`? /cc @emmenko

package.json
```diff
@@ -21,8 +21,7 @@
     "test:coverage": "npm run build && npm test -- --no-watchman --coverage",
     "test:ci": "npm run test -- --no-watchman --maxWorkers=10",
     "test:coverage:ci": "npm run test:ci -- --coverage && codecov",
-    "release:semantic": "lerna-semantic-release pre && lerna-semantic-release post && lerna-semantic-release perform",
-    "release:manual": "./scripts/publish-manually",
+    "release": "npm run build && lerna publish --registry=https://registry.npmjs.org/ --conventional-commits",
```
wizzy25: Maybe we should add `docs:publish` here?
`npm run build && lerna publish --registry=https://registry.npmjs.org/ --conventional-commits && npm run docs:publish`

tdeekens: Ah fair point.

tdeekens: However, I think it should only be a `postrelease` step cause it otherwise won't work as `lerna publish` triggers an interactive command.

wizzy25: It should still work even if the command is interactive. As long as it exits successfully, control will then be passed to the next command (`npm run docs:publish` in this case)

package.json
```diff
@@ -21,8 +21,7 @@
     "test:coverage": "npm run build && npm test -- --no-watchman --coverage",
     "test:ci": "npm run test -- --no-watchman --maxWorkers=10",
     "test:coverage:ci": "npm run test:ci -- --coverage && codecov",
-    "release:semantic": "lerna-semantic-release pre && lerna-semantic-release post && lerna-semantic-release perform",
-    "release:manual": "./scripts/publish-manually",
+    "release": "npm run build && lerna publish --registry=https://registry.npmjs.org/ --conventional-commits && npm run docs:publish",
```
emmenko: I don't think the `docs:publish` should be included in the release. We can actually always run this on master.

tdeekens: 1. Releasing should always happen from `master` only with the workflow
   - Feature branch gets merged
   - Pull master and release
2. It won't be included in the release as lerna publish will commit a "Publish" commit

emmenko: I mean that we should build the docs as a continuous step on master, not include it when we manually release something.

emmenko: Alright, all good 👍

PR #361, opened 2017-11-21
chore(build): add default commit message

### LINKED ISSUE #360: [build/release] lerna commit message
### Description

Add a default commit message to lerna.json to enable release. Currently, running `yarn release` fails at the publish step because the commit message doesn't pass the commit message hook

wizzy25:
#### Summary

resolves #360

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -14,5 +14,10 @@
     },
     "cacheDir": ".changelog",
     "ignoreCommitters": ["greenkeeper"]
+  },
+  "command": {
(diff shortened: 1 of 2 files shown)


lerna.json
```diff
@@ -14,5 +14,10 @@
     },
     "cacheDir": ".changelog",
     "ignoreCommitters": ["greenkeeper"]
+  },
+  "command": {
```
tdeekens: Very nice catch!

PR #363, opened 2017-11-21
chore(build): refactor package.jsons

### LINKED ISSUE #362: [build] set proper node environments
### Description

Currently, we always build with the `NODE_ENV` set to `cli`; almost every package had this: 
```json
"scripts": {
    "build": "cross-env NODE_ENV=cli babel src --out-dir lib"
  }
```
With the move to rollup and refactor of the .`babelrc` file in #324 the current node version on a users machine is set as the `NODE_ENV` babel uses for transformation. This causes compatibility issues with releases, as packages released on the "current" node version may fail on an earlier version.

The build setup should be dynamic with regards to the necessary environment

wizzy25:
affects: @commercetools/csv-parser-discount-code, @commercetools/csv-parser-orders,
@commercetools/csv-parser-price, @commercetools/discount-code-exporter,
@commercetools/discount-code-generator, @commercetools/discount-code-importer,
@commercetools/get-credentials, @commercetools/inventories-exporter, @commercetools/price-exporter,
@commercetools/product-exporter

resolves #362


packages/csv-parser-discount-code/package.json
```diff
@@ -28,7 +28,7 @@
     "lib"
   ],
   "scripts": {
-    "build": "cross-env NODE_ENV=cli babel src --out-dir lib"
+    "build": "babel src --out-dir lib"
```
tdeekens: So this would mean that in the `.babelrc` we don't get to use the `cli` environment anymore and it will default to `development` as the env key will be taken from `process.env.BABEL_ENV`, when this is not available then it uses `process.env.NODE_ENV` if even that is not available then it defaults to "development". This means we transpile with our settings for rollup builds. Which is fine and will solve the issue. However, then we can get rid of the `cli` option in the `.babelrc` overall? /cc @emmenko

wizzy25: It wouldn't default to `development` as we will instead set the environment variables [here for development](https://github.com/commercetools/nodejs/pull/363/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R9) and [here for production](https://github.com/commercetools/nodejs/pull/363/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R24)

.babelrc
```diff
@@ -88,22 +92,20 @@
         ]
       ]
     },
-    "production": {
+    "cli-production": {
```
tdeekens: Why do we postfix this? Do we have a `cli-development`?

wizzy25: No.
But there has to be separation between rollup and non-rollup packages for production

wizzy25:
[Rollup requires that the `modules` field in `.babelrc` be set to false](https://github.com/rollup/rollup-plugin-babel#modules). This wasn't a problem because each package always set the `NODE_ENV` to `cli`, while the rollup packages used `development` and `production`. Setting the environment in the root `package.json` leads to module problems as the non-rollup packages need to use `commonjs` modules.
The solution here is to separate the rollup environment presets from the non-rollup preset

.babelrc
wizzy25: @tdeekens The packages that use rollup cannot use the same presets as the ones that don't for both production and development. I'd have to rename them somehow to make it clearer.

wizzy25:
Okay. I think this is good for a merge. cc @tdeekens

.babelrc
```diff
@@ -29,19 +30,22 @@
         ]
       ]
     },
-    "test": {
+    "rollup": {
```
tdeekens: We are mixing concerns here a little. Target and tool we're using for bundling which feels odd.

wizzy25: I would like 3 environments as well. But doing that opens an entire can of worms. I also feel this naming could be better. Do you have any suggestions maybe?

PR #368, opened 2017-11-28
fix(api-request-builder): adds the whereOperator to the knownKeys array

damarna85:
#### Summary
Include the `whereOperator` as known key in the corresponding `knownKeys` array.

#### Description
When passing the `whereOperator` as param we were getting the following error:
image

Basically the `whereOperator` key is not in the `knownKeys`array making it useless. This PR just fixes that by adding the operator in the array.

### DIFF
--- a/packages/api-request-builder/src/default-params.js
+++ b/packages/api-request-builder/src/default-params.js
@@ -115,6 +115,7 @@ export function setParams(params: ServiceBuilderParams) {
     'filterByFacets',
     'searchKeywords',
     'where',
+    'whereOperator',
     'version',
   ]
   Object.keys(params).forEach((key: string) => {


tdeekens:
I think @wizzy25 should have a "final" go over this and publish it.

damarna85:
I will add him to be in the flow :)

dferber90:
I think this commit is missing an `affects` message?

wizzy25:
@dferber90 With the shift to manual releases, the `affects` message is not as necessary anymore

damarna85:
@wizzy25 Is this fix deployed? We need to use it and we cannot get this updated in the library. We need this feature since is blocking us to continue developing. Am I missing something here to get the right version?

cc @tdeekens @dferber90

tdeekens:
I released this to unblock development. Sadly lerna made me release all many components but I trust the tool. Maybe something for later to keep an eye on.

PR #375, opened 2017-12-04
feat(sync-actions/product-discounts): add validFrom and validUntil

### LINKED ISSUE #374: Add `validFrom` and `validUntil` to product discounts sync
### Description

This platform has the notion of time based validity of discount for cart discounts. This notion will be added to product discounts too. The sync-actions package should keep parity with these update actions.

tdeekens:
#### Summary

⚠️  Please do not merge yet until this feature is available in the platform ⚠️ 

Closes #374


tdeekens:
I've verified with platform developers that this has been released but is not documented yet but will be today. I will release this now with confidence as this is the same change as with discount codes.

PR #378, opened 2017-12-04
feat(query-id): add ability to filter by cart id

damarna85:
#### Summary
adds the `cartId` param as a valid one to filter like we already have with `customerId` in the `queryOne` feature.

#### Description
we can call the `shipping-method` endpoint by passing a `cartId` param with the cart id in order to get the available shipping methods for the cart. See [here](https://dev.commercetools.com/http-api-projects-shippingMethods.html#get-shippingmethods-for-a-cart) the docs.

This PR adds the possibility to send as param a `cartId` in order to be able to use this feature when retrieving shipping methods.

- Tests
    - [x] Unit
- [x] Documentation

### DIFF
--- a/packages/api-request-builder/src/query-id.js
+++ b/packages/api-request-builder/src/query-id.js
@@ -19,6 +19,12 @@ export function byId(id: string): Object {
         'You cannot use both `byId` and `byCustomerId`.'
     )
 
+  if (this.params.cartId)
+    throw new Error(
+      'A cartId for this resource has already been set. ' +
+        'You cannot use both `byId` and `byCartId`.'
+    )
+
   this.params.id = id
   return this
 }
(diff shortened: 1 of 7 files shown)


damarna85:
What do you mean @tdeekens ?

tdeekens:
All cool. Found out [here](https://github.com/commercetools/nodejs/blob/ab4edf5bb773b32a99343059e5fd3f9d676c2ed6/packages/api-request-builder/src/default-params.js#L136) that it's already added there.

packages/api-request-builder/src/query-id.js
```diff
@@ -62,3 +68,23 @@ export function byCustomerId(custId: string): Object {
...
+ * @throws If `id` is missing.
+ * @return {Object} The instance of the service, can be chained.
+ */
+export function byCartId(cartId: string): Object {
+  if (!cartId) throw new Error('Required argument for `byCartId` is missing')
+  if (this.params.id)
+    throw new Error(
+      'An ID for this resource has already been set. ' +
+        'You cannot use both `byId` and `byCartId`.'
+    )
```
dferber90: What happens when you set `byCartId` and `byCustomerId`? Are they mutually exclusive?

damarna85: I cannot figure out a case where a combination of a cart id and a customer id filter is valid to be honest. Not sure if we should restrict the case though.

wizzy25: Ideally, if a customer has multiple carts (which is possible, at least on the API level), having both query parameters returns the cart that belongs to the customer. This same cart will be returned if queried with only the `cartId`, so I think we should not restrict it as it is a valid query parameter.

dferber90: Great, thanks for clarifying!

damarna85:
Ok thanks for the quick feedback. Since we need this in one of the features we are working on, do you have any estimation when this can be published and ready to be used? Thanks!!

PR #379, opened 2017-12-05
docs(sync-action): Correct arguments in buildActions method

### LINKED ISSUE #365: Documentation arguments switched
### Description
The docs state `buildActions(before, now)` but [in the code](https://github.com/commercetools/nodejs/blob/master/packages/sync-actions/src/utils/create-build-actions.js#L6
) it is the other way around.

[The docs](https://commercetools.github.io/nodejs/sdk/api/syncActions.html) should be fixed in all these places.

--- 

On a side-note, I also found it weird that the examples are not minimal, but entangle usage with the client.

Minimal example would be

```js
import { createSyncProducts } from '@commercetools/sync-actions';

const before = {"name":{"en":"Hajo"}}
const now = {"name":{"en":"ojaH"}}

const syncProducts = createSyncProducts();
const syncActions = syncProducts.buildActions(now, before)
console.log(syncActions)
// -> [{ "action": "changeName", "name": { "en": "ojaH" } }]
```

Minimal examples are important because its easier to understand one thing at a time for users. There can be another example/guide which shows the usage of sync-actions with the client.

junajan:
#### Summary
This PR fixes an incorrect order or `buildActions` method arguments in docs.
Fixes #365

### DIFF
--- a/docs/sdk/Glossary.md
+++ b/docs/sdk/Glossary.md
@@ -99,7 +99,7 @@ type UpdateAction = {
   [key: string]: any;
 }
 type SyncAction = {
-  buildActions: (before: Object, now: Object) => Array<UpdateAction>;
+  buildActions: (now: Object, before: Object) => Array<UpdateAction>;
 }
 
 type ActionGroup = {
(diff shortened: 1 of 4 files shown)


junajan:
No problem @wizzy25 - updated here 6abc5ef

PR #381, opened 2017-12-13
feat(api-request-builder): Query shippingMethods by location

### LINKED ISSUE #370: Add support to fetch shipping methods by location
### Description
The API allows us to filter shipping methods by calling the following endpoint:
`/{projectKey}/shipping-methods`
and passing as param the `cartId` we want to use or passing as params the `country`, `region` and `currency` code to get back the shipping methods that can be applied to the cart/location passed. Do you know if this is already available in the sdk somehow? I could not find anything related to this particular case.

Thanks!

### Expected Behavior
Being able to filter shipping methods by cart id or location.

### Current Behavior
Do not find that case.

### Possible Solution
¿Adding that case in the api request builder?

junajan:
This PR adds a possibility to query `shippingMethods` using location params described here: http://dev.commercetools.com/http-api-projects-shippingMethods.html#get-shippingmethods-for-a-location
Resolves #370 

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [x] Documentation


packages/api-request-builder/src/create-service.js
```diff
@@ -66,6 +67,7 @@ export default function createService(
           ...acc,
           ...query,
           ...queryPage,
+          ...queryLocation,
```
wizzy25: Why do you add this here?

packages/api-request-builder/src/create-service.js
```diff
@@ -74,6 +76,13 @@ export default function createService(
           ...queryId,
         }
 
+      if (feature === defaultFeatures.queryLocation)
+        return {
+          ...acc,
+          ...queryLocation,
+          ...queryExpand,
```
wizzy25: Why do you also add this here?

packages/api-request-builder/test/create-service.spec.js
```diff
@@ -131,6 +135,27 @@ describe('createService', () => {
         )
       })
 
+      // query-location
+      it('should support `country`', () => {
```
wizzy25: These tests should fail. Haven't tested it on my machine, but you didn't add `queryLocation` to the `features` on L98

packages/api-request-builder/test/default-params.spec.js
```diff
@@ -3,7 +3,11 @@ import { setDefaultParams } from '../src/default-params'
 
 describe('defaultParams', () => {
   it('should set default params for a normal endpoint', () => {
-    const serviceFeatures = [features.query, features.queryOne]
+    const serviceFeatures = [
```
wizzy25: Instead of modifying this test, it'll be better to write a new test case. Something like:
```js
  it('should set default params for shipping-methods', () => {
    const serviceFeatures = [
      features.query,
      features.queryOne,
      features.queryExpand,
      features.queryLocation,
    ]
    const params = {}
    setDefaultParams.call({ features: serviceFeatures, params })
    expect(params).toEqual({
      id: null,
      expand: [],
      location: {
        country: null,
        currency: null,
        state: null,
      },
      pagination: {
        page: null,
        perPage: null,
        sort: [],
      },
      query: {
        operator: 'and',
        where: [],
      },
    })
  })
```

packages/api-request-builder/test/query-location.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  it('should throw if state is missing', () => {
+    expect(() => service.byState()).toThrowError(
+      /Required argument for `byState` is missing/
+    )
+  })
+
+  it('should throw while using byCurrency if country was not set', () => {
+    expect(() => service.byCurrency('EUR')).toThrowError(
+      // eslint-disable-next-line max-len
+      /A `country` for this resource has not been set. You must set the country in order to use the `byCurrency` method./
```
wizzy25: You can truncate this string to just the first sentence to make lint happy 😉

packages/api-request-builder/test/default-params.spec.js
tdeekens: I can only recommend not to have these super big `toEqual` expectations. Rather split them up into

desc: params
it: should contain id
it: should contain expand
it: should contain location
...

And use `expect.objectContaining` to just specify what you expect.

packages/api-request-builder/src/create-service.js
junajan: you can use expand parameter while querying `shippingMethods` by location. At least this is how I understood that this code works.

docs/sdk/api/apiRequestBuilder.md
```diff
@@ -123,6 +123,11 @@ This declarative `parse` API accepts an object of the following shape:
   customerId?: ?string;
   cartId?: ?string;
 
+  // query-location
```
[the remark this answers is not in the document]
junajan: added here https://github.com/commercetools/nodejs/pull/381/commits/d86d95a6e56544a34f4251ebc8b019391699cc57#diff-feb822a4cb46cfc8410bb7dcb806c4d7R35

packages/api-request-builder/src/create-service.js
junajan: removed https://github.com/commercetools/nodejs/pull/381/commits/d86d95a6e56544a34f4251ebc8b019391699cc57#diff-feb822a4cb46cfc8410bb7dcb806c4d7R35

packages/api-request-builder/test/create-service.spec.js
junajan: strange - for me it was working .. but I added it here https://github.com/commercetools/nodejs/pull/381/commits/d86d95a6e56544a34f4251ebc8b019391699cc57#diff-fce1f76e56627dde45f04ff65082bc13R102

packages/api-request-builder/test/default-params.spec.js
junajan: added here https://github.com/commercetools/nodejs/pull/381/commits/d86d95a6e56544a34f4251ebc8b019391699cc57#diff-a63965709a1d0dead073c9d0e25e9c37R99

packages/api-request-builder/test/query-location.spec.js
junajan: yup yup yup https://github.com/commercetools/nodejs/pull/381/commits/d86d95a6e56544a34f4251ebc8b019391699cc57#diff-abcb522cc38076b7c7e7042c98e38978R49

junajan:
Hello @wizzy25 ... thank you for a good review - I was afraid of missing something because of too many files which had to be modified. So thank you and hope the PR is ok now.

packages/api-request-builder/test/default-params.spec.js
```diff
@@ -104,4 +95,21 @@ describe('defaultParams', () => {
       },
     })
   })
+
+  it('should set default params for shipping-methods', () => {
```
tdeekens: why not

```txt
describe: shipping-methods
  it: should set country in location
  it: should set currency in location
  it: should set state in location
```

packages/api-request-builder/src/create-service.js
wizzy25: @junajan Yes, and that's why in the `default-services.js` file, `features.queryExpand` is listed as one of the features. This should not be added here

packages/api-request-builder/test/create-service.spec.js
wizzy25: It's related to [this comment](https://github.com/commercetools/nodejs/pull/381#discussion_r156667277)
But nice that you added it

packages/api-request-builder/test/default-params.spec.js
junajan: Updated here d8a1be1

packages/api-request-builder/src/create-service.js
junajan: Removed here 13254e2 .. Hope now it is ok.

PR #387, opened 2017-12-19
Add custom fields on Cart Discount and Discount Codes

tdeekens:
#### Summary

This adds support for custom fields on cart discounts and discount codes.

- It reuses and improves the existing custom fields from categories
- To generalise them for both new cases


packages/sync-actions/src/category-actions.js
```diff
@@ -51,45 +50,3 @@ export function actionsMapMeta(diff, oldObj, newObj) {
     newObj,
   })
 }
-
-export function actionsMapCustom(diff, oldObj, newObj) {
```
tdeekens: This is now extracted into a module.

packages/sync-actions/src/utils/action-map-custom.js
```diff
@@ -0,0 +1,62 @@
+import * as diffpatcher from './diffpatcher'
+
+const Actions = {
+  setCustomType: 'setCustomType',
+  setCustomField: 'setCustomField',
+}
+
+const hasCustomFields = diff => Array.isArray(diff.custom)
```
tdeekens: The following little function make the code below shorter while describing what happens in each case.

packages/sync-actions/test/discount-codes-sync.spec.js
```diff
@@ -269,4 +269,68 @@ describe('Actions', () => {
     const actual = discountCodesSync.buildActions(now, before)
     expect(actual).toEqual(expected)
   })
+
+  describe('custom fields', () => {
```
tdeekens: These are "thin" integration test to the otherwise thoroughly tested module to generate the custom fields. Just so that when somebody removes them from the syncer she/he gets a more descriptive error than a e.g. a mock not being called.

packages/sync-actions/src/categories.js
```diff
@@ -3,6 +3,7 @@ import type { SyncAction, ActionGroup } from 'types/sdk'
 import flatten from 'lodash.flatten'
 import createBuildActions from './utils/create-build-actions'
 import createMapActionGroup from './utils/create-map-action-group'
+import actionsMapCustom from './utils/create-map-action-custom'
 import * as categoryActions from './category-actions'
```
gnerkus: Is this preferred over importing each object as a separate variable?

packages/sync-actions/src/utils/create-map-action-custom.js
```diff
@@ -1,40 +1,60 @@
 import * as diffpatcher from './diffpatcher'
 
-export default function actionsMapCustom(diff, oldObj, newObj) {
+const Actions = {
```
gnerkus: Nice refactor. ✅

packages/sync-actions/src/categories.js
tdeekens: It depends a bit. In this case it a. seems to be a pattern in the code base and b. often times when destructuring the named exports you lose a bit of context where they are from.

gnerkus: Thanks.

(b) is the explanation I've seen often in articles.

packages/sync-actions/src/utils/create-map-action-custom.js
```diff
@@ -1,40 +1,60 @@
...
+  Array.isArray(diffedFields[fieldName])
+    ? diffpatcher.getDeltaValue(diffedFields[fieldName])
+    : nextFields[fieldName]
+
+export default function actionsMapCustom(diff, previousObject, nextObject) {
   const actions = []
   if (!diff.custom) return actions
 
-  if (Array.isArray(diff.custom)) {
+  if (hasCustomFields(diff)) {
```
PhilippSpo: Just thinking about what these three conditions are actually trying to find out:
- `hasSingleCustomFieldChanged`
- `hasCustomTypeChanged`
- `haveMultipleCustomFieldsChanged`

packages/sync-actions/test/cart-discounts-sync.spec.js
```diff
@@ -332,4 +332,68 @@ describe('Cart Discounts Actions', () => {
     const actual = cartDiscountsSync.buildActions(now, before)
     expect(actual).toEqual(expected)
   })
+
+  describe('custom fields', () => {
+    it('should build `setCustomType` action', () => {
```
PhilippSpo: A test for changing multiple fields is missing

packages/sync-actions/src/discount-codes.js
```diff
@@ -13,6 +14,11 @@ function createDiscountCodesMapActions(mapActionGroup) {
     allActions.push(
       mapActionGroup('base', () => actionsMapBase(diff, oldObj, newObj))
     )
+
+    allActions.push(
+      mapActionGroup('custom', () => actionsMapCustom(diff, newObj, oldObj))
```
PhilippSpo: Now the two actions map functions have different signatures 😔 probably time for a bigger refactor?

tdeekens: Yeah but that mix and confusion with `old` and `new` is beyond me. It's flipped here and there and different in the docs.

packages/sync-actions/src/utils/create-map-action-custom.js
tdeekens: Very nice. This is what I was aiming for. Will change!

packages/sync-actions/test/cart-discounts-sync.spec.js
tdeekens: That test is on the custom field spec itself. This is just meant - as for categories and discount codes too - to be a rather thin integration test. I saw two options

1. Mocking 
2. Thin integration test

I felt mocking is less obvious to know what happens once you change the implementation.

tdeekens:
> good tests, but I'd like to see the actions when setting multiple fields as @PhilippSpo pointed out

I elaborated on why I did not plan to do that. It seems fragile to repeat tests in that regard.

adnasa:
aah all right. I missed that. thanks

PR #388, opened 2017-12-19
Update dependencies in bulk

tdeekens:
#### Summary

- Closes #385
- Closes #384
- Closes #382
- Closes #380
- Closes #373
- Closes #372
- Closes #371
- Closes #367
- Closes #366
- Closes #364
- Closes #356


.circleci/config.yml
```diff
@@ -1,7 +1,3 @@
-node4Environment: &node4Environment
```
tdeekens: babel-plugin-module-resolver@3 does not support node@4. Either we stay on the old version of remove resting on node@4. Note: this does not remove the support of the library for node@4.

PR #389, opened 2017-12-19
Add support custom fields for customers

tdeekens:
#### Summary

Adds custom field support for customers.


tdeekens:
@damarna85 please review. You can then quite easily improve of frontend code with this.

Issue #390, opened 2017-12-21
Switch to more powerful middleware system


dferber90:
### Description
- suggests to move `client` to a more powerful middleware system (the one of Redux)

This is not of high priority, but I want to open the issue to inform about our failed use-case for the middlewares.

#### Measuring request times
Implementing a timer is a good way to show that a middleware has control over the way up the middleware chain and the way down. While we do not want to use this, it is an easy use-case which highlights some of our requirements.

#### Retrying requests
In the MC we want to retry requests with a different token in case we receive a 401 response. Retrying (with a modified) request based on the response is a common use-case, so I wanted to build it into `client` through a middleware.
This use-case not only means that we need control on the way down, it also means that we should be able to retry the remaining middlewares (the ones invoked by `next`).

### Middleware problems

To me, the middleware system is broken as we only have control of one part of the chain, not the other. This means we can not retry requests depending on the response.

Middlewares can be used to measure the time it takes to complete a request.

```js
next => async request => {
  const start = Date.now()
  const response = await next(request)
  const duration = Date.now() - start
  return Promise.resolve({ ...response, duration })
}
```

or as in Koa

```js
app.use((ctx, next) => {
  const start = Date.now();
  return next().then(() => {
    const ms = Date.now() - start;
    console.log(`${ctx.method} ${ctx.url} - ${ms}ms`);
  });
});
```

In our case we can not do that, as we manipulate the request on the way up, but never hear about the way down. This is solved differently in Redux than it is in Express. It is also different in Koa than it is in Express which indicates that TJ (author of Express and Koa) has learned along the way.

For example https://github.com/koajs/koa#async-functions-node-v76

I think we should actually rewrite `nodejs` to the more powerfull middleware approach instead of using the existing one.

A proper middleware system calls each middleware with the passed value on the way up. Middlewares can manipulate the value on the way up. But each middleware knows about the return value as well on the way down (through the return values of `next()` calls). This gives them a chance to inspect and manipulate the return value.

A middleware can implicitly also invoke the remaining middlewares again by calling `next` multiple times. However it can only return once (unless generators are used).

This also enables usage of non-mutating middlewares.

The redux middleware system also offers a way to run the complete middleware chain again (by using `dispatch`).

### Possible Solution
Move to a middleware system like Redux or like Koa.

wizzy25:
@dferber90 Thanks for creating this issue. I really think it'll be a great way to improve the usability and robustness of the client.

[...]

PR #398, opened 2017-12-27
Redux style middlware

dferber90:
#### Summary

This is an example implementation of a redux-style middleware system suggested in #390.

This PR is meant to enable discussion on specific code, and to eventually build upon it. However I won't/can't continue with this without the company's support.

#### Next steps

We'd have to adapt all provided middlewares to this system. They will get easier to test (no mutability anymore) and more powerful.

For example, our OAuth middlewares are making HTTP requests (using `fetch`) on their own which seems weird, since the middleware / client itself should be the service used to make those requests. But this is a side-note and just a gut-feeling that something is off there. Doesn't really matter though.

#### Open questions

There is one built-in middleware in systems like this. In our case it returns the value passed through `next()`, serving as an identity middleware. This built-in middleware is the last one in the system before the return chain calls start. As described in the source code we could replace this built-in middleware for one which takes a request description and makes an http request based on it. It then returns the promise which results form it. This would declare an API contract which all other middlewares could adhere to (request description -> Promise). On the way back down they could each hook into the success- and error-cases with `return next(request).then(successHandler, failureHandler)`.

In the image below Middleware #3 would be the built-in one which turns the description into a promise: owin-middleware-chain_thumb ([source](http://johnatten.com/2015/01/04/asp-net-understanding-owin-katana-and-the-middleware-pipeline/))

*Server* would actually be the client making a request, but apart from that the image is pretty fitting.

We could alternatively keep the existing identity middleware (which just returns the passed value) and provide a `fetch` middleware instead, which users would have to manually add at the end. This would enable them to put middlewares after that. But at this point they wouldn't have access to the original request anymore, so I'm not too sure this would be useful.

#### Removed features

This PR removes the `process` function. This function should probably be implemented in a middleware instead.

#### Todo

- [ ] integrate `fetch` middleware (integrated or standalone)
- [ ] offer pagination middleware as replacement for `process`
- [ ] adapt existing middlewares

### DIFF
--- a/packages/sdk-client/src/client.js
+++ b/packages/sdk-client/src/client.js
@@ -1,25 +1,46 @@
 /* @flow */
-import qs from 'querystring'
 import type {
   Client,
   ClientOptions,
   ClientRequest,
   ClientResult,
-  MiddlewareRequest,
-  MiddlewareResponse,
-  ProcessFn,
-  ProcessOptions,
-  SuccessResult,
 } from '../../../types/sdk'
 import validate from './validate'
 
-function compose(...funcs: Array<Function>): Function {
-  // eslint-disable-next-line no-param-reassign
-  funcs = funcs.filter(func => typeof func === 'function')
+// copied from https://github.com/reactjs/redux/blob/master/src/compose.js
+function compose(...funcs: [Function]): Function {
+  if (funcs.length === 0) {
+    return <T>(arg: T): T => arg
+  }
 
-  if (funcs.length === 1) return funcs[0]
+  if (funcs.length === 1) {
+    return funcs[0]
+  }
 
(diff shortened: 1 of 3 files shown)


PhilippSpo:
Thanks for getting started here 🏆

I am definitely on board with the ideas and concepts you lay out above. If I recall correctly this would also allow us in the MC to retry requests after the access token has been refreshed 💭.

I think what needs to be discussed is how we can approach migrating the SDK and the dependent code to this new middleware concept. Will we cause breaking changes? What middlewares do we need to reimplement and how?

dferber90:
I think this was closed in error by #407 which seems unrelated

dferber90:
Seems like it's currently not worth the effort.

PR #405, opened 2018-01-16
feat(sync-actions): add custom fields to customer groups

tdeekens:
#### Summary

Adds custom fields to customer groups. This PR adds it the same way we added it for cart discounts and discount codes etc. It also contains the same restructuring of the specs.

### DIFF
--- a/packages/sync-actions/test/customer-group-sync.spec.js
+++ b/packages/sync-actions/test/customer-group-sync.spec.js
@@ -1,38 +1,136 @@
 import customerGroupSyncFn, { actionGroups } from '../src/customer-group'
 import { baseActionsList } from '../src/customer-group-actions'
 
-describe('Exports', () => {
+describe('Customer Groups Exports', () => {
   it('action group list', () => {
-    expect(actionGroups).toEqual(['base'])
+    expect(actionGroups).toEqual(['base', 'custom'])
   })
 
-  it('correctly defined based actions list', () => {
-    expect(baseActionsList).toEqual([
-      { action: 'changeName', key: 'name' },
-      { action: 'setKey', key: 'key' },
-    ])
+  describe('action list', () => {
+    it('should contain `changeName` action', () => {
+      expect(baseActionsList).toEqual(
+        expect.arrayContaining([{ action: 'changeName', key: 'name' }])
+      )
+    })
+
+    it('should contain `setKey` action', () => {
+      expect(baseActionsList).toEqual(
+        expect.arrayContaining([
+          {
+            action: 'setKey',
+            key: 'key',
+          },
+        ])
+      )
+    })
   })
 })
 
-describe('Actions', () => {
+describe('Customer Groups Actions', () => {
   let customerGroupSync
   beforeEach(() => {
     customerGroupSync = customerGroupSyncFn()
   })
 
-  it('should build `changeName` action', () => {
-    const before = { name: 'the-greatest-name-before' }
-    const now = { name: 'the-greatest-name-now' }
-    const expected = [{ action: 'changeName', ...now }]
+  it('should build the `changeName` action', () => {
+    const before = {
+      name: { en: 'en-name-before', de: 'de-name-before' },
+    }
+
+    const now = {
+      name: { en: 'en-name-now', de: 'de-name-now' },
+    }
+
+    const expected = [
+      {
+        action: 'changeName',
+        name: { en: 'en-name-now', de: 'de-name-now' },
+      },
+    ]
     const actual = customerGroupSync.buildActions(now, before)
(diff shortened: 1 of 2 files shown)


packages/sync-actions/test/customer-group-sync.spec.js
```diff
@@ -1,38 +1,136 @@
...
+    }
+
+    const now = {
+      name: { en: 'en-name-now', de: 'de-name-now' },
+    }
+
+    const expected = [
+      {
+        action: 'changeName',
+        name: { en: 'en-name-now', de: 'de-name-now' },
```
adnasa: forgive the dumb question, but help me regain sanity after reading this test block 5 times because I'm missing something here.

if `now` would be

```js
{
	name: { en: 'en-name-before', de: 'de-name-now' },
}
```

What would the action look like?

packages/sync-actions/test/customer-group-sync.spec.js
```diff
@@ -1,38 +1,136 @@
...
+            typeId: 'type',
+            id: 'customType2',
+          },
+          fields: {
+            customField1: true,
+          },
+        },
+      }
+      const actual = customerGroupSync.buildActions(now, before)
+      const expected = [{ action: 'setCustomType', ...now.custom }]
```
adnasa: whoa... This is possible on the API? to change the custom-type __entirely__?
I wonder what would happen to the values.

packages/sync-actions/test/customer-group-sync.spec.js
```diff
@@ -1,38 +1,136 @@
...
+    const actual = customerGroupSync.buildActions(now, before)
+    expect(actual).toEqual(expected)
+  })
+
+  describe('custom fields', () => {
+    it('should build `setCustomType` action', () => {
+      const before = {
+        custom: {
+          type: {
+            typeId: 'type',
```
adnasa: I just realised that this is slightly different from `reference`. Let's use a real example here?
the [list is rich](http://dev.commercetools.com/http-api-projects-custom-fields.html#customizable-resources) to pick from :)

packages/sync-actions/test/customer-group-sync.spec.js
tdeekens: Uff, I guess not much different as the setName action always contains the whole new name to be set (as a localized string).

packages/sync-actions/test/customer-group-sync.spec.js
tdeekens: This is a copy from elsewhere. Not sure if I'd rather keep it "consistent" in that sense.

packages/sync-actions/test/customer-group-sync.spec.js
tdeekens: I would assume it only serves as a blueprint for new entities from there on out not for existing ones.

packages/sync-actions/test/customer-group-sync.spec.js
wizzy25: @adnasa The `type` in the `custom` object is in fact, [a reference to a type](http://dev.commercetools.com/http-api-projects-custom-fields.html#customfields)

packages/sync-actions/test/customer-group-sync.spec.js
```diff
@@ -1,38 +1,136 @@
...
-    expect(actionGroups).toEqual(['base'])
+    expect(actionGroups).toEqual(['base', 'custom'])
   })
 
-  it('correctly defined based actions list', () => {
-    expect(baseActionsList).toEqual([
-      { action: 'changeName', key: 'name' },
-      { action: 'setKey', key: 'key' },
-    ])
+  describe('action list', () => {
```
wizzy25: You should also assert for the `setCustomType` and `setCustomField` actions

tdeekens: Nope. They're not part of the actionsBaseList.

packages/sync-actions/test/customer-group-sync.spec.js
adnasa: I did write (but discarded) that it is easy to confuse custom-type to reference :)
thanks for the info 👍

packages/sync-actions/test/customer-group-sync.spec.js
wizzy25: Ah yes. Missed that 💯

wizzy25:
@tdeekens Please remember to link to the issue so it closes on merge 🙂

tdeekens:
No, cause I don't want to create the issue in the first place 😸

matsluni:
💯

PR #409, opened 2018-01-17
Add `groups` to Discount Codes

### LINKED ISSUE #408: Add `groups` support on Discount Codes
### Description

The API supports groups (a list of strings) for Discount Codes. It's not documented yet but already available in GraphQL for instance.

image 2018-01-17 at 11 46 39 am

The only update action is `changeGroups` so we don't need to use the array removal/change detection like on tax-rates.

tdeekens:
#### Summary

Please refer to #408 for more information.

- Closes #408


packages/sync-actions/src/discount-codes.js
```diff
@@ -5,7 +5,7 @@ import actionsMapCustom from './utils/action-map-custom'
 import { actionsMapBase } from './discount-codes-actions'
 import * as diffpatcher from './utils/diffpatcher'
 
-export const actionGroups = ['base', 'references']
+export const actionGroups = ['base', 'custom']
```
tdeekens: I am pretty sure this was an error which did no cause a bug. /cc @wizzy25

tdeekens:
This is a very simple change. I will merge as we need the other export fix in the MC.

PR #418, opened 2018-01-25
332 prod json csv cli

### LINKED ISSUE #332: Implement CLI wrapper for JSON to CSV products parser
### Description

Implement CLI wrapper functionality

Part of #325

wizzy25:
#### Summary

This PR adds a CLI wrapper and integration tests for the `product-json-to-csv` package
#### Description

Part of #325 
In the continuous effort to rewrite the product exporter, this PR adds a CLI wrapper and integration tests for the `product-json-to-csv` package.
This parser can either be piped to directly from the exporter or read products from a `json` file.

The integration tests are pretty granular and may "seem" repetitive but they are necessary!

A separate PR will be created with all necessary documentation on usage. This PR is already large enough :-)

Resolves #332


.babelrc
```diff
@@ -6,7 +6,6 @@
           "env",
           {
             "targets": {
-              "browsers": ["last 1 versions"],
```
PhilippSpo: Why are we removing this? Isn't this `.babelrc` important for supporting using the sdk in a browser?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  .option('categoryBy', {
+    choices: ['name', 'key', 'externalId', 'namedPath'],
+    default: 'name',
+    describe:
+      'Define which identifier should be used for the categories column.',
+  })
+  .option('categoryOrderHintBy', {
+    choices: ['name', 'key', 'externalId', 'namedPath'],
+    default: 'name',
+    describe: 'Used CSV delimiter.',
```
PhilippSpo: Is this the right description? 🤔

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  })
+  .option('fillAllRows', {
+    describe:
+      'Define which identifier should be used for the categoryOrderHints column.',
```
PhilippSpo: Ah this one probably belongs to the option above 💡

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  if (logger.level === 'debug')
+    process.stderr.write(`ERR: ${errorFormatter.render(error)}`)
+  else process.stderr.write(`ERR: ${error.message || error}`)
+}
+
+const errorHandler = errors => {
+  if (Array.isArray(errors)) errors.forEach(logError)
+  else logError(errors)
+
+  process.exitCode = 1
```
PhilippSpo: We also sometimes needed this to be configurable in the past. Something like `continueOnErrors`. Have you thought about this?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  prettyPrint: args.prettyLogs,
+}
+const logger = pino(loggerConfig)
+
+// print errors to stderr if we use stdout for data output
+// if we save data to output file errors are already logged by pino
+const logError = error => {
+  const errorFormatter = new PrettyError()
+
+  if (logger.level === 'debug')
```
PhilippSpo: Should we read the `logLevel` from the `args` like [above](https://github.com/commercetools/nodejs/pull/418/files#diff-8ed3566bd92ec3973c8d29d39d2e0ab3R145)?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+// instantiate logger
+const loggerConfig = {
+  level: args.logLevel,
+  prettyPrint: args.prettyLogs,
+}
+const logger = pino(loggerConfig)
+
+// print errors to stderr if we use stdout for data output
+// if we save data to output file errors are already logged by pino
+const logError = error => {
```
PhilippSpo: So when we use `stdout` for data output we are logging to `stderr` using this function **and** to a file using `pino`?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  .option('logLevel', {
+    default: 'info',
+    describe: 'Logging level: error, warn, info or debug',
+    type: 'string',
+  })
+  .option('prettyLogs ', {
+    describe: 'Pretty print logs to the terminal',
+    type: 'boolean',
+  })
+  .option('logFile', {
```
PhilippSpo: This option is required when there is no `-o` given, right?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  return getCredentials(_args.projectKey)
+}
+
+// If the stdout is used for a data output, save all logs to a log file.
+// pino writes logs to stdout by default
+if (args.output === process.stdout)
+  logger.stream = fs.createWriteStream(args.logFile)
+
+// Register error listener
+args.output.on('error', errorHandler)
```
PhilippSpo: What does this line do? When is the `error` event called?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+
+let headerFields
+getHeaders(args)
```
PhilippSpo: Could we do:
```js
Promise.all([getHeaders(args), resolveCredentials(args)])
  .then(([headers, credentials]) => {
    // ...
  });
```

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  })
+  .coerce('output', arg => {
+    if (arg !== 'stdout') return fs.createWriteStream(String(arg))
+
+    return process.stdout
+  })
+  .option('categoryBy', {
+    choices: ['name', 'key', 'externalId', 'namedPath'],
+    default: 'name',
+    describe:
```
PhilippSpo: Man, it has been a while since I worked on such a tool 😅 What does this option do exactly again? 🙈

Does this depend on whether there is a category header in the template file?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+      categoryOrderHintBy: args.categoryOrderHintBy,
+      delimiter: args.delimiter,
+      fillAllRows: args.fillAllRows,
+      language: args.language,
+      multiValueDelimiter: args.multiValueDelimiter,
+      productSeparator: args.productSeparator,
+      headerFields,
+    }
+    const myLogger = {
+      error: logger.error.bind(logger),
```
PhilippSpo: Just curious: Why do we need to do this? Why can't we just use the `logger` as is?

packages/product-json-to-csv/src/map-headers.js
```diff
@@ -7,10 +7,10 @@ type HeaderObj = {
 
 export default function mapHeaders(headers: Array<string>): Array<HeaderObj> {
   return headers.map((header: string): HeaderObj => {
-    if (header === '_published' || header === 'publish')
+    if (header === '_published' || header === 'published')
       return {
         label: '_published',
-        value: 'publish',
+        value: 'published',
```
PhilippSpo: Sneaking some bugfixes in? 😉

.babelrc
tdeekens: Building, the rollup target, further down includes the last two versions. This might only affect testing.

integration-tests/cli/helpers/product-json2csv.data.js
```diff
@@ -0,0 +1,196 @@
+export const sampleProductType = {
```
tdeekens: I would recommend making these functions in `createSampleProduct` otherwise you might end up with some wonky sharing of "references" of the modules. As ES6 modules export bindings live connections to values. Hope that makes sense and I am not wrong. Tried to dig up and old article [here](http://2ality.com/2015/07/es6-module-exports.html).

integration-tests/cli/helpers/utils.js
```diff
@@ -20,7 +21,11 @@ export function clearData(apiConfig, entityName) {
     method: 'GET',
   }
   return client.process(request, payload => {
-    const results = payload.body.results
+    // Built-in states cannot be deleted
+    const results =
+      entityName === 'states'
+        ? filter(payload.body.results, { builtIn: false })
```
tdeekens: Interesting picking `lodash.filter` for it instead of saying `payload.body.results.every(result => result.builtIn)` which is part of almost all platforms now. [ref](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

packages/product-json-to-csv/package.json
```diff
@@ -23,28 +23,36 @@
...
   },
   "dependencies": {
     "@commercetools/api-request-builder": "^3.2.0",
-    "@commercetools/get-credentials": "^2.1.0",
+    "@commercetools/get-credentials": "^2.1.2",
     "@commercetools/sdk-client": "^1.5.0",
     "@commercetools/sdk-middleware-auth": "^3.4.0",
     "@commercetools/sdk-middleware-http": "^2.2.0",
     "@commercetools/sdk-middleware-user-agent": "^1.2.0",
+    "JSONStream": "^1.3.2",
```
tdeekens: Ah nice one, so you found a library.

packages/product-json-to-csv/package.json
```diff
@@ -23,28 +23,36 @@
...
     "archiver": "^2.1.0",
     "bluebird": "^3.5.0",
     "flat": "^4.0.0",
     "highland": "^2.11.1",
     "json2csv": "^3.11.5",
     "lodash": "^4.17.4",
+    "pino": "^4.10.2",
```
tdeekens: Hell, yeah <3! I love pino. Next up comes https://github.com/lrlna/pino-colada

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+  else logError(errors)
+
+  process.exitCode = 1
+}
+
+// Retrieve the headers from the template file
+// Only the first line of the file is read
+const getHeaders = _args =>
+  new Promise((resolve, reject) => {
+    if (!_args.template) resolve(null)
```
tdeekens: Curious: why resolve with `null`?

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+// pino writes logs to stdout by default
+if (args.output === process.stdout)
+  logger.stream = fs.createWriteStream(args.logFile)
+
+// Register error listener
+args.output.on('error', errorHandler)
+
+let headerFields
+getHeaders(args)
+  .then(headerFieldsfromTemplate => {
```
tdeekens: `headerFieldsFromTemplate`

packages/product-json-to-csv/src/cli.js
```diff
@@ -0,0 +1,239 @@
...
+      categoryBy: args.categoryBy,
+      categoryOrderHintBy: args.categoryOrderHintBy,
+      delimiter: args.delimiter,
+      fillAllRows: args.fillAllRows,
+      language: args.language,
+      multiValueDelimiter: args.multiValueDelimiter,
+      productSeparator: args.productSeparator,
+      headerFields,
+    }
+    const myLogger = {
```
tdeekens: Please rename. `myLogger` feels odd. `parserLogger` of whatever it belongs to might be a better option.

packages/product-json-to-csv/src/cli.js
wizzy25: @PhilippSpo The `categories` column in the exported product can either reference the `name`, `key`, `externalId`, or full `namedPath` of the category

packages/product-json-to-csv/src/cli.js
wizzy25: Oops

packages/product-json-to-csv/src/cli.js
wizzy25: No. It's set to a default `product-json-to-csv.log` if the products are piped to `stdout`

packages/product-json-to-csv/src/cli.js
wizzy25: Yes. Exactly

packages/product-json-to-csv/src/cli.js
wizzy25: I tried to avoid using the `args` directly in function definitions. I dunno which would be better

packages/product-json-to-csv/src/cli.js
wizzy25: I'm not sure if we should continue on errors for exports. Depending on the error, it may be a bad idea. We currently continue on errors for only imports

integration-tests/cli/helpers/product-json2csv.data.js
wizzy25: Makes sense. But I don't think it will bw a problem in this particular case as side effects are not a problem here (each variable is used only once)

integration-tests/cli/helpers/utils.js
wizzy25: 🙈

packages/product-json-to-csv/package.json
wizzy25: I checked it out! ❤️

packages/product-json-to-csv/src/cli.js
wizzy25: Without binding, initialising `logger` in the constructor (in `main.js`) sets the `this` to the object, which overrides `pino` and causes logging to fail.

packages/product-json-to-csv/src/map-headers.js
wizzy25: That's why we write tests 😉

packages/product-json-to-csv/src/cli.js
wizzy25: I need to resolve to a falsy value, so the exporter knows that we're doing a multifile export

packages/product-json-to-csv/src/cli.js
PhilippSpo: Ah ok got it thanks 👍  Maybe naming this `referenceCategoryBy` or something similar would clear things up 💡

packages/product-json-to-csv/src/cli.js
PhilippSpo: I'd just treat one as the single source of truth. In this case, I guess the `args` are better suited for that since they are exactly what the user inputs.

packages/product-json-to-csv/src/cli.js
PhilippSpo: Hm yeah makes sense 👍

packages/product-json-to-csv/src/cli.js
PhilippSpo: OK. There is no way to configure `pino` to behave like that in the first place?

wizzy25: I don't think so

wizzy25:
@tdeekens @PhilippSpo What do you think about it now? I made some refactorings.

...and the integration tests can be found [here](https://github.com/commercetools/nodejs/blob/244b0750d9f35fa1a0a1a8f8a3242f01982f77fe/integration-tests/cli/product-json-to-csv.it.js)

wizzy25:
I will merge this PR now, as it's getting stale. A new PR will be created with package documentation. This new PR will also be the final PR: => to `master`

PR #436, opened 2018-02-07
feat(shipping-methods-sync-actions): Add sync actions for Shipping Methods 

qmateub:

Add sync actions for shipping methods

#### Summary
Adds sync actions for shipping methods.

#### Description
Add sync actions for handling changes with Shipping Methods in the `project-settings` section of the MC.

As you may notice, I've added a lot of test for the shipping methods since the "updates" of `zones` and `shippingRates` are handled quite different by doing a `remove***` + `add***` respectively. 

Hope that make sense

Feel free to suggest anything, this took half my day because was quite "hard" (first experience with this use case) for building the actions for updating `zoneRates`, so maybe I mess up something in the middle 😅


qmateub:
Some feedback here taking into account the solution suggested by @wizzy25 in #437 

I don't know if we can use that here since the update actions are handling two different levels inside the shipping method entity

- `zoneRates` (array)
  - `zone` (obj)
  - `shippingRates` (array)
    - `price`
    - more attributes

The solution in the zones PR works for every use case of `zone`, which is a first level attribute,
 (create, remove, change/swap).

But when adding, removing or changing/swapping `shippingRates` the solution is not valid.

We have two approaches here as far as I understand:

1. Go with the current solution of a custom build actions method
2. Use `ADD_ACTIONS` and `REMOVE_ACTIONS` for zones and in the `CHANGE_ACTIONS` write a custom logic/code for taking into account the "zone-swapping" and the "create/remove/change" of `shippingRates`.

Let me know your thoughts on this 👍 

cc/ @tdeekens

tdeekens:
> Array.prototype.forEach over lodash

There is generally a lot of "uncalled-for" lodash-ing going on in this library. I agree recommending against using it is good but maybe also add an "refactoring" task to remove lodash whereever possible? We could add polyfills for most things such as `Object.{values, keys, entries}` so things don't break for consumers (especially browsers). Which brings me to the question if we have an "official" browser version support matrix for packages which can be used in the browser?

packages/sync-actions/src/shipping-methods-actions.js
```diff
@@ -0,0 +1,91 @@
...
+          zone: oldObj.zoneRates[index].zone,
+        })
+        zoneActions.push({
+          action: 'addZone',
+          zone: newObj.zoneRates[index].zone,
+        })
+      }
+      // updating shippingRates
+    } else if (zoneRate.shippingRates) {
+      forEach(zoneRate.shippingRates, (shippingRate, subIndex) => {
```
tdeekens: The nested loops feel a bit off but I fear we'll introduce more complexity when removing them than having when keeping 'em. It's due mostly to the fact that you seem to need access to the `oldObj.zoneRates[index].zone` which you would likely not have if you'd flatten/pick all `shippingRates` from all `zones` unless you add the "backward" index to each picked `shippingRates`. In that scenario you'll not reduce the iterations (as in Big O) after all I think. /cc @wizzy25

packages/sync-actions/test/shipping-methods.spec.js
```diff
@@ -0,0 +1,484 @@
...
+      expect(actual).toEqual(expected)
+    })
+  })
+
+  describe('`removeShippingRate`', () => {
+    it('should build `removeShippingRate` removing one shippingRate', () => {
+      const before = {
+        zoneRates: [
+          {
+            zone: { typeId: 'zone', id: 'z1' },
```
tdeekens: This is where your test is flawed. The zone can be expanded or is not. Your test assume that a consumer of the library always provides a non-expanded zone which is wrong I believe.

wizzy25:
I don't know about an official browser support matrix. The `.babelrc` file sets compatibility for "last 2 versions". I don't know how much that means in practical terms.

But yes, there should be a refactor of this entire library in the very near future

tdeekens:
> But yes, there should be a refactor of this entire library in the very near future

I can only recommend slitting the refactor one small bit at a time to not choke in the process.

> The .babelrc file sets compatibility for "last 2 versions"

Yeah, I don't think libraries should never ship any browser polyfills (Promise, Object.entries or anything from core.js). If they would consumers would end up with doubled polyfills and potential breakage due to outdated polyfills polluting the global scopes.

qmateub:
Alrighty guys! New approach 💅 

I've separated the logic for the zoneRates:
- On one hand, we handle the zone actions as usual since is a first level attribute.
- On the other hand, we have our `createBuildArrayActions` for shippingRates since is an array.

In the `CHANGE_ACTIONS` of the `zoneRates` we deal with the action builder for both attributes.

Let me know what you think! Most of the comments from @tdeekens are already solved with this solution 🎉

tdeekens:
What's the status here @qmateub? Can you rebase the branch so @wizzy25 and I review again.

qmateub:
> What's the status here @qmateub?

Basically this was waiting for some final reviews, but get outdated with the base branch 😓


> Can you rebase the branch so @wizzy25 and I review again.

Sure!

packages/sync-actions/src/shipping-methods-actions.js
```diff
@@ -0,0 +1,102 @@
...
+      zone: newZoneRate.zone,
+    }),
+    [REMOVE_ACTIONS]: oldZoneRate => ({
+      action: 'removeZone',
+      zone: oldZoneRate.zone,
+    }),
+    [CHANGE_ACTIONS]: (oldZoneRate, newZoneRate) => {
+      let hasZoneActions = false
+      let shippingRateActions = []
+      Object.keys(diff.zoneRates).forEach(key => {
```
tdeekens: I think we don't need mutable `shippingRateActions` here. We can just `reduce` it down?

qmateub: Something like this @tdeekens ? bc83c11 🤔

PR #437, opened 2018-02-08
fix(zone-actions): refactor actions mapping for locations

qmateub:

refactor actions mapping for locations for taking into account changes and not just add + delete

#### Summary
Refactor actions mapping for locations

#### Description

So, this is the same "problem" that in #436 when editing arrays of objects . Basically if we have this setup:

```js
// old
['Spain']

// new
['Italy']
```

The current implementation doesn't detect any changes but what we need to do is just

1. `deleteLocation` Spain
2. `addLocation` Italy

Hope that makes sense, I don't know if there's other approach for this, feel free to point it out ☝️ 

- Tests
    - [X] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


packages/sync-actions/src/zones-actions.js
```diff
@@ -19,16 +16,40 @@ export function actionsMapBase(diff, oldObj, newObj) {
...
-      action: 'removeLocation',
-      location: removedLocation,
-    }),
+  const actions = []
+  if (!diff.locations) return actions
+
+  const deleteActions = []
+  const addActions = []
+
+  forEach(diff.locations, (location, index) => {
```
tdeekens: Why did you move away from `createBuildArrayActions`?

qmateub: For what stated above in the description of the PR, I need to handle the locations changes "manually" to taking into account the location "swapping" (I don't know how to call it honestly 😅)

tdeekens: I see. Lets wait for @wizzy25 feedback on this then.

qmateub: Absolutely, I would love to know if there's a better approach for that case or not. This really gave me some headaches this last days haha

packages/sync-actions/src/zones-actions.js
```diff
@@ -19,16 +16,40 @@ export function actionsMapBase(diff, oldObj, newObj) {
...
+        addActions.push({
+          action: 'addLocation',
+          location: newObj.locations[index],
+        })
+      }
+    }
   })
 
-  return handler(diff, oldObj, newObj)
+  return addActions.concat(deleteActions)
```
wizzy25: Prefer `[...spread]` 😉

qmateub: Yeah sure! I've just based the solution from the `product-actions`, but sure we can change it to use the spread operator

qmateub:
@wizzy25 

>Also, just curious to know: what happens now if input is:

The output for your code would be this (I tried it also during development). Kind of strange eh 😅

```js
[
    { action: 'addLocation', location: { country: 'France' } },
    { action: 'addLocation', location: { country: 'Spain' } },
    { action: 'removeLocation', location: { country: 'Spain' } },
    { action: 'removeLocation', location: { country: 'France' } },
],
```

packages/sync-actions/src/zones-actions.js
wizzy25: @qmateub Instead of this, why not consider using the `createBuildArrayActions`, and modifying your code as such:
```js
// zone-actions.js
export function actionsMapLocations(diff, oldObj, newObj) {
  const handler = createBuildArrayActions('locations', {
    [ADD_ACTIONS]: newLocation => ({
      action: 'addLocation',
      location: newLocation,
    }),
    [REMOVE_ACTIONS]: oldLocation => ({
      action: 'removeLocation',
      location: oldLocation,
    }),
    [CHANGE_ACTIONS]: (oldLocation, newLocation) => [
      {
        action: 'removeLocation',
        location: oldLocation,
      },
      {
        action: 'addLocation',
        location: newLocation,
      },
    ],
  })

  return handler(diff, oldObj, newObj)
}
```
Then add one more flatten in your `zones` file
```js
// zones.js
function createZonesMapActions(mapActionGroup) {
  return function doMapActions(diff, newObj, oldObj /* , options */) {
    const allActions = []

    allActions.push(
      mapActionGroup('base', () =>
        zonesActions.actionsMapBase(diff, oldObj, newObj)
      )
    )

    allActions.push(
      flatten( // <- new flatten added here because we returned 2 objects in actions file
        mapActionGroup('locations', () =>
          zonesActions.actionsMapLocations(diff, oldObj, newObj)
        )
      )
    )

    return flatten(allActions)
  }
}
```

qmateub:
All addressed here guys!

@tdeekens @wizzy25

PR #448, opened 2018-02-12
feat(api-request-builder): add password token and password reset

### LINKED ISSUE #447: [api-request-builder]: Add support for `password-token` and `password/reset`
### Description

This `api-request-builder` maps default services to features [here](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-services.js). This is meant to be a full list of what the CTP platform supports to my understanding. But it is lacking password and token management. These are specific for customers and are documented [here](https://docs.commercetools.com/http-api-projects-customers.html#create-a-token-for-resetting-the-customers-password).

tdeekens:
#### Summary

For more please refer to #447.

Closes #447

### DIFF
--- a/packages/api-request-builder/src/default-services.js
+++ b/packages/api-request-builder/src/default-services.js
@@ -31,6 +31,16 @@ export default {
     endpoint: '/customers',
     features: [features.query, features.queryOne, features.queryExpand],
   },
+  customersPasswordToken: {
+    type: 'customers-password-token',
+    endpoint: '/customers/password-token',
+    features: [features.create],
+  },
+  customersPasswordReset: {
+    type: 'customers-password-reset',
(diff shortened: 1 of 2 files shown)


packages/api-request-builder/src/default-services.js
```diff
@@ -31,6 +31,16 @@ export default {
     endpoint: '/customers',
     features: [features.query, features.queryOne, features.queryExpand],
   },
+  customersPasswordToken: {
+    type: 'customers-password-token',
+    endpoint: '/customers/password-token',
```
wizzy25: From [here](https://docs.commercetools.com/http-api-projects-customers.html#get-customer-by-password-token), is `password-token` part of the path or the query?

PR #467, opened 2018-02-19
feat(sync-actions): add `extract-matching-pairs`

adnasa:
#### Summary

Extracting this utility out of [`product-actions`](https://github.com/commercetools/nodejs/blob/master/packages/sync-actions/src/product-actions.js#L157-L172)

This is needed as well for attributes and enums on the [product-types-actions](https://github.com/commercetools/nodejs/blob/aa-add-product-types-sync-actions/packages/sync-actions/src/product-types-actions.js#L53-L58).

Import on the product-actions will follow.

### DIFF
--- a/packages/sync-actions/src/utils/extract-matching-pairs.js
+++ b/packages/sync-actions/src/utils/extract-matching-pairs.js
@@ -1,0 +1,16 @@
+export default function extractMatchingPairs(hashMap, key, before, now) {
+  let oldObjPos
+  let newObjPos
+  let oldObj
+  let newObj
+
+  if (hashMap[key]) {
+    oldObjPos = hashMap[key][0]
+    newObjPos = hashMap[key][1]
+    if (before && before[oldObjPos]) oldObj = before[oldObjPos]
+
+    if (now && now[newObjPos]) newObj = now[newObjPos]
+  }
+
+  return { oldObj, newObj }
+}
--- a/packages/sync-actions/src/product-actions.js
+++ b/packages/sync-actions/src/product-actions.js
@@ -2,6 +2,7 @@
 import forEach from 'lodash.foreach'
 import uniqWith from 'lodash.uniqwith'
 import * as diffpatcher from './utils/diffpatcher'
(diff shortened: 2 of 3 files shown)


packages/sync-actions/src/utils/extract-matching-pairs.js
```diff
@@ -0,0 +1,16 @@
...
+
+  if (hashMap[key]) {
+    oldObjPos = hashMap[key][0]
+    newObjPos = hashMap[key][1]
+    if (before && before[oldObjPos]) oldObj = before[oldObjPos]
+
+    if (now && now[newObjPos]) newObj = now[newObjPos]
+  }
+
+  return { oldObj, newObj }
```
adnasa: There is of course a great opportunity to rename these to `previous` and `next`. I prefer that naming, tbh. Not sure tho if it is part of this PR, at least in the scope of a new utility.

packages/sync-actions/src/utils/extract-matching-pairs.js
```diff
@@ -0,0 +1,16 @@
+export default function extractMatchingPairs(hashMap, key, before, now) {
```
adnasa: shall we use named arguments such as `({ path, key, previous, next })`?
I've inspected `hashMap` and it really isn't a hashMap at all. what it is a key value object where the value is actually the path to the specified changes. the key it is indication of change or remove

packages/sync-actions/src/utils/extract-matching-pairs.js
tdeekens: Nah I think that's something for another day. Maybe even across the whole library otherwise things might get even more confusing than clear :).

adnasa:
awesome. I'm down any day to get rid of the old namings that helps nobody.

tdeekens:
Yes, I would assume so. The problem of needing matching pairs will also occur when dealing with product types.

adnasa:
@wizzy25 I just extracted out a function into a utility file so I can re-use it for the product-types sync-action

wizzy25:
Okay 👍

PR #468, opened 2018-02-20
432 groups export

wizzy25:
#### Summary

Support exporting of (discount-code)[https://dev.commercetools.com/http-api-projects-discountCodes.html#discountcode] groups in CSV

Also contains minor refactorings to make code more readable, and removes the redundant `index.js` file

resolves #432

PR #469, opened 2018-02-20
431 code groups csv parser

### LINKED ISSUE #431: Discount Code Import: add code groups
### Description
Discount Code Groups were recently added to the API. In order for customers to appropriately import discount codes including the group information, we need to provide the possibility to import groups within the discount code import. 

API documentation: https://dev.commercetools.com/http-api-projects-discountCodes.html#discountcode

FYI @ValeSauer

wizzy25:
#### Summary

Support parsing of (discount-code)[https://dev.commercetools.com/http-api-projects-discountCodes.html#discountcode] groups in CSV for importing

Also contains minor refactorings to make code more readable, removes the redundant `index.js` file, and removes `lodash` dependency (Minor step in "delodashing" this repo 😃)

resolves #431

### DIFF
--- a/packages/csv-parser-discount-code/src/main.js
+++ b/packages/csv-parser-discount-code/src/main.js
@@ -48,11 +48,10 @@ export default class CsvParserDiscountCode {
 
   // Remove fields with empty values from the code objects
   static _removeEmptyFields(item: Object) {
-    const fileredObj = {}
-    Object.keys(item).forEach(key => {
-      if (item[key] !== '') fileredObj[key] = item[key]
-    })
-    return fileredObj
+    return Object.keys(item).reduce((acc, key) => {
+      if (item[key] !== '') acc[key] = item[key]
+      return acc
+    }, {})
   }
 
   parse(input: stream$Readable, output: stream$Writable) {
@@ -91,9 +90,9 @@ export default class CsvParserDiscountCode {
       ? Object.assign(rest, {
           cartDiscounts: cartDiscounts
             .split(this.multiValueDelimiter)
-            .map(cartDiscount => ({
+            .map(cartDiscountId => ({
               typeId: 'cart-discount',
-              id: cartDiscount,
+              id: cartDiscountId,
             })),
         })
       : rest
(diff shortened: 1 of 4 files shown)


tdeekens:
So what's expected here? Should we review everything again? Have all parts been reviewed in chunks? Is there anything that was added?

wizzy25:
@tdeekens I think you're mixing this PR with #325 
Am I right?

packages/csv-parser-discount-code/src/main.js
```diff
@@ -79,16 +85,25 @@ export default class CsvParserDiscountCode {
...
-      return Object.assign(item, { cartDiscounts })
-    }
-    return item
+    const { cartDiscounts, ...rest } = item
+
+    return cartDiscounts
+      ? Object.assign(rest, {
+          cartDiscounts: cartDiscounts
+            .split(this.multiValueDelimiter)
+            .map(cartDiscount => ({
```
adnasa: cartDiscount -> cartDiscountId

packages/csv-parser-discount-code/src/main.js
```diff
@@ -40,14 +40,19 @@ export default class CsvParserDiscountCode {
...
 
     this._rowIndex = 0
   }
 
   // Remove fields with empty values from the code objects
   static _removeEmptyFields(item: Object) {
-    return _.omitBy(item, val => val === '')
+    const fileredObj = {}
+    Object.keys(item).forEach(key => {
+      if (item[key] !== '') fileredObj[key] = item[key]
```
adnasa: use `reduce` 

*cough* transducers is fully possible here *cough*

packages/csv-parser-discount-code/src/main.js
wizzy25: this wouldn't work as I'm returning the object

adnasa: the point is that you should rename this.
you do assign `cartDiscount` as id in the code below.
when I read `cartDiscount`, I think of the entity, not the id

packages/csv-parser-discount-code/src/main.js
wizzy25: Using `acc` here is the only way to keep lint happy with regards to [this comment](https://github.com/commercetools/nodejs/pull/468#discussion_r169278076)

PR #470, opened 2018-02-21
sync-actions: add product-types

adnasa:
#### Summary

- adds `sync-actions/product-types`
- please [use the documentation](https://docs.commercetools.com/http-api-projects-productTypes.html) as a reference point as you review

#### Todo

- Tests
    - [x] Unit
- [x] Documentation


packages/sync-actions/src/product-types.js
```diff
@@ -0,0 +1,53 @@
...
+function createProductTypeMapActions(
+  mapActionGroup: (
+    type: string,
+    fn: () => Array<UpdateAction>
+  ) => Array<UpdateAction>
+): (diff: Object, next: Object, previous: Object) => Array<UpdateAction> {
+  return function doMapActions(
+    diff: Object,
+    next: Object,
+    previous: Object
```
wizzy25: What're the `next` and `previous` parameters? And what's the rationale behind using them?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,205 @@
...
+      attributeName: next.name,
+      action: 'removeEnumValue',
+      value: deletedEnum,
+    }),
+  })
+  const actions = []
+  const removedKeys = []
+  const newEnumValuesOrder = []
+  flatten(handler(attributeDiff, previous, next)).forEach(action => {
+    if (action.action === 'removeEnumValue') removedKeys.push(action.value.key)
```
wizzy25: I think `action.action` looks weird. Could you maybe destructure the parameter to the `forEach` callback

adnasa: no destructs please. this is technically an `updateAction` anyway.

packages/sync-actions/src/product-types.js
adnasa: I just don't like `newObj` and `oldObj`. please don't make me go back those namings

to me `new` vs `old` sounds to me that they are objects without the same reference.
at least with `next` vs `previous`, we referr to the same entity of _potentially different versions_

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,205 @@
+import forEach from 'lodash.foreach'
```
adnasa: I wanted to go for `flow` here, but I gave up

docs/sdk/api/syncActions.md
```diff
@@ -432,3 +435,50 @@ client.execute(zonesRequests)
 .then(result => ...)
 .catch(error => ...)
 ```
+
+## `createSyncProductTypes(actionGroups)`
+
+> From package [@commercetools/sync-actions](/sdk/api/README.md#sync-actions).
+
+Creates a [sync action](/sdk/Glossary.md#sync-action) that allows to build API update actions for _zones_.
```
luhmann: > update actions for _zones_

Zones? Shouldn't it be `ProductTypes`?

docs/sdk/api/syncActions.md
```diff
@@ -432,3 +435,50 @@ client.execute(zonesRequests)
...
+
+1. `actionGroups` _(Array)_: A list of [action group](/sdk/Glossary.md#sync-action) in case some actions need to be _blacklisted_ or _whitelisted_.
+
+#### Usage example
+
+```js
+import { createSyncProductTypes } from '@commercetools/sync-actions'
+import { createClient } from '@commercetools/sdk-client'
+
+const syncZones = createSyncProductTypes()
```
luhmann: Zones again? Hmm maybe I am unaware of something. What are zones?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,205 @@
...
+            action: 'addAttributeDefinition',
+            attribute: deltaValue,
+          })
+        }
+      } else if (getIsAnAttributeDefinitionBaseFieldChange(diffValue)) {
+        actions.push(
+          ...buildBaseAttributesActions({
+            actions: attributeDefinitionsActionsList,
+            diff: diffValue,
+            oldObj: extractedPairs.oldObj,
```
luhmann: Shouldn't these be `new` and `previous` in line with what you said below?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+  diff.isSearchable
+
+function actionsMapEnums(attributeType, attributeDiff, previous, next) {
+  const getAddActionName = type =>
+    type === 'enum' ? 'addPlainEnumValue' : 'addLocalizedEnumValue'
+  const getChangeEnumOrderActionName = type =>
+    type === 'enum'
+      ? 'changePlainEnumValueOrder'
+      : 'changeLocalizedEnumValueOrder'
+  const getChangeEnumLabelActionName = type =>
```
tdeekens: I think we can hoist them up a scope next to `getIsAnAttributeDefinitionBaseFieldChange` to not create new functions on each call to `actionsMapEnums`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+      attributeName: next.name,
+      action: getAddActionName(attributeType),
+      value: newEnum,
+    }),
+    [CHANGE_ACTIONS]: (oldEnum, newEnum) => {
+      const oldEnumInNext = next.values.find(
+        nextEnum => nextEnum.key === oldEnum.key
+      )
+      const changeActions = []
+      if (oldEnumInNext) {
```
tdeekens: Does this mean something like `isPreviouslyExistingEnum`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+    [ADD_ACTIONS]: newEnum => ({
+      attributeName: next.name,
+      action: getAddActionName(attributeType),
+      value: newEnum,
+    }),
+    [CHANGE_ACTIONS]: (oldEnum, newEnum) => {
+      const oldEnumInNext = next.values.find(
+        nextEnum => nextEnum.key === oldEnum.key
+      )
+      const changeActions = []
```
tdeekens: These are the `nestedActions`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+      }
+      return changeActions
+    },
+    [REMOVE_ACTIONS]: deletedEnum => ({
+      attributeName: next.name,
+      action: 'removeEnumValue',
+      value: deletedEnum,
+    }),
+  })
+  const actions = []
```
tdeekens: These are the `flattenedActions`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+  const getChangeEnumOrderActionName = type =>
+    type === 'enum'
+      ? 'changePlainEnumValueOrder'
+      : 'changeLocalizedEnumValueOrder'
+  const getChangeEnumLabelActionName = type =>
+    type === 'enum'
+      ? 'changePlainEnumValueLabel'
+      : 'changeLocalizedEnumValueLabel'
+
+  const handler = createBuildArrayActions('values', {
```
tdeekens: `buildArrayActions` instead of `handler` as it's what we `create*`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -0,0 +1,208 @@
...
+    },
+    [REMOVE_ACTIONS]: deletedEnum => ({
+      attributeName: next.name,
+      action: 'removeEnumValue',
+      value: deletedEnum,
+    }),
+  })
+  const actions = []
+  const removedKeys = []
+  const newEnumValuesOrder = []
```
tdeekens: Why do we need to keep track of the order? As the flatting the nested array gives us an unexpected order from it's recursion?

packages/sync-actions/src/product-types-actions.js
adnasa: the convention has been `handler`, but yeah you are right

packages/sync-actions/src/product-types-actions.js
adnasa: no. This is the value passed to the updateAction of `changePlainEnumValueOrder`?
I want all values aggregated into one updateAction, instead of one

packages/sync-actions/src/product-types-actions.js
adnasa: the name of the parameter in `buildBaseAttributesActions` and what is given from `extractMatchingPairs` is `oldObj`.
it is so far only in this action file that I started introducing `previous`, and `next`. the rest we have to change in the future. that is if we like the new namings better cc @tdeekens @wizzy25

docs/sdk/api/syncActions.md
adnasa: same as above :P

docs/sdk/api/syncActions.md
adnasa: cmd + shift + d (duplicate) and missed to replace this :P

packages/sync-actions/src/product-types-actions.js
adnasa: hhmmm... nyaaaa... I see the point, but I like to isolate this into `enum` scope. moving this into the evaluation of attribute field would be mixing concerns.

packages/sync-actions/src/product-types-actions.js
adnasa: yes, but this is not a boolean value tho. I need the entity to evaluate the label, which helps determine if it is a change in the object itself or a move within the list

packages/sync-actions/src/product-types-actions.js
adnasa: they pose a nested structure of the accumulated list of actions once the handler is complete generating all our actions, but not within this utility. And calling this `nested*` doesn't reflect the actions they contain, which at the moment is a mix of remove and change..

do you still want to change the name?

packages/sync-actions/src/product-types-actions.js
tdeekens: Ah I see. Thanks for the explanation.

packages/sync-actions/src/product-types-actions.js
tdeekens: Not really. I was just trying to think of names which might lay out a bit clearer what things are and why we need to keep track of things.

adnasa: I'll document that 👍

adnasa:
I've addressed all your valuable feedback :)
hope to see this released soon, since I need it in my amazingly secret project 😉

adnasa:
@luhmann 

> It does make sense to me, but I have no real depth of understanding what exactly is happening here

If the tests make 100% sense to you, then I consider that as "approved" from your side 😉

adnasa:
okay, I'm ready for a merge here guys.
If there are no further feedback, please give me an approval or two 😉

PR #480, opened 2018-02-27
fix(http-user-agent): remove eval in favor of CSP compliant browser detection

### LINKED ISSUE #479: [http-user-agent] Avoid using "new Function" to check browser environment
### Description
Using `new Function` is [considered harmful](https://developer.chrome.com/extensions/contentSecurityPolicy#JSEval) and is blocked by CSP.

Since [we are using it](https://github.com/commercetools/nodejs/blob/master/packages/http-user-agent/src/create-user-agent.js#L5-L22) to check if the code runs on nodejs or browser, can find a better solution, thus allowing CSP to be happy about it.

tdeekens:
#### Summary

This relates to #479 and closes #479.

The browser detection was previously performed using a `new Function` which breaks harder CSP rules. Side note: jQuery also used `new Function` to parse JSON for a long time.

Detecting the browser can be done in a mirage of ways. Often using an `eval` or `new Function` statement. There are also packages like [`global-or-window`](https://github.com/purposeindustries/window-or-global/blob/master/lib/index.js) which return the Node.JS `global` or the `window`. Something we don't specifically need.

The new check is aligned with what we do at [DOMPurify](https://github.com/cure53/DOMPurify). Where we also had an issue filed around the same issue [here](https://github.com/cure53/DOMPurify/issues/249).


wizzy25:
Unit tests aren't passing

tdeekens:
> Unit tests aren't passing

Already on it.

Issue #446, opened 2018-02-12
[Auth Middleware] add documentation for extensibility


wizzy25:
### Description

Add explicit documentation on how a user may implement a custom middleware to attach a bearer token to the request header without calling the auth endpoint.

An actual middleware for this may be implemented sometime in the future; However, this issue is just for the documentation

Middleware should have this signature:
```js
const accessToken = "123"; // Got it from anywhere (not yet my concern)
 
const authHeaderWithExistingTokenMiddleware = next => (request, response) => {
  const requestWithAuth = {
    ...request,
    headers: { ...request.headers, Authorization: `Bearer: ${accessToken}` }
  };
  next(request, response);
};
```

PR #481, opened 2018-02-28
 feat(auth-middleware): add middleware for existing token

### LINKED ISSUE #464: [Middleware] Directly append access token
### Description

Write a middleware that takes in an access token and directly appends it to a request header.

This is the middleware implementation of [this issue](https://github.com/commercetools/nodejs/issues/446) and more or less an extension to it

wizzy25:
#### Summary

Add middleware for directly attaching token to  request headers
#### Description

Related to #446 this package adds the middleware to attach an already known access token to the request header
#### Talking points
It will be nice to get input on a few things before this gets merged:
 - The object properties used are `token`, `tokenType` and `force`. Does this method make sense? Are `force` and `tokenType` necessary? should they be renamed to `accessToken`, `access_token`, `accessTokenType`, `access_token_type`, etc?
 - Also, I implemented it such that it simply passes the request and response to the next middleware if there is no token passed in? I think this makes sense in certain situations such as modules that can be built to use this middleware, or an authentication flow if there is no access token. Does this make sense?

Would be nice to get your feedback/review.

resolves #464

### DIFF
--- a/docs/cli/product-exporter.md
+++ b/docs/cli/product-exporter.md
@@ -60,7 +60,7 @@ Options:
   - If the file specified already exists, it will be overwritten.
   - The default location for status report logging is the standard output.
   - If no output path is specified, the exported products will be logged to the standard output.
-- The `predicate` flag specifies an optional (where) query predicate to be included in the request. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates)
+- The `--predicate` flag specifies an optional (where) query predicate to be included in the request. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates)
 - The `--expand` flag specifies the Reference or References to expand in the returned products. The required references for expansion should be passed in as normal strings separated by a space. More information about reference expansion can be found [here](https://docs.commercetools.com/http-api.html#reference-expansion)
 - The `--exportType` flag specifies if products returned should be in JSON file format or chunks. The chunk output is particularly useful if a different output format is desired (such as CSV), in which case, the chunks can be piped to a parser to get the desired format.
 - The `--staged` flag specifies the projection of the products to be fetched.
--- a/docs/sdk/api/sdkMiddlewareAuth.md
+++ b/docs/sdk/api/sdkMiddlewareAuth.md
@@ -151,9 +151,9 @@ const client = createClient({
 })
 ```
 
-## `createAuthMiddlewareForExistingToken(options)`
+## `createAuthMiddlewareWithExistingToken(options)`
 
-Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token a request header.
+Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token `Authorization` header.
 
 #### Named arguments (options)
 
@@ -165,11 +165,11 @@ Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access
 
 ```js
 import { createClient } from '@commercetools/sdk-client'
-import { createAuthMiddlewareForExistingToken } from '@commercetools/sdk-middleware-auth'
+import { createAuthMiddlewareWithExistingToken } from '@commercetools/sdk-middleware-auth'
 
 const client = createClient({
   middlewares: [
-    createAuthMiddlewareForExistingToken({
+    createAuthMiddlewareWithExistingToken({
       token: 'my-access-token',
       tokenType: 'Bearer',
       force: true,
--- a/docs/sdk/api/sdkMiddlewareAuth.md
+++ b/docs/sdk/api/sdkMiddlewareAuth.md
@@ -151,7 +151,7 @@ const client = createClient({
 })
 ```
 
-## `createAuthMiddlewareWithExistingToken(options)`
+## `createAuthMiddlewareWithExistingToken(authorization, options)`
 
 Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token `Authorization` header.
 
--- a/docs/sdk/api/sdkMiddlewareAuth.md
+++ b/docs/sdk/api/sdkMiddlewareAuth.md
@@ -150,3 +150,30 @@ const client = createClient({
   ],
 })
 ```
+
+## `createAuthMiddlewareForExistingToken(options)`
+
+Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token a request header.
+
+#### Named arguments (options)
+
+`options` can be a string or an object. If `options` is a _(String)_, it is treated as the access token with other possible options set to their defaults. If `options` is an _(Object)_, it can have the following properties:
+
+1. `token` _(String)_: the access token to be attached to the request
+2. `tokenType` _(String)_: the type of the access token (Default: `Bearer`)
+3. `force` _(Boolean)_: if set to true, existing Authorization header (if any) in the request will be overridden with the supplied access token (Default: `true`)
+
+```js
+import { createClient } from '@commercetools/sdk-client'
+import { createAuthMiddlewareForExistingToken } from '@commercetools/sdk-middleware-auth'
+
+const client = createClient({
+  middlewares: [
+    createAuthMiddlewareForExistingToken({
+      token: 'my-access-token',
+      tokenType: 'Bearer',
+      force: true,
+    }),
+  ],
+})
+```
--- a/docs/sdk/api/sdkMiddlewareAuth.md
+++ b/docs/sdk/api/sdkMiddlewareAuth.md
@@ -155,25 +155,26 @@ const client = createClient({
 
 Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token `Authorization` header.
 
-#### Named arguments (options)
+#### Named arguments (authorization, options)
 
-`options` can be a string or an object. If `options` is a _(String)_, it is treated as the access token with other possible options set to their defaults. If `options` is an _(Object)_, it can have the following properties:
+`authorization` _(String)_: the Authorization to attach to the request header. Can be in the format `"Bearer myAccessToken"`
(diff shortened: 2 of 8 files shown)


docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -150,3 +150,30 @@ const client = createClient({
...
+2. `tokenType` _(String)_: the type of the access token (Default: `Bearer`)
+3. `force` _(Boolean)_: if set to true, existing Authorization header (if any) in the request will be overridden with the supplied access token (Default: `true`)
+
+```js
+import { createClient } from '@commercetools/sdk-client'
+import { createAuthMiddlewareForExistingToken } from '@commercetools/sdk-middleware-auth'
+
+const client = createClient({
+  middlewares: [
+    createAuthMiddlewareForExistingToken({
```
tdeekens: ~for~withExitingToken?

cneijenhuis:
I think `tokenType` is not necessary, as `Bearer` is the only value that makes sense.
I can't comment on `force` - I don't see a use case for it, but I'm not so familiar with the middleware concept of this SDK.

Otherwise 👍

docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -150,3 +150,30 @@ const client = createClient({
   ],
 })
 ```
+
+## `createAuthMiddlewareForExistingToken(options)`
+
+Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token a request header.
```
emmenko: > ...that attaches a given access token to the `Authorization` header

packages/sdk-middleware-auth/src/existing-token.js
```diff
@@ -0,0 +1,51 @@
...
+        (request.headers && request.headers.Authorization)) &&
+        config.force === false)
+    ) {
+      return next(request, response)
+    }
+    const requestWithAuth = {
+      ...request,
+      headers: {
+        ...request.headers,
+        Authorization: `${config.tokenType} ${config.token}`,
```
emmenko: What other types can be used here? 🤔

I think to be more generic we can simply accept an "authorization header value" instead of the access token, the type etc.

```js
createAuthMiddlewareWithAuthorizationHeader(`Bearer ${access_token}`)

createAuthMiddlewareWithAuthorizationHeader(`Bearer ${access_token}`, { force: false })
```

docs/cli/product-exporter.md
```diff
@@ -60,7 +60,7 @@ Options:
   - If the file specified already exists, it will be overwritten.
   - The default location for status report logging is the standard output.
   - If no output path is specified, the exported products will be logged to the standard output.
-- The `predicate` flag specifies an optional (where) query predicate to be included in the request. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates)
+- The `--predicate` flag specifies an optional (where) query predicate to be included in the request. This predicate should be wrapped in single quotes ('single quoted predicate'). More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates)
```
adnasa: good find

docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -150,3 +150,30 @@ const client = createClient({
...
+
+#### Named arguments (options)
+
+`options` can be a string or an object. If `options` is a _(String)_, it is treated as the access token with other possible options set to their defaults. If `options` is an _(Object)_, it can have the following properties:
+
+1. `token` _(String)_: the access token to be attached to the request
+2. `tokenType` _(String)_: the type of the access token (Default: `Bearer`)
```
PhilippSpo: Does the API accept anything else than a Bearer?

packages/sdk-middleware-auth/src/existing-token.js
```diff
@@ -0,0 +1,51 @@
...
+} from '../../../types/sdk'
+
+export default function createAuthMiddlewareForExistingToken(
+  options: ExistingTokenMiddlewareOptions | string
+): Middleware {
+  return (next: Next) => (
+    request: MiddlewareRequest,
+    response: MiddlewareResponse
+  ) => {
+    const config = options
```
PhilippSpo: Should we bail out immediately when there are no options given?

packages/sdk-middleware-auth/src/existing-token.js
```diff
@@ -0,0 +1,51 @@
...
+      ? {
+          tokenType:
+            typeof options === 'object' && options !== null && options.tokenType
+              ? options.tokenType
+              : 'Bearer',
+          token: typeof options === 'string' ? options : options.token,
+          force: options.force !== undefined ? options.force : true,
+        }
+      : null
```
PhilippSpo: Maybe we can make use of Lodash's `default` function here. That will also allow us to define the default options as a variable that we can refer to in order to find out what the default options are.

PhilippSpo: Maybe something like this:
```js
const defaultOptions = {
  tokenType: 'Bearer',
  force: false,
};
if (typeof options === 'string') {
  options = { ...defaultOptions, token: options }
} else if (typeof options === 'object') {
  options = default(defaultOptions, options)
}
```

packages/sdk-middleware-auth/src/existing-token.js
```diff
@@ -0,0 +1,51 @@
...
+
+    /** The request will not be modified if:
+     *  1. no argument is passed
+     *  2. an object is passed that doesn't contain a `token`
```
PhilippSpo: Why would one not pass the `token` as part of the options. Shouldn't that be considered an error?

docs/sdk/api/sdkMiddlewareAuth.md
wizzy25: No. I left it configurable in the event that there is a future supported token type. `Bearer` is still the default

packages/sdk-middleware-auth/src/existing-token.js
wizzy25: I thought about that. But it will make this middleware unusable in current CLI modules where the accessToken is optional

packages/sdk-middleware-auth/src/existing-token.js
wizzy25: Took my inspiration from [here](https://tools.ietf.org/html/draft-ietf-oauth-v2-15#section-7.1); could basically be anything. 

Only `Bearer` is supported by our API, (Besides `Basic` used for authentication), but I thought it will be a good future-proofing idea to make it configurable

packages/sdk-middleware-auth/src/existing-token.js
wizzy25: In our CLI tools, accessToken is not mandatory. Enforcing it in this middleware will make it incompatible

packages/sdk-middleware-auth/src/existing-token.js
emmenko: Making it configurable is perfectly fine, that's why I suggested to just let the user define the value itself 😉

docs/sdk/api/sdkMiddlewareAuth.md
emmenko: It would be good if we keep those middlewares as agnostic as possible in regards to our API. After all, the `sdk-client` can be used by any API.

docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -150,3 +150,31 @@ const client = createClient({
...
 })
 ```
+
+## `createAuthMiddlewareWithExistingToken(authorization, options)`
+
+Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token `Authorization` header.
+
+#### Named arguments (authorization, options)
+
+`authorization` _(String)_: the Authorization to attach to the request header. Can be in the format `"Bearer myAccessToken"`
```
emmenko: Nitpick: just to make things a bit more clear we can write something like

> the value for the `Authorization` header. For example, for you can pass the scheme `Bearer` (`'Bearer 123'`) or for `Basic` (`'Basic 123'`) and so on, depending on your authentication mechanism.

docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -151,29 +151,30 @@ const client = createClient({
...
 const client = createClient({
   middlewares: [
-    createAuthMiddlewareForExistingToken({
-      token: 'my-access-token',
-      tokenType: 'Bearer',
-      force: true,
-    }),
+    createAuthMiddlewareWithExistingToken(
+      `Bearer ${accessToken}`,
+      { force: true }
```
tdeekens: Ah so now we "only" work with bearer tokens.

emmenko: No, it's just an example. It's up to the user to define the auth scheme.

tdeekens: Ah ok. Sorry missed that.

docs/sdk/api/sdkMiddlewareAuth.md
```diff
@@ -150,3 +150,31 @@ const client = createClient({
   ],
 })
 ```
+
+## `createAuthMiddlewareWithExistingToken(authorization, options)`
+
+Creates a [middleware](/sdk/Glossary.md#middleware) that attaches a known access token `Authorization` header.
```
butenkor: known -> provided

packages/sdk-middleware-auth/src/existing-token.js
PhilippSpo: Why would I want to use this middleware if I don't want to use my existing token?

packages/sdk-middleware-auth/src/existing-token.js
PhilippSpo: Isn't it then just a matter of only using this middleware when the user of the CLI provides their token?

PR #482, opened 2018-03-01
fix(zones-locations): fix zone locations updates

qmateub:

#### Summary

This PR fixes some issues that appear during the UX/UAT that were indeed related to the sync-actions of zones.

I've added some code comment to explaining the use cases. Hope that makes sense 👍 

#### Description
Basically there were some cases that the sync-actions were not handling correctly guiding us to some UI errors (API errors) in the MC FE.

Working example. Let's say I have this object:

`locations: [{country: 'Spain'} ,{country: 'Germany'}]`

And my new object is:

`locations: [{country: 'Germany'}]`

The sync actions builded were:
```js
  [
    {action: "removeLocation", location: {country: "Spain"}},
    {action: "addLocation", location: {country: "Germany"}},
    {action: "removeLocation", location: {country: "Germany"}}
  ] 
```

That set of actions leads us to an API error because we can't add a location that is already assigned to the zone (Germany).

### DIFF
--- a/packages/sync-actions/src/zones-actions.js
+++ b/packages/sync-actions/src/zones-actions.js
@@ -25,20 +25,45 @@ export function actionsMapLocations(diff, oldObj, newObj) {
       action: 'addLocation',
       location: newLocation,
     }),
-    [REMOVE_ACTIONS]: oldLocation => ({
-      action: 'removeLocation',
-      location: oldLocation,
-    }),
-    [CHANGE_ACTIONS]: (oldLocation, newLocation) => [
-      {
-        action: 'removeLocation',
-        location: oldLocation,
-      },
-      {
-        action: 'addLocation',
-        location: newLocation,
-      },
-    ],
+    [REMOVE_ACTIONS]: oldLocation =>
+      // We only add the action if the location is not included in the new object.
+      !newObj.locations.find(
+        location => location.country === oldLocation.country
+      )
+        ? {
+            action: 'removeLocation',
+            location: oldLocation,
+          }
+        : null,
+    [CHANGE_ACTIONS]: (oldLocation, newLocation) => {
+      const result = []
+
+      // We only remove the location in case that the oldLocation is not
+      // included in the new object
+      if (
+        !newObj.locations.find(
+          location => location.country === oldLocation.country
+        )
+      )
+        result.push({
+          action: 'removeLocation',
+          location: oldLocation,
+        })
+
+      // We only add the location in case that the newLocation was not
+      // included in the old object
+      if (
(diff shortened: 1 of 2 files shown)


packages/sync-actions/src/zones-actions.js
```diff
@@ -25,20 +25,45 @@ export function actionsMapLocations(diff, oldObj, newObj) {
...
-        location: oldLocation,
-      },
-      {
-        action: 'addLocation',
-        location: newLocation,
-      },
-    ],
+    [REMOVE_ACTIONS]: oldLocation =>
+      // We only add the action if the location is not included in the new object.
+      !newObj.locations.find(
```
tdeekens: Using `.some` might be more explanatory.

packages/sync-actions/src/zones-actions.js
```diff
@@ -25,20 +25,45 @@ export function actionsMapLocations(diff, oldObj, newObj) {
...
+            location: oldLocation,
+          }
+        : null,
+    [CHANGE_ACTIONS]: (oldLocation, newLocation) => {
+      const result = []
+
+      // We only remove the location in case that the oldLocation is not
+      // included in the new object
+      if (
+        !newObj.locations.find(
```
tdeekens: Same.

packages/sync-actions/src/zones-actions.js
```diff
@@ -25,20 +25,45 @@ export function actionsMapLocations(diff, oldObj, newObj) {
...
+      )
+        result.push({
+          action: 'removeLocation',
+          location: oldLocation,
+        })
+
+      // We only add the location in case that the newLocation was not
+      // included in the old object
+      if (
+        !oldObj.locations.find(
```
tdeekens: Having the same thing three times I wonder if we can extract it into a

```js
const hasLocation(locations, otherLocation) =>
   locations.some(location => location.country === otherLocation.country
```

and just use it everywhere like

```js
if(!hasLocation(oldObj.locations, newLocation) {}
```

qmateub: Uh! Good point 👍

packages/sync-actions/src/zones-actions.js
adnasa: even go as far as to put this logic into a util

packages/sync-actions/src/zones-actions.js
adnasa: yeah, as I mentioned in my previous comment

qmateub: a180141 😎

PR #485, opened 2018-03-05
feat(sdk-middleware-http): allow to pass credentials mode to fetch object

emmenko:
#### Summary
When working with HTTP Cookies, the SDK HTTP middleware should be able to be configured to set the `credentials` mode (https://github.com/github/fetch#sending-cookies).

#### Description
Adds a new option `credentialsMode` to the `createHttpMiddleware`.

NOTE: I wasn't able to properly write a test for this, it's kind of tricky to reach into the internals of `fetch`.

### DIFF
--- a/docs/sdk/api/sdkMiddlewareHttp.md
+++ b/docs/sdk/api/sdkMiddlewareHttp.md
@@ -21,6 +21,7 @@ Creates a [middleware](/sdk/Glossary.md#middleware) to handle HTTP requests for 
 #### Named arguments (options)
 
 1. `host` *(String)*: the host of the HTTP API service
+2. `credentialsMode` *(String)*: one of the supported `credentials` modes (`omit`, `same-origin`, `include`), useful when working with HTTP Cookies. (Default: undefined)
 2. `includeResponseHeaders` *(Boolean)*: flag whether to include the response headers in the response, if omitted headers is omitted from response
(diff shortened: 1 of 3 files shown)


emmenko:
It's called credentials mode in the spec

https://fetch.spec.whatwg.org/#cors-protocol-and-credentials

tdeekens:
Yup, saw that too after some Googling all good!

docs/sdk/api/sdkMiddlewareHttp.md
```diff
@@ -21,6 +21,7 @@ Creates a [middleware](/sdk/Glossary.md#middleware) to handle HTTP requests for
 #### Named arguments (options)
 
 1. `host` *(String)*: the host of the HTTP API service
+2. `credentialsMode` *(String)*: one of the supported `credentials` modes (`omit`, `same-origin`, `include`), useful when working with HTTP Cookies. (Default: undefined)
```
wizzy25: I don't think the `(Default: undefined)` part is necessary. Makes it seem like `undefined` is a usable mode

emmenko: Yes good point, changed that to `optional` https://github.com/commercetools/nodejs/pull/485/commits/80552b4885b1c4cda35a148e34b06411311fb96a

emmenko:
@wizzy25 would you like to release it please? 🙏

PR #486, opened 2018-03-06
fix(sync-action): setDescription -> changeDescription

adnasa:
oops.

### DIFF
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -18,7 +18,7 @@ const getIsRemovedOperation = key => REGEX_UNDERSCORE_NUMBER.test(key)
 export const baseActionsList = [
   { action: 'changeName', key: 'name' },
   { action: 'setKey', key: 'key' },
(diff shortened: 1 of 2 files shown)


adnasa:
@junajan 
that's not really the problem i'm solving here.
in the [docs](https://docs.commercetools.com/http-api-projects-productTypes.html#change-description) for product-types, the action to update the description is `changeDescription`

dferber90:
Would it make sense to change this to `setDescription` in the API then (while keeping `changeDescription` around for backwards compatibility), and to not merge this PR?

yanns:
`changeXxx` is used for non-optional fields, and `setXxx` is used for optional fields.
With `setXxx` you can remove a value.

adnasa:
thanks for clarifying this!

butenkor:
Price is optional but update action is called [changePrice](https://docs.commercetools.com/http-api-projects-products.html#change-price) and not setPrice. Same with `changeAssetOrder` and `changeAssetName`. Would be easier if only `set` is used.

PR #503, opened 2018-03-20
463 support price selection

### LINKED ISSUE #463: [Sync-actions] Support Price selection
### Description

Add possibility to create `changePrice` action based on `priceSelection`. This is currently only done by comparing the prices IDs, which is not always feasible.

Similar to [this](https://github.com/sphereio/sphere-node-sdk/pull/251)

daern91:
#### Summary
Add possibility to create changePrice action based on priceSelection. This is currently done by comparing the prices IDs, so if no ID is passed in, a removePrice and addPrice would be built instead of changePrice.

#### Description
This PR will loop over the new prices array and check if they do not have any IDs passed in. If they don't, it will check priceSelection in old prices array for a match and add its ID to it. Which will successfully create the changePrice action.

#### Todo

- Tests
    - [x] Unit
    - [x] Integration
    
resolves #463


packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,57 @@
...
   return fn && typeof fn === 'function' ? fn(before, now) : [before, now]
 }
 
+function setPriceId(newPrice, oldVariantArray) {
+  let newPriceId = ''
+  const comparisonNewPrice = {
+    value: { currencyCode: newPrice.value.currencyCode },
+    channel: newPrice.channel,
+    country: newPrice.country,
+    customerGroup: newPrice.customerGroup,
```
butenkor: `validFrom` and `validUntil` is also part of price scope. There are some use-cases that do not seem to be covered here. We are working on same feature for Java currently and we use this [spec](https://wiki.commercetools.com/display/PS/Product+Variant+prices+sync) as base for implementation. Maybe we could sync together? FYI: @heshamMassoud @andrii-kovalenko-ct

andrii-kovalenko-ct: [Add Price](https://docs.commercetools.com/http-api-projects-products.html#add-price) says:

Adding a price is rejected:
  - if the product already contains a price with the same price scope (same currency, country, customer group and channel).
  - if **two prices have validity periods that overlap within the same price scope**. A price without validity period does not conflict with a price defined for a time period.

So, matching prices by only currency-country-channel-group is not enough, because each unique scope could have multiple non-overlapping validity periods.

daern91: Aha! Thanks a lot @butenkor, I didn't know this. Will add those to the scope.

wizzy25: So, just to be clear, it's safe to generate a `changePrice` action if a price has overlapping validity in the same scope (as a new price cannot be added), even if this validity cannot be updated (No `changeValidFrom`/`changeValidUntil`/`setValidFrom`/`setValidUntil` update actions exist)

daern91: @butenkor Please have another look now. I added `validFrom` and `validUntil` to the scope and some more tests to verify that it works.

andrii-kovalenko-ct: having 2 separate prices for _Jan_ and _Feb_ in **`UK-GBP`** :
```json
[
  {
    "value": {
      "currencyCode": "GBP",
      "centAmount": 123
    },
    "id": "416d1f01-e17e-41e7-abbd-dc9c6167b273",
    "country": "UK",
    "validFrom": "2018-01-01T00:00:00.000Z",
    "validUntil": "2018-01-31T00:00:00.000Z"
  },
  {
    "value": {
      "currencyCode": "GBP",
      "centAmount": 123,
    },
    "id": "f11a1159-9827-4793-b6ec-31791d2b9511",
    "country": "UK",
    "validFrom": "2018-02-01T00:00:00.000Z",
    "validUntil": "2018-02-28T00:00:00.000Z"
  }
]
```

We **can**: decrease _Jan_ price validity, for instance make it valid only _01-15 Jan_:
```json
{
   "actions": [
    {
      "action": "changePrice",
      "priceId": "416d1f01-e17e-41e7-abbd-dc9c6167b273",
      "price": {
        "value": {
          "currencyCode": "GBP",
          "centAmount": 123
        },
        "country": "UK",
        "validFrom": "2018-01-01",
        "validUntil": "2018-01-15"
      }
    }
  ]
}
```

We **can NOT**: increase _Jan_ price validity to _Feb_, like this:
```json
{
  "actions": [
    {
      "action": "changePrice",
      "priceId": "416d1f01-e17e-41e7-abbd-dc9c6167b273",
      "price": {
        "value": {
          "currencyCode": "GBP",
          "centAmount": 123
        },
        "country": "UK",
        "validFrom": "2018-01-01",
        "validUntil": "2018-02-15"
      }
    }
  ]
}
```
this causes error with message:

> The value '{\"overlappingPrices\":[{\"value\":{\"type\":\"centPrecision\",\"currencyCode\":\"GBP\",\"centAmount\":123,\"fractionDigits\":2},\"id\":\"416d1f01-e17e-41e7-abbd-dc9c6167b273\",\"country\":\"UK\",\"validFrom\":\"2018-01-01T00:00:00.000Z\",\"validUntil\":\"2018-02-15T00:00:00.000Z\"},{\"value\":{\"type\":\"centPrecision\",\"currencyCode\":\"GBP\",\"centAmount\":123,\"fractionDigits\":2},\"id\":\"f11a1159-9827-4793-b6ec-31791d2b9511\",\"country\":\"UK\",\"validFrom\":\"2018-02-01T00:00:00.000Z\",\"validUntil\":\"2018-02-28T00:00:00.000Z\"}]}' is not valid for field 'price'. Two prices have overlapping validity periods."

So, **it's not safe to generate `changePrice` action if the new `PriceDraft` contains overlapping validity date in the same scope**.

wizzy25: This module should identify the prices by `id` or by `scope` and generate actions for it. If you want to update an existing price (by `id`) to another price's scope, the action will be generated but will fail on the API. I think failing on the API is the better way as choosing not to generate the action at all requires an overly complex logic; also including the case where the id is passed in

andrii-kovalenko-ct: it is a responsibility of prices source/supplier: they should supply prices with valid scopes and validity period.
But in the same time we have explicitly specify what happens to existing prices on the target project. In our [Prices Sync](https://wiki.commercetools.com/display/PS/Product+Variant+prices+sync) documentation we explicitly specify that all the prices which don't exist in the _source_ set will be removed from _target_! This means during the sync you can't just add new prices or change them: you always obligated to supply **full prices set**.

But if you implement a service which expects to just add/update prices from some prices diff - you should somehow manage the case, when new prices in the source are valid and are not overlapping, but some existent prices on target overlap with incoming prices. 
The easiest way, of course, just to create update action and fail on its execution, but at least you have to specify this case as a known service behavior.

wizzy25: Currently, if a price is not matched, a `removePrice` action is generated. So the price with overlapping scope, if not in the product to be imported will be removed. This error should only be generated if there are two prices with overlapping scope **to be imported** from the same file.
Sorry I missed that implementation detail in my earlier post @andrii-kovalenko-ct

andrii-kovalenko-ct: ok, this sounds reasonable: 
***the error will be thrown only in case the new source set has overlapping prices*** 
e.g. if the source supplier is corrupted.

daern91:
Hi @wizzy25, @Oehmi, @butenkor and @andrii-kovalenko-ct,

I’ve taken a stab at solving this one after looking into the validity periods and its overlaps. Below are the current implementation of this PR and how it's solved.

First of all, the priceSelection will simply loop over the old prices array and see which ones have matching `currencyCode`, `channel`, `country` and `customerGroup` together with `validFrom` **AND/OR** `validUntil`. If it finds a match a changeAction will be generated.

It gets a little more complex if we do not find an exact match above and instead have validity periods which overlap with each other.

Below are a couple examples and how proposed PR will deal with them. All of this, of course, only applies if no `Id` is provided and the client instead relies on `priceSelection`.

| #   | Old                                                                                                                                        | New                                                                                                    | Action                                   | Comments                                                                                                                                                                                                 |
| --- | ------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------ | ---------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | id: 1,<br/> from: 1/1-18,<br/> until: 1/12-18                                                                                              | from:1/3-18, until:1/6-18                                                                              | changePrice                              | It will change the price period here since the new validity<br/> period is inside the old scope.                                                                                                         |
| 2   | [{ id: 1,<br/> from: 1/1-18, until:1/3-18 },<br/>{ id: 2,<br/> from:1/3-18, until:1/6-18 },<br/>{ id: 3,<br/> from:1/6-18, until:1/9-18 }] | from: 1/1-18,<br/> until: 1/12-18                                                                      | [ changePrice, removePrice, removePrice] | Changes price with id 1 to be <br/>the larger span and removes <br/>the other two overlapping prices.                                                                                                    |
| 3   | id: 1,<br/> from: 1/1-18,<br/> until: 1/12-18                                                                                              | [{ from: 1/1-18, until:1/3-18 },<br/>{ from:1/3-18, until:1/6-18 },<br/>{ from:1/6-18, until:1/9-18 }] | [ changePrice, addPrice, addPrice]       | Changes price with id 1 to <br/>be the first one until march <br/>then adds two more.                                                                                                                    |
| 4   | id: 1,<br/> from: 1/1-18,<br/> until: 1/6-18                                                                                               | from: 1/3-18,<br/> until: 1/9-18                                                                       | changePrice                              | Since the `validFrom` date of <br/>this one is inside the scope of price with id 1 it will generate<br/> a changePrice action for it. Same would apply if only the <br/>`validUntil` would be inside instead. |


Please let me know your thoughts and if it makes sense.

andrii-kovalenko-ct:
@daern91 
In our [Product Variant prices sync](https://wiki.commercetools.com/display/PS/Product+Variant+prices+sync) documentation we decided not to change by id, but remove/add a price if all 6 fields don't match (we don't have a case when price id is available).
Thus _ changePrice_ action is generated only in case both validity periods are the same on the source and target. 
Otherwise the logic becomes a bit obfuscated and unpredictable. For example:
  - in cases 2, 3: why exactly the first price should be changed, and there should be updated?
  - imagine these 2 cases:
<table>
<tr><th>#</th><th>Old</th><th>New</th><th>Note</th><th>Question</th></tr>

<tr><td><b>1</b></td><td><code>{01.05.2018;&nbsp;31.07.2018}</code></td><td><code>{01.04.2018;&nbsp;01.05.2018},</code><br/><code>{02.05.2018;&nbsp;31.07.2018}</code></td><td>First new price overlaps 1 day with the old one, the second price overlaps from second day till the end.</td><td>Which new price data should be used to update the old one?</td></tr>

<tr><td><b>2</b></td><td><code>{01.05.2018;&nbsp;31.07.2018}</code></td><td><code>{01.04.2018;&nbsp;30.07.2018},</code><br/><code>{31.07.2018;&nbsp;30.08.2018}</code></td><td>First new price overlaps almost all old period except last day, the second price overlaps from only the last day</td><td>Which new price data should be used to update the old one?</td></tr>

<tr><td><b>3</b></td><td><code>{01.05.2018;&nbsp;31.05.2018}</code>,<br/><code>{01.06.2018;&nbsp;30.06.2018}</code></td><td><code>{15.05.2018;&nbsp;15.06.2018},</code></td><td>New price starts from the middle of first old price, ends in the middle of the second old price</td><td>Which old price should be updated with actual new data?</td></tr>

</table>

 Analyzing the cases like this (and this is not a full list of possible overlaps) we decided, that price should be updated **only if all 6 fields, including dates, are match exactly**. In all other cases the prices are removed/added completely.

heshamMassoud:
@andrii-kovalenko-ct but does it really matter which price should get updated? as long as in the end you will have the new prices supplied in the target variant?

andrii-kovalenko-ct:
that's the question: too keep the service behavior consistent we have to specify explicitly when price is updated and when removed/added. Otherwise the customers could be mislead by undefined/unspecified/inconsistent behavior when some prices are get updated (keeping the same UUID) and some of them are removed. 

Of course, we could put a statement like: _some prices on the target CTP will be updated, some of them will be removed, which one - undefined_, but guess the customers would be disappointed with such service behavior: they want to know what's going on.

andrii-kovalenko-ct:
well, having a price id removes any doubts - it should be updated "as is" :)

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+) {
+  if (newValidFrom > oldValidFrom && newValidFrom < oldValidUntil) return true
+  if (newValidUntil < oldValidUntil && newValidUntil > oldValidFrom) return true
+  return false
+}
+
+function getPriceId(newPrice, oldVariantArray) {
+  let newPriceId
+
+  const newPriceComparison = {
```
tdeekens: We should likely extract this into a 

```js
const createPriceComparator = (currencyCode, channel, country, customerGroup) => ({
  value: { currencyCode },
  channel,
  country,
  customerGroup,
})
```

and hoist it out of `getPriceId` to not have the VM create new functions on each invocation.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+      if (
+        isEqual(
+          { ...newPriceComparison, validFrom: newPrice.validFrom },
+          { ...oldPriceComparison, validFrom: oldPrice.validFrom }
+        ) ||
+        isEqual(
+          { ...newPriceComparison, validUntil: newPrice.validUntil },
+          { ...oldPriceComparison, validUntil: oldPrice.validUntil }
+        ) ||
+        isDateOverlap(
```
tdeekens: Nitpick: `isDateOverlap` feels a bit to generically named. How about `hasPriceValidityOverlap`?

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+          oldPrice.validUntil,
+          newPrice.validFrom,
+          newPrice.validUntil
+        )
+      ) {
+        newPriceId = oldPrice.id
+        return true
+      }
+
+      return false
```
tdeekens: The fact that we have to return makes the code feel a bit smelly.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+  let newPriceId
+
+  const newPriceComparison = {
+    value: { currencyCode: newPrice.value.currencyCode },
+    channel: newPrice.channel,
+    country: newPrice.country,
+    customerGroup: newPrice.customerGroup,
+  }
+
+  oldVariantArray.map(oldVariant =>
```
tdeekens: Please do not add type information to variable names. It's not really a good practice. I know that the `sync-actions` use `old` and `new` as wordings. I think `next` and `previous` is actually better. This could mean these would become `previousVariants` or - in case you do decide not to rename - `oldVariants`.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+  )
+
+  return newPriceId
+}
+
+function updateMissingPriceIds(newVariantArray, oldVariantArray) {
+  // loop over and mutate newVariant price entry
+  const addedIdsArray = []
+  newVariantArray.map(newVariant => {
+    if (!newVariant.prices) return newVariant
```
tdeekens: Why do we `return` this? We never actually use it? Then we should likely return no value?

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+    return newVariant.prices.map(price => {
+      const priceWithId = price
```
tdeekens: This is a bit odd. We name something `priceWithId` before even knowing if it has an `id`?

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+      if (!priceWithId.id) {
+        const id = getPriceId(price, oldVariantArray)
+        // reference original price entry and add id to it
+        if (id && addedIdsArray.indexOf(id) < 0) {
+          priceWithId.id = id
+          addedIdsArray.push(id)
```
tdeekens: Mutation is not really a good thing. There is no real harm in returning a fresh array.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+      }
+
+      return false
+    })
+  )
+
+  return newPriceId
+}
+
+function updateMissingPriceIds(newVariantArray, oldVariantArray) {
```
tdeekens: I fail to follow. This function does not return anything neither mutates (which would be bad) anything passed in.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+        // reference original price entry and add id to it
+        if (id && addedIdsArray.indexOf(id) < 0) {
+          priceWithId.id = id
+          addedIdsArray.push(id)
+        }
+      }
+      return priceWithId
+    })
+  })
+}
```
tdeekens: Sorry, I just fail to follow what the intend of this function is I tried splitting it while realising I can't refactor it as I fail to see what it's meant to do.

Maybe it becomes easier when flattening/plucking prices first?

```js
// This plucks all prices of a given list of variants
const pluckPrices = variants => variants.reduce((prices, variant) => ([...prices, ...variant.prices]), [])
```

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -167,4 +167,551 @@ describe('Actions', () => {
     const actions = productsSync.buildActions(now, before)
     expect(actions).toEqual([])
   })
+
+  describe('build actions for prices without ID', () => {
```
tdeekens: This `describe` should likely be renamed for readability. 

```text
describe: without id
  it: should build actions for prices
```

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -167,4 +167,551 @@ describe('Actions', () => {
...
+            value: { currencyCode: 'GBP', centAmount: 10000 },
+            country: 'UK',
+            validFrom: dateNow,
+            validUntil: twoWeeksFromNow,
+          },
+        },
+      ])
+    })
+
+    it('should build changePrice actions to update dates when validFrom or validUntil priceSelection matches and there is no priceId', () => {
```
tdeekens: The "when" part is a `describe`.

So 

```txt
describe: when `validForm` or `validFrom` match
  describe: without `priceId`
     it: should build `changePrice` actions
```

This way we can also build a way nicer "tree" of expectations which reads super nice when running the tests.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+  // loop over and mutate newVariant price entry
+  const addedIdsArray = []
+  newVariantArray.map(newVariant => {
+    if (!newVariant.prices) return newVariant
+    return newVariant.prices.map(price => {
+      const priceWithId = price
+      if (!priceWithId.id) {
+        const id = getPriceId(price, oldVariantArray)
+        // reference original price entry and add id to it
+        if (id && addedIdsArray.indexOf(id) < 0) {
```
tdeekens: We could use `Array.includes` here to make it read nicer as in `!addedIdsArray.includes(id)`.

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+
+      if (
+        isEqual(
+          { ...newPriceComparison, validFrom: newPrice.validFrom },
+          { ...oldPriceComparison, validFrom: oldPrice.validFrom }
+        ) ||
+        isEqual(
+          { ...newPriceComparison, validUntil: newPrice.validUntil },
+          { ...oldPriceComparison, validUntil: oldPrice.validUntil }
+        ) ||
```
tdeekens: This should/could be extracted into a `arePricesStructruallyEqual` as this seems to be what we're doing. We don't have an `id` but want to figure out if they're actually the same structurally.

Cause then you'd have an `if` which would be

```js
if (arePricesStructruallyEqual(..., ...) || hasPriceValidatityOverlap(...))
```

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -16,6 +97,10 @@ export default function createBuildActions(differ, doMapActions, onBeforeDiff) {
       onBeforeDiff
     )
 
+    if (processedNow.variants && processedBefore.variants)
+      // run updateMissingPriceIds function to mutate processedNow data
+      updateMissingPriceIds(processedNow.variants, processedBefore.variants)
```
tdeekens: This could be then

```js
processedNow.variants = updateMissingPriceIds(processedNow.variants, processedBefore.variants)
```

where you'd maybe need an eslint ignore next-line if it barks due to mutation.

packages/sync-actions/src/utils/create-build-actions.js
tdeekens: I think we don't actually have to mutate we can just return and re-assign on the call-site?

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -167,4 +167,551 @@ describe('Actions', () => {
...
+      },
+    }
+    const now = {
+      id: '456-def',
+      masterVariant: {
+        id: 1,
+      },
+    }
+
+    it('should build changePrice actions to update centAmount through priceSelection if there is no priceId', () => {
```
tdeekens: I think the structure of the tests can be improved to e..g

```txt
describe: without `priceId`
   it: should build `changePrice` actions
```

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,88 @@
...
+  const addedIdsArray = []
+  newVariantArray.map(newVariant => {
+    if (!newVariant.prices) return newVariant
+    return newVariant.prices.map(price => {
+      const priceWithId = price
+      if (!priceWithId.id) {
+        const id = getPriceId(price, oldVariantArray)
+        // reference original price entry and add id to it
+        if (id && addedIdsArray.indexOf(id) < 0) {
+          priceWithId.id = id
```
daern91: return { 
  ...price,
  id,
}

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -167,4 +167,183 @@ describe('Actions', () => {
...
+            value: { currencyCode: 'GBP', centAmount: 1000 },
+            country: 'US',
+            validFrom: twoWeeksFromNow,
+            validUntil: threeWeeksFromNow,
+          },
+        ],
+      },
+    }
+
+    const shufflePrices = array =>
```
tdeekens: You could just use https://www.npmjs.com/package/lodash.shuffle

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,59 @@
...
+  validUntil: price.validUntil,
+})
+
+function arePricesStructurallyEqual(oldPrice, newPrice) {
+  const oldPriceComparison = createPriceComparator(oldPrice)
+  const newPriceComparison = createPriceComparator(newPrice)
+  return isEqual(newPriceComparison, oldPriceComparison)
+}
+
+function getPriceId(newPrice, previousVariants) {
```
tdeekens: Maybe rename to `extractPriceIdFromPreviousVariant`?

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -1,7 +1,59 @@
...
+        return true
+      }
+      return false
+    })
+  })
+
+  return newPriceId
+}
+
+function updateMissingPriceIds(nextVariants, previousVariants) {
```
tdeekens: Maybe rename to `injectMissingPriceIds` or `augmentWithMissingPriceIds`. I know I am strange with naming :)

packages/sync-actions/test/product-sync-prices.spec.js
daern91: Ahh, nice one. Didnt know they had this 👍

packages/sync-actions/src/utils/create-build-actions.js
daern91: Bit long but definitely makes it 100% clear whats happening. Ill bite 👍

PR #504, opened 2018-03-20
feat(state-importer): add new state importer package

### LINKED ISSUE #433: Add a State Import
### Description
We would like to offer a CSV State Import in Impex, so that the states can be updated at once.

### Context
For the customer the import/export of product states via CSV is relevant in order to be able to change the states of a lot of products at once.

wizzy25:
#### Summary

Add state importer package
#### Description

This PR adds a package which imports states to the CTP in JSON format.

It will be integrated into the Node CLI importer ([PR](https://github.com/sphereio/sphere-node-cli/pull/78))

resolves #433


integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
+import { oneLine } from 'common-tags'
+import StateImport from '@commercetools/state-importer'
+import { getCredentials } from '@commercetools/get-credentials'
+import { clearData } from './helpers/utils'
+import states from '../../packages/state-importer/test/helpers/sampleStates.json'
+
+let projectKey
+if (process.env.CI === 'true') projectKey = 'state-import-integration-test'
```
tdeekens: This feels a bit dirty I think. It there another way to configure this? Can't the test run pass the arg?

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+    })
+  )
+
+  afterAll(() => clearData(apiConfig, 'states'))
+
+  beforeEach(() => {
+    stateImport = new StateImport({ apiConfig }, logger)
+  })
+
+  test('should create states on the CTP', async () => {
```
tdeekens: Why `test` and not `it`? Is that a pattern for integration tests in the repo?

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+      detailedSummary: {
+        created: 0,
+        updated: 3,
+        unchanged: 1,
+        createErrorCount: 0,
+        updateErrorCount: 0,
+        errors: [],
+      },
+    }
+    await stateImport.run(statesToUpdate)
```
tdeekens: Just curious: is `run` the naming pattern? I've seen `execute` elsewhere.

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+    } catch (caughtError) {
+      // Because import is done in batches, we cannot guarantee how many will
+      // be successful before the failure, but the errors array must be only one
+      expect(caughtError.message).toMatch(/Processing batch failed/)
+      expect(caughtError.error.errors).toHaveLength(1)
+      expect(caughtError.error.errors[0]).toMatch(/'key' should not be empty/)
+    }
+  })
+
+  test('should continueOnProblems if `continueOnProblems`', async () => {
```
tdeekens: Just a thought: it might be nice to go the "given*, then, that" pattern which means

```
describe: with `continueOnProblems`
  it: should not stop with errors
```

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+    // Make states invalid
```
tdeekens: Another thought: `continueOnProblems` feels an odd name. How do other CLIs call this? More like a `force`?

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+        updateErrorCount: 0,
+        errors: [
+          `'key' should not be empty.`,
+          `'key' should not be empty.`,
+          `'key' should not be empty.`,
+          `'key' should not be empty.`,
+        ],
+      },
+    }
+    await stateImport.run(statesToImport)
```
tdeekens: Up until here it's setup. Should this Maybe go into a `beforeEach`?

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+    })
+
+    this.syncStates = createSyncStates()
+
+    this.logger = logger || {
+      error: () => {},
+      warn: () => {},
+      info: () => {},
+      debug: () => {},
+    }
```
tdeekens: Somewhat feels like eventually some time we want a bit smarter setup for logging. With levels and hooks kind of.

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+    return this._processBatches(states)
+  }
+
+  _processBatches(states: Array<StateData>) {
+    const predicate = StateImport._buildPredicate(states)
+    const service = this._createService()
+    const uri = service.where(predicate).build()
+    const existingStatesRequest = { uri, method: 'GET' }
+    return this.client
+      .execute(existingStatesRequest)
```
tdeekens: /cc @luhmann this is where I got the `execute` from.

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+      .then(({ body: { results: existingStates } }: Object) =>
+        this._createOrUpdate(states, existingStates)
+      )
+      .then(() => Promise.resolve())
```
tdeekens: Err, why are we doing this?

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+      const existingState = existingStates.find(
+        state => state.key === newState.key
+      )
+      if (existingState)
+        return this._update(newState, existingState)
+          .then(response => {
+            if (response && response.statusCode === 304)
+              this._summary.unchanged += 1
+            else this._summary.updated += 1
+            return Promise.resolve()
```
tdeekens: I don't think you need this.

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+          })
+          .catch(error => {
+            if (this.continueOnProblems) {
+              this._summary.updateErrorCount += 1
+              this._summary.errors.push(error.message || error)
```
tdeekens: Why do we need to keep track of both? Count and an error in itself. Feels like wonky bookkeeping. One is an derived value of the other. Either `errors.updates` to which we push for a `errors.filter(error => error.isUpdateError)` would do the same?

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+          this._summary.errors.push(error.message || error)
+          throw error
+        })
+    })
+  }
+
+  _update(newState: StateData, existingState: StateData) {
+    const actions = this.syncStates.buildActions(newState, existingState)
+    // don't call API if there's no update action
+    if (!actions.length) return Promise.resolve({ statusCode: 304 })
```
tdeekens: Nice!

packages/state-importer/src/main.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+    const req = { uri, method: 'POST', body }
+    this.logger.debug('Updating existing code entry')
+    return this.client.execute(req)
+  }
+
+  _create(state: StateData) {
+    const service = this._createService()
+    const uri = service.build()
+    const req = { uri, method: 'POST', body: state }
```
tdeekens: Can't this be curried over so we don't have to do this on each create? Feels a bit wasteful? like `createCreator` which returns a function having closed other the before?

packages/state-importer/test/main.spec.js
```diff
@@ -0,0 +1,296 @@
...
+  })
+
+  describe('::constructor', () => {
+    it('should be a function', () => {
+      expect(typeof StateImport).toBe('function')
+    })
+
+    it('should throw if no `apiConfig` in `options` parameter', () => {
+      expect(() => new StateImport({ foo: 'bar' })).toThrowError(
+        /The constructor must be passed an `apiConfig` object/
```
tdeekens: 👍

packages/state-importer/test/main.spec.js
```diff
@@ -0,0 +1,296 @@
...
+        expect.objectContaining({ body: states[0] })
+      )
+    })
+  })
+
+  describe('::_createService', () => {
+    it('should be defined', () => {
+      expect(stateImport._createService).toBeDefined()
+    })
+  })
```
tdeekens: This is pretty subjective but I would structure tests to have setup in `beforeEach` and split cases into `describe`s.

integration-tests/cli/state-import.it.js
wizzy25: Passing it as args at run time has creates 2 issues:
1. This would imply using one project for all integration tests. As the tests are run in parallel, race conditions could occur (for example: `product-json-to-csv` still running and `product-exporter` begins its teardown phase, deleting all existing products in the project). A way around this will be to run the `--runInBand` flag on CI but this will drastically slow down the int tests

2. Some of the tests are not properly structured and do not properly setup/teardown as should. So even using the `--runInBand` as described above will cause some tests to fail due to unexpected data.

It should be noted that this was the former approach, and we only changed to use a separate project for each test file due to the issues above (See #267)

docs/cli/state-importer.md
```diff
@@ -0,0 +1,168 @@
...
+  warn: console.warn,
+  info: console.log,
+  debug: console.debug,
+}
+
+const stateImport = new StateImport(options, logger)
+
+stateImport.run(states)
+  .then(() => {
+  stateImport.summaryReport()
```
adnasa: nitpick: indentation is weird here

integration-tests/cli/state-import.it.js
wizzy25: it actually makes sense. I'll change it, and create another PR for the other int tests. There seems to be different uses everywhere

integration-tests/cli/state-import.it.js
wizzy25: We use `run` for the exporters and `parse` for the parsers. The `execute` is used by the client

integration-tests/cli/state-import.it.js
wizzy25: Users said they prefer more "imperative" terms and `force` is not imperative enough. I personally like it though, but.....

integration-tests/cli/state-import.it.js
```diff
@@ -0,0 +1,131 @@
...
+
+  beforeAll(() =>
+    getCredentials(projectKey).then(credentials => {
+      apiConfig = {
+        host: 'https://auth.sphere.io',
+        apiUrl: 'https://api.sphere.io',
+        projectKey,
+        credentials,
+      }
+      // Clear all existing states
```
adnasa: oh sh*t... the above _looks dangerous_ and users/contributors of this tool can accidentally delete data from their project if configured incorrectly. Do we have a general __guide__ for developing/running this locally for example?

if not, can we add that?

integration-tests/cli/state-import.it.js
wizzy25: It could. I just didn't think it was necessary as it's used only by one test case

packages/state-importer/src/main.js
wizzy25: This is something that can be looked into. But changing this here will break compatibility with the [node-importer](https://github.com/sphereio/sphere-node-cli)

packages/state-importer/src/main.js
wizzy25: It's a hack to make the CLI work. Without this, the CLI fails even on successful import. I'm still investigating why, but so far, I've been unsuccessful

packages/state-importer/src/main.js
wizzy25: Also a hack for the CLI 👆🏾

packages/state-importer/src/main.js
wizzy25: I had something like that but I refactored it away. I will reimplement it.

packages/state-importer/test/main.spec.js
wizzy25: There's a beforeEach block:
```js
  beforeEach(() => {
    stateImport = new StateImport({ apiConfig }, logger)
  })
```
But as almost every test case requires a rather unique setup, I also do some of the setups in the test cases

wizzy25:
Thanks for the review @tdeekens 🥇💯

docs/cli/state-importer.md
wizzy25: I wonder why `prettier` didn't catch that 🤔

integration-tests/cli/state-import.it.js
wizzy25: Hmmmm. This is valid. I will add a note to the `CONTRIBUTING.md` file that setup/teardown of integration tests could cause potential loss of data

integration-tests/cli/state-import.it.js
tdeekens: What, really. Mad.

integration-tests/cli/state-import.it.js
wizzy25: Added a note here: https://github.com/commercetools/nodejs/pull/504/commits/4cdbd8b1fcdcbcaecf0919077d4ce4c84d6f2b1f

integration-tests/cli/state-import.it.js
```diff
@@ -38,94 +38,104 @@ describe('State importer', () => {
...
 
-  test('should stop import on first error by default', async () => {
-    // Make states invalid
-    const statesToImport = states.map(state => ({ ...state, key: '' }))
+  describe('Error handling', () => {
+    let invalidStates
+    beforeEach(() => {
+      invalidStates = states.map(state => ({ ...state, key: '' }))
+    })
+    describe('without `continueOnProblems', () => {
```
tdeekens: 👍

packages/state-importer/test/main.spec.js
```diff
@@ -101,6 +101,23 @@ describe('StateImport', () => {
         response.body.results
       )
     })
+
+    it('should throw Stateimport error to outer scope (CLI)', async () => {
```
tdeekens: Isn't this also like

```txt
describe: when `execute` rejects
  it: should throw `StateImport` error to CLI
```

?

packages/state-importer/test/main.spec.js
```diff
@@ -127,138 +125,166 @@ describe('StateImport', () => {
...
-      stateImport._update.mockImplementationOnce(() =>
-        Promise.reject(new Error('Second invalid state'))
-      )
-      await stateImport._createOrUpdate(states, existingStates)
-      expect(stateImport._update).toHaveBeenCalledTimes(2)
-      expect(stateImport._summary.updated).toBe(0)
-      expect(stateImport._summary.errors.update).toHaveLength(2)
-      expect(stateImport._summary.errors.update[0]).toBe('First invalid state')
-      expect(stateImport._summary.errors.update[1]).toBe('Second invalid state')
+      describe('when error occurs', () => {
```
tdeekens: when rejects?

packages/state-importer/test/main.spec.js
```diff
@@ -127,138 +125,166 @@ describe('StateImport', () => {
...
+    test('should POST a state update', async () => {
       await stateImport._update(states[2], currentState)
       expect(stateImport.client.execute).toHaveBeenCalled()
     })
-  })
 
-  describe('::_create', () => {
-    it('should be defined', () => {
-      expect(stateImport._create).toBeDefined()
+    describe('when no update actions', () => {
```
tdeekens: without actions

packages/state-importer/test/main.spec.js
```diff
@@ -268,59 +294,77 @@ describe('StateImport', () => {
...
+          ]
+          stateImport._summary.errors.update = [
+            'update error 1',
+            'update error 2',
+          ]
+        })
+
+        test('should display correct report messages', () => {
+          const report = stateImport.summaryReport()
+          expect(report.reportMessage).toMatch(oneLine`
```
tdeekens: <3

wizzy25:
@tdeekens @adnasa What do you think now? Do you think this is good enough for a merge? This is currently blocking two other PRs and I'd really love to move ahead with them... 😊

CONTRIBUTING.md
```diff
@@ -35,6 +35,8 @@ accurate comments, etc.) and any other requirements (such as test coverage).
   ```
   replace "testing-project" with your project
 
+**Note: Due to the setup/teardown nature of the integration tests, there is the risk of potential data loss in your project**
```
adnasa: 👍

PR #516, opened 2018-03-28
fix(shipping-methods): Fix issue when adding a new zoneRate with shippingRates

qmateub:

#### Summary

This PR fixes an issue when trying to add a new zoneRate to a shipping method with also shippingRates.

I've added some test for this use case. Hope that makes sense 👍 

#### Description

Basically in the `ADD_ACTION` section we were not checking if when we `addZone` , we are also adding shippingRates. Example:

```js
// before
zoneRates: []

// now
zoneRates:[ { zone: {....}, shippingRates: [{ price: {...} }, { price: {...} ]

// Expected behaviour
actions: 'addZone', 'addShippingRate', 'addShippingRate'

// Observed behaviour
actions: 'addZone'

### DIFF
--- a/packages/sync-actions/src/shipping-methods-actions.js
+++ b/packages/sync-actions/src/shipping-methods-actions.js
@@ -55,10 +55,19 @@ function actionsMapZoneRatesShippingRates(diff, oldObj, newObj) {
 
 export function actionsMapZoneRates(diff, oldObj, newObj) {
   const handler = createBuildArrayActions('zoneRates', {
-    [ADD_ACTIONS]: newZoneRate => ({
-      action: 'addZone',
-      zone: newZoneRate.zone,
-    }),
+    [ADD_ACTIONS]: newZoneRate => [
+      {
+        action: 'addZone',
+        zone: newZoneRate.zone,
+      },
+      ...(newZoneRate.shippingRates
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/sync-actions/src/shipping-methods-actions.js [hunk 1]
```diff
@@ -55,10 +55,19 @@ function actionsMapZoneRatesShippingRates(diff, oldObj, newObj) {
...
-    [ADD_ACTIONS]: newZoneRate => ({
-      action: 'addZone',
-      zone: newZoneRate.zone,
-    }),
+    [ADD_ACTIONS]: newZoneRate => [
+      {
+        action: 'addZone',
+        zone: newZoneRate.zone,
+      },
+      ...(newZoneRate.shippingRates
```
tdeekens: Can we extract this into an function defined outside? `addFixedShippingRates` or something?

qmateub: Ah sure 👍

qmateub: Done! dc7bc9a

Note: I decided to call it `addShippingRates` because we are going to add not just fixed but tiered too 👍

PR #517, opened 2018-03-28
fix(tax-categories-actions): Fix issue when doing multiple complex actions on rates

qmateub:

**Summary**
This PR fixes an issue when doing multiple complex actions with tax categories rates.

I've added some test for this use case. Hope that makes sense 👍

**Description**
We were not taking into account if we were updating the correct rate in the `replaceTaxRate`, specially when we delete and update rates. I guess is because how the diff is returned.

NOTE--> I've already slap Dani for this 😝💪

```js
// before
rates: [{ id: 1, name: 'A' }, { id: 2, name: 'B' }]

// now
rates: [{ id: 2, name: 'C' }]

// Expected behaviour
actions: [{
      {
        action: 'replaceTaxRate',
        taxRate: { id: 2, name: 'C' },
        taxRateId: 2,
      },
      { action: 'removeTaxRate', taxRateId: 1 }, // removed first tax rate
}]

// Observed behaviour
actions: [{
      {
        action: 'replaceTaxRate',
        taxRate: { id: 2, name: 'C' }, 🤷‍♂️
        taxRateId: 1, 😡
      },
      { action: 'removeTaxRate', taxRateId: 1 }, // removed first tax rate
}]

### DIFF
--- a/packages/sync-actions/src/tax-categories-actions.js
+++ b/packages/sync-actions/src/tax-categories-actions.js
@@ -30,18 +30,12 @@ export function actionsMapRates(diff, oldObj, newObj) {
       action: 'removeTaxRate',
       taxRateId: objectToRemove.id,
     }),
-    [CHANGE_ACTIONS]: (oldObject, updatedObject) =>
-      oldObject.id === updatedObject.id
-        ? {
-            action: 'replaceTaxRate',
-            taxRateId: oldObject.id,
-            taxRate: updatedObject,
-          }
-        : {
-            action: 'replaceTaxRate',
-            taxRateId: updatedObject.id,
-            taxRate: updatedObject,
-          },
+    [CHANGE_ACTIONS]: (oldObject, updatedObject) => ({
+      action: 'replaceTaxRate',
+      taxRateId:
+        oldObject.id === updatedObject.id ? oldObject.id : updatedObject.id,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/sync-actions/src/tax-categories-actions.js [hunk 1]
```diff
@@ -30,11 +30,18 @@ export function actionsMapRates(diff, oldObj, newObj) {
...
       taxRateId: objectToRemove.id,
     }),
-    [CHANGE_ACTIONS]: (oldObject, updatedObject) => ({
-      action: 'replaceTaxRate',
-      taxRateId: oldObject.id,
-      taxRate: updatedObject,
-    }),
+    [CHANGE_ACTIONS]: (oldObject, updatedObject) =>
+      oldObject.id === updatedObject.id
+        ? {
```
tdeekens: This could be simpler I think. It's quite noisy

```js
[CHANGE_ACTIONS]: (oldObject, updatedObject) => ({
  action: 'replaceTaxRate',
  taxRateId: oldObject.id === updatedObject.id ? oldObject.id : updatedObject.id,
  taxRate: updatedObject,
})
```

qmateub: Good catch --> e494503

PR #520, opened 2018-03-29
refactor(cli-int-tests): make all tests use snapshots and promises

### LINKED ISSUE #506: Refactor CLI int tests to use promisified verision of exec
### Description

To make tests easier to follow, promisified versions of some native node modules can/should be used. 
[This module](https://www.npmjs.com/package/mz) is a goo place to start

daern91:
#### Summary
To make tests easier to follow, promisified versions of some native node modules can/should be used. In addition, PR also makes tests use snapshots instead of expecting hard-coded results.

- Tests
    - [x] Unit
    - [x] Integration

resolves #506


integration-tests/cli/csv-parser-orders.it.js
```diff
@@ -14,161 +14,87 @@ describe('CSV and CLI Tests', () => {
...
-          expect(error.code).toBe(1)
-          expect(stdout).toBeFalsy()
-          expect(stderr).toMatch(
-            /Required headers missing: 'fromState,toState'/
-          )
-          done()
-        }
-      )
+      try {
+        await exec(
```
tdeekens: Nice! Just for completeness: this is one way of solving the problem. You could also just `await expect(...).rejects.toEqual(...)`.

integration-tests/cli/csv-parser-orders.it.js
```diff
@@ -14,161 +14,87 @@ describe('CSV and CLI Tests', () => {
...
-            expect(error).toBeFalsy()
-            expect(JSON.parse(data)).toEqual(expectedResult)
-            done()
-          })
-        }
+      await exec(
+        `${binPath} -i ${csvFilePath} -o ${jsonFilePath} -t returninfo`
       )
+      const data = await fs.readFile(jsonFilePath, { encoding: 'utf8' })
+      expect(JSON.parse(data)).toMatchSnapshot()
```
wizzy25: You don't need to parse it to json. You can simply match it with the string

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -22,123 +22,105 @@ describe('CSV and CLI Tests', () => {
...
-            expect(error).toBeFalsy()
-            done()
-          })
-        }
+      await exec(
+        `${binPath} -p ${projectKey} -i ${csvFilePath} -o ${jsonFilePath}`
       )
+
+      const data = await fs.readFile(jsonFilePath, { encoding: 'utf8' })
+      expect(data).toMatchSnapshot()
```
wizzy25: You can do this instead
```js
await expect(fs.readFile(jsonFilePath, { encoding: 'utf8' })).resolves.toMatchSnapshot()
```

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -22,123 +22,105 @@ describe('CSV and CLI Tests', () => {
...
+        expect(String(error)).toMatch(/Row length does not match headers/)
+      }
     })
 
-    it('on parsing errors', done => {
-      const csvFilePath = path.join(samplesFolder, 'missing-type-sample.csv')
-      const jsonFilePath = tmp.fileSync().name
+    it(
+      'on parsing errors',
+      async () => {
```
wizzy25: I think this can all be on the same line:
```js
it('on parsing errors', async () => {
```

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -186,32 +168,30 @@ describe('CSV and CLI Tests', () => {
...
-    it('CLI exits on type mapping errors', done => {
-      const csvFilePath = path.join(samplesFolder, 'wrong-type-sample.csv')
-      const jsonFilePath = tmp.fileSync().name
-
-      exec(
-        `${binPath} -p ${projectKey} -i ${csvFilePath} -o ${jsonFilePath}`,
-        (error, stdout, stderr) => {
+    it(
+      'CLI exits on type mapping errors',
+      async () => {
```
wizzy25: Same line:
```js
it('CLI exits on type mapping errors', async () => {
```

integration-tests/cli/discount-codes.it.js
```diff
@@ -308,31 +265,27 @@ describe('DiscountCode tests', () => {
...
-              expect(codeObj.createdAt).toMatch(UTCDateTimeRegex)
-              expect(codeObj.lastModifiedAt).toMatch(UTCDateTimeRegex)
-            })
-
-            done()
-          })
-        }
-      )
+        data = await fs.readFile(jsonFilePath, { encoding: 'utf8' })
+      } catch (error) {
```
wizzy25: This catch block should be removed. As we don't expect any errors, a rejected promise will fail the test.

integration-tests/cli/product-json-to-csv.it.js
```diff
@@ -481,54 +479,52 @@ describe('CSV and CLI Tests', () => {
...
-                      csvContents1 += data.toString()
-                    })
-                  }
+          try {
+            const [parseStdout, parseStderr] = await exec(
+              `${binPath} -p ${projectKey} -i ${productsJsonFile} --referenceCategoryBy namedPath --fillAllRows -o ${zipFile}`
+            )
+            expect(parseStdout).toBeTruthy()
+            expect(parseStderr).toBeFalsy()
+          } catch (error) {
```
wizzy25: same thing here ^^

integration-tests/cli/csv-parser-orders.it.js
daern91: Good find, fixed.

integration-tests/cli/csv-parser-orders.it.js
daern91: Good point, I actually looked into that way but didnt get the string matching to work with `toEqual()` since its rejects to an error Object. However, it worked fine with `toThrowError()` so will change it now 👍

integration-tests/cli/csv-parser-price.it.js
daern91: I think this gets a bit too long for one line. I also saw that prettier splits it into three lines instead.

integration-tests/cli/csv-parser-price.it.js
daern91: Prettier..

integration-tests/cli/csv-parser-price.it.js
daern91: Same here, our prettier config prefers shorter lines.

integration-tests/cli/discount-codes.it.js
daern91: Good one, fixed!

integration-tests/cli/product-json-to-csv.it.js
daern91: Awesome!

integration-tests/cli/price-export.it.js
```diff
@@ -26,170 +26,139 @@ describe('Price Exporter', () => {
...
+      actual.prices.forEach(variantPrice => {
+        expect(variantPrice['variant-sku']).toMatch(/variant-sku/)
+        expect(variantPrice.prices[0]).toMatchObject(expectedPrice)
+      })
     })
   })
 
   describe('CSV export', () => {
-    it('should resolve refs ans export only fields in template', done => {
+    it('should resolve refs ans export only fields in template', async () => {
```
wizzy25: Nitpick: `ans` -> `and`

PR #522, opened 2018-04-02
505 csv parser state

### LINKED ISSUE #505: Add State CSV parser
### Description

Add new package to parse `states` from CSV to JSON to be imported to the commercetools platform

wizzy25:
#### Summary

Add  new package to parse commmercetools state from CSV to JSON to be imported to the commercetools platform

resolves #505 

#### Todo

- Tests
    - [x] Unit
    - [x] Integration
- [x] Documentation


integration-tests/cli/csv-parser-state.it.js
```diff
@@ -0,0 +1,85 @@
...
+          })
+        })
+
+        describe('with `continueOnProblems`', () => {
+          it('should skip rows having transition', async () => {
+            const [stdout, stderr] = await exec(
+              `${binPath} --input ${statesCsv} --continueOnProblems`
+            )
+            expect(stderr).toBeFalsy()
+            expect(JSON.parse(stdout)).toHaveLength(3)
```
tdeekens: Should we be more specific of what those three items are?

packages/csv-parser-state/src/main.js
```diff
@@ -0,0 +1,190 @@
...
+
+  parse(input: stream$Readable, output: stream$Writable) {
+    this.logger.info('Starting conversion')
+    highland(input)
+      .through(csv({ separator: this.csvConfig.delimiter, strict: true }))
+      .map(CsvParserState._removeEmptyFields)
+      .map(CsvParserState._parseInitialToBoolean)
+      .map(unflatten)
+      .map(state => this._mapMultiValueFieldsToArray(state))
+      .flatMap(state => highland(this._transformTransitions(state)))
```
tdeekens: Nitpick: we dont need the wrapping fn.

packages/csv-parser-state/src/main.js
```diff
@@ -0,0 +1,190 @@
...
+        this._rowIndex += 1
+        this.logger.debug(`At row: ${this._rowIndex}, Successfully parsed`)
+      })
+      .pipe(JSONStream.stringify())
+      .pipe(output)
+  }
+
+  async _transformTransitions({
+    transitions,
+    ...restOfState
```
tdeekens: Nitpick: remainingState

integration-tests/cli/csv-parser-state.it.js
```diff
@@ -0,0 +1,85 @@
...
+            expect(stdout).toMatchSnapshot()
+          })
+        })
+      })
+
+      describe('with `projectKey`', () => {
+        it('should resolve transition and parse states', async () => {
+          const [stdout, stderr] = await exec(
+            `${binPath} -p ${projectKey} --input ${statesCsv}`
```
tdeekens: Thibgs like this could go into a beforeEach to keep tests extensible.

packages/csv-parser-state/test/main.spec.js
```diff
@@ -0,0 +1,423 @@
...
+
+      csvParser._fetchStates(uri)
+      expect(csvParser.client.execute).toBeCalled()
+      expect(csvParser.client.execute).toHaveBeenCalledWith(expectedRequest)
+    })
+
+    describe('when multiple calls with same parameter', () => {
+      it('should fetch only once', () => {
+        const uri = 'dummy-uri-2'
+        const expectedRequest = { uri, method: 'GET' }
```
tdeekens: All of that jazz into a beforeEach?

packages/csv-parser-state/src/main.js
```diff
@@ -0,0 +1,190 @@
...
+    return { ...restOfState, transitions: stateReferences }
+  }
+
+  _buildStateRequest(stateKey: string): Promise<SuccessResult> {
+    const stateService = this._createStateService()
+    const uri = stateService.byKey(stateKey).build()
+    return this._fetchStates(uri)
+  }
+
+  _setupClient() {
```
tdeekens: Nitpick: createCliebt?

packages/csv-parser-state/src/main.js
```diff
@@ -0,0 +1,190 @@
...
+      this.logger.warn(`Ignoring error at row: ${this._rowIndex}, ${error}`)
+    else
+      // <- Rethrow the error to `.stopOnError()`
+      callback(error)
+  }
+
+  // Remove fields with empty values from the state objects
+  static _removeEmptyFields(item: Object): Object {
+    // "acc" is "state" but lint doesn't like mutations
+    return Object.entries(item).reduce((acc, entry) => {
```
tdeekens: You could destructure the item inline herr already to not have to refer to it by deferring it below.

packages/csv-parser-state/src/main.js
adnasa: I'm wondering why this is part of the instance in the first place.
The only time the `client` itself is used is with `execute` below.

docs/cli/csv-parser-state.md
```diff
@@ -0,0 +1,163 @@
...
+                             encounters an error.     [boolean] [default: false]
+  --logLevel                 Logging level: error, warn, info or debug
+                                                      [string] [default: "info"]
+  --prettyLogs               Pretty print logs to the terminal         [boolean]
+  --logFile                  Path to file where logs should be saved
+                                   [string] [default: "csv-parser-state.log"]
+```
+
+#### Info on flags
+- The `--projectKey` is only required if states have a `transitions` field containing one or more states that need to be resolved, otherwise it can be omitted. More info in the **Examples** below
```
adnasa: is there a default if `projectKey` isn't provided?

docs/cli/csv-parser-state.md
```diff
@@ -0,0 +1,163 @@
...
+- The `--input` flag specifies the path to the states CSV file. If this flag is omitted, the module will attempt to read the data from the `standard input`.
+- The `--output` flag specifies where to output/save the parsed states as JSON file. Several notes on this flag:
+  - If the file specified already exists, it will be overwritten.
+  - The default location for status report logging is the standard output.
+  - If no output path is specified, the parsed states will be written to the standard output as a result, status reports will be logged to a `csv-parser-state.log` file in the current directory.
+- The `--delimiter` flag specifies the delimiter used in the input file. Defaults to `','` if omitted.
+- The `--multiValueDelimiter` flag specifies the delimiter for multiValue cells in CSV. Note that only the `transitions` and `roles` fields if present should contain multiple values. Defaults to `';'` if omitted.
+- The `--continueOnProblems` flag specifies if the module should continue parsing states if it encounters an error. Defaults to `false` if omitted.
```
adnasa: Is there another word to `continueOnProblems`.

docs/cli/csv-parser-state.md
```diff
@@ -0,0 +1,163 @@
...
+                                   [string] [default: "csv-parser-state.log"]
+```
+
+#### Info on flags
+- The `--projectKey` is only required if states have a `transitions` field containing one or more states that need to be resolved, otherwise it can be omitted. More info in the **Examples** below
+- The `--input` flag specifies the path to the states CSV file. If this flag is omitted, the module will attempt to read the data from the `standard input`.
+- The `--output` flag specifies where to output/save the parsed states as JSON file. Several notes on this flag:
+  - If the file specified already exists, it will be overwritten.
+  - The default location for status report logging is the standard output.
+  - If no output path is specified, the parsed states will be written to the standard output as a result, status reports will be logged to a `csv-parser-state.log` file in the current directory.
```
adnasa: Isn't this saying the same thing as above?
Can we remove the previous bullet point, if you agree?

packages/csv-parser-state/src/main.js
```diff
@@ -0,0 +1,190 @@
...
+        createUserAgentMiddleware({
+          libraryName: pkg.name,
+          libraryVersion: pkg.version,
+        }),
+        createHttpMiddleware({ host: this.apiConfig.apiUrl }),
+      ],
+    })
+  }
+
+  _createStateService() {
```
adnasa: I think this can be moved out of the class.

```js
const createCow = () => {}

class CowFarm {
  super() {
    this.cow = createCow()
  }
}
```

docs/cli/csv-parser-state.md
tdeekens: We had that discussion on previous PRs. By now it's established lingua. People liked it more than `--force` which would have been my preference.

adnasa: I would have preferred `--force` as well :)

integration-tests/cli/csv-parser-state.it.js
wizzy25: I thought snapshot testing makes this not essentially necessary. I could be wrong though

packages/csv-parser-state/src/main.js
wizzy25: We do. The `.flatmap()` expects a stream to be passed to the callback and because `. _transformTransitions()` returns a promise, the `highland()` wrapper transforms the contents of the resolved (or rejected promise to stream chunks

docs/cli/csv-parser-state.md
wizzy25: No there isn't. It defaults to undefined

docs/cli/csv-parser-state.md
wizzy25: Me too. But for "usability" reasons, users of our tools prefer this

packages/csv-parser-state/src/main.js
wizzy25: Or maybe a static method would be better

wizzy25:
Thanks for the review guys @adnasa @tdeekens 
I refactored the code to use purer static methods. Please have a look. Thanks

adnasa:
👍

PR #528, opened 2018-04-10
fix(HTTP): mask both 'Auth' and 'auth' data in all errors

daern91:
#### Summary
Follow up PR to #524 as we found that both `authorization` and `Authorization` is used in the headers so we need to mask both. Also refactored to use `maskAuthData` function for all errors.

### DIFF
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -40,6 +40,13 @@ function calcDelayDuration(
   return retryDelay
 }
 
+function maskAuthData(request) {
+  if (request.headers.authorization)
+    request.headers.authorization = 'Bearer ********'
+  if (request.headers.Authorization)
+    request.headers.Authorization = 'Bearer ********'
+}
+
 export default function createHttpMiddleware({
   host,
   credentialsMode,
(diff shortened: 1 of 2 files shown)


packages/sdk-middleware-http/src/http.js
```diff
@@ -91,9 +98,9 @@ export default function createHttpMiddleware({
                   ...requestObj,
                   headers: parseHeaders(requestObj.headers),
                 }
-                if (maskSensitiveHeaderData)
-                  parsedResponse.request.headers.authorization =
-                    'Bearer ********'
+                if (maskSensitiveHeaderData) {
```
junajan: Hi @daern91 .. what do you think about moving this condition to `maskAuthData` function? So we will test if it is enabled only on one place.

packages/sdk-middleware-http/src/http.js
```diff
@@ -40,6 +40,13 @@ function calcDelayDuration(
   return retryDelay
 }
 
+function maskAuthData(request: Object) {
+  if (request.headers.authorization)
```
junajan: Do we still have both cases `authorization` and `Authorization`?

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -389,9 +389,9 @@ describe('Http', () => {
       })
       const response = { resolve, reject }
       const next = (req, res) => {
-        const expectedError = new Error('non json error occured')
+        const expectedError = new Error('non json error occurred')
```
junajan: 👍

packages/sdk-middleware-http/src/http.js
daern91: Great idea, thanks, done!

packages/sdk-middleware-http/src/http.js
daern91: Yup. I checked with the team and they are both used everywhere in the code so it makes sense to check/fix both of them.

PR #541, opened 2018-04-17
feat(sync-actions): omit action when field is optional and value is undefined, null or empty string

adnasa:
#### Summary

- omits the action `key` when the "new" value is one of `undefined`, `null`, or empty string
- __does not__ generate an update action if the previous value is one of `undefined`, `null` or an empty string

added @qmateub, @butenkor  and @tdeekens for information since this may affect their applications.

#### Description

So, we discovered in our application that the `sync-actions` unnecessarily generates `update-actions` when the next value is actually not a **value** (empty string)

At the moment, when a value moves from `undefined -> null` and vice versa, the sync-actions doesn't generate anything since we are relying on `lodash.isNil` [ref](https://lodash.com/docs/4.17.5#isNil) which takes care of this for us.

However when we go from `{undefined|null} -> ''`, this is where the sync-actions unecessarily generates update-actions.

#### Possible implications

Depending on the implementation on the client side and general use case, this can be a breaking change. I'd like to hear feedback from you guys if this is the case. At the very least for the product I work on, this would ease a lot of pain point of us having to apply work-arounds to "trick" the sync-actions to not generate any.

The worst case scenario for a breaking change is if the project work on, actually __relies__ saving fields with an empty string. 🤷🏽‍♂️

#### Strategy

- Get your feedback
- If we move on with this, I will update the test across all actions in `sync-actions` to cover this.

#### TODO 

- Tests
    - [ ] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


packages/sync-actions/src/product-types-actions.js
```diff
@@ -15,16 +15,19 @@ const REGEX_UNDERSCORE_NUMBER = new RegExp(/^_\d+$/)
...
   { action: 'changeDescription', key: 'description' },
+  // optional
+  { action: 'setKey', key: 'key' },
 ]
 
 export function actionsMapBase(diff, previous, next) {
   return buildBaseAttributesActions({
     diff,
-    actions: baseActionsList,
+    actions: updateActionsDefinitions,
```
adnasa: this was a renaming to help myself to make sense of `buildBaseAttributesActions` 😄 
I'll keep it if you don't mind

tdeekens:
This should only affect product types, right? I think custom fields in general use empty strings to unset things.

dferber90:
Here's a table of possible states and the generated update actions

upload


> This should only affect product types, right? I think custom fields in general use empty strings to unset things.

IMHO the `sync-actions` should either do this for all scopes, or for none. Doing it only for certain sections could be quite surprising for users of the package.


---

The PR subscription further doesn't mention that there are different cases which can happen:
- there are string fields in the API which are optional
- there are string fields in the API which are required

A required field can never be `null` or `undefined`, so we'd have to always keep the empty string.
Not sure what should happen when the new value is `null` or `undefined`.

For optional fields this is probably different.

Did you look into this @adnasa ?

tdeekens:
> IMHO the sync-actions should either do this for all scopes

Agreed. I was only unclear in meaning:

1. Currently - as this PR stands right now - this would "only" affect product types
2. Some update actions with empty strings may be used for unsetting (instead of null). We should check if that's true or false to ensure that we don't break things

That table is ace!

tdeekens:
Sorry @adnasa my bad. Skimmed through the files and only saw product type related things being touched. Makes sense that it affects all with common-actions. 

Regarding my suspicions with custom fields this is the API docs https://docs.commercetools.com/http-api-projects-customers sadly nothig about what the undefined value is. maybe its null and were all cool.

dferber90:
> I don't know where this is coming from and I'd like to hear your expectation of the outcome of this PR. 

I didn't think about this too much, but I have a feeling it might cause problems.

When the user supplies `null` or `undefined` as the next value of an optional text field, then we'll keep an empty string if that was the previous value, otherwise we'll use `null`/ `undefined`?

When the user supplies `null` or `undefined` as the next value of a required text field, then that's basically invalid. What do we do? Keep what was there before or set it to an empty string?

We should be able to clearly state the different cases and which sync-actions will be generated. Sort of like with the table I supplied above. That table is the current state though. I'd expect tables like this:

**Required Text Field**

Old value | New Value | Final Value
-- | -- | --
`""` | `null` | `""`
`""` | `undefined` | `""`
`""` | `"foo"` | `"foo"`
`"foo"` | `""` | `""`

**Optional Text Field**

Old value | New Value | Final Value
-- | -- | --
`""` | `null` | `""`
`""` | `undefined` | `""`
`""` | `"foo"` | `"foo"`
`"foo"` | `""` | `""`
`undefined` | `null` | `undefined`
`undefined` | `""` | `undefined`
`null` | `undefined` | `null`

adnasa:
A documentation has been updated with a table where it is discussed all the possible outcomes of the `sync-actions` can give you depending on your `old vs now` values.

8a6ec45

If we are all okay with this, then I add the tests on each action, starting today.

butenkor:
Is it ensured that with above use cases one can always unset a value? Also does value `""` always means unset for the API? According to documentation in order to unset an attribute of a variant, value should not be supplied or be `null`. It does not mention behaviour on values with `""`

https://docs.commercetools.com/http-api-projects-products.html#set-attribute

adnasa:
> Also does value "" always means unset for the API?

For the API, to my understanding; `""` does not mean unsetting the value.

It is to actually give the value of `""`. This is also why this change can cause breaking change to client side applications with insufficient workaround;

__side note__
e.g they would use `attribute.value === ''` to determine that the attribute has no value. The only use case on the FE shops I can think of needing an empty string as a value, is if they were to need to display the attribute (label) but empty value. but IMO they would solve that properly by screening the product-type.

Hope that clarifies things for you

wizzy25:
@adnasa Is this valid for only updates on text fields? And if so, in what scenario is the old value (returned from the API) `null` or `undefined`?
Maybe I'm wrong, but from my little digging, if a value is unset, the API will not return `undefined`

adnasa:
The diff so far covers "base field" for the product type.
And it will continue to do so (I'll add tests later once the approach for this is approved)

>  if a value is unset, the API will not return undefined

the API doesn't return "undefined" in the response. the key of the value simply doesn't exist.

you can try by updating a product (that has a key) with the following:

```json
{
  "actions": [{ "action": "setKey" }]
}
```

docs/sdk/api/syncActions.md
```diff
@@ -556,3 +558,37 @@ client.execute(statesRequests)
...
+
+---
+
+### Note
+
+When it comes to value evaluation for fields of possibly any type, is that we won't generate any `update-actions` for you in these following cases.
+
+| Before      | Now         | Will generate update action? | Value for the action    |
+| ----------- | ----------- | ---------------------------- | ----------------------- |
+| `""`        | `null`      | no                           |                         |
```
adnasa: After discussing with @butenkor, it looks like this is supposed to generate an update action,
where we omit where the value is omitted from the action. good catch 🎣

docs/sdk/api/syncActions.md
```diff
@@ -556,3 +558,37 @@ client.execute(statesRequests)
 .then(result => ...)
 .catch(error => ...)
 ```
+
+---
+
+### Note
```
emmenko: This could be "troubleshooting"?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -15,16 +15,19 @@ const REGEX_UNDERSCORE_NUMBER = new RegExp(/^_\d+$/)
...
+    shouldOmitOptional: true,
```
emmenko: What's this?

packages/sync-actions/test/product-types-sync-base.spec.js
```diff
@@ -60,22 +60,178 @@ describe('Actions', () => {
...
+        beforeEach(() => {
+          before = {
+            key: 'sneakers-key',
+          }
+          now = {
+            key: 'kicks-key',
+          }
+          updateActions = productTypesSync.buildActions(now, before)
+        })
+        test('should return `setKey` update-action', () => {
```
emmenko: Nitpick: `should return action with key value`

packages/sync-actions/test/product-types-sync-base.spec.js
```diff
@@ -60,22 +60,178 @@ describe('Actions', () => {
...
+        beforeEach(() => {
+          before = {
+            key: 'sneakers-key',
+          }
+          now = {
+            key: '',
+          }
+          updateActions = productTypesSync.buildActions(now, before)
+        })
+        it('should return `setKey` update-action', () => {
```
emmenko: Nitpick: `should unset action value`

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -20,8 +20,8 @@ export function buildBaseAttributesActions({ actions, diff, oldObj, newObj }) {
       const delta = diff[key]
       const before = oldObj[key]
       const now = newObj[key]
-      const isNotDefinedBefore = isNil(oldObj[key])
-      const isNotDefinedNow = isNil(newObj[key])
+      const isNotDefinedBefore = isNil(oldObj[key]) || oldObj[key] === ''
+      const isNotDefinedNow = isNil(newObj[key]) || newObj[key] === ''
```
emmenko: I guess this part is the actual fix?

packages/sync-actions/src/product-types-actions.js
adnasa: accidental commit during an experimentation. Will remove

adnasa:
I had a pretty intense discussion with @butenkor about this, and we both discovered that this could open a can of worms, causing more trouble than it is worth.

The main issue is that these changes would serve one component of our product very well, to which this could cause problems for external importers relying on the `sync-actions` to specifically change values from an `empty string -> null` and vice versa ([ref](https://github.com/commercetools/nodejs/blob/8a6ec455dab6177e53d1ac7f960a779347827c55/docs/sdk/api/syncActions.md#note)). At the moment, we don't have any real numbers as to how many importers there are using this lib.

Also, there is a strong argument to which I agree on as well; the API itself does allow values to move from `empty string -> null|undefined` and vice versa, on different update-actions that the sync-actions shouldn't diverge on, given above.

#### For future reference

Personally, I've learned something here. As mentioned already to @butenkor is that a change to the official SDK that serves a core application well doesn't mean that it is valid for all.

kudos to everyone for taking their valuable time to look into this PR and @butenkor for provided insight as to what consequences this change could lead to.

I'm closing it.

adnasa:
One thing to add here, is that we could make this an opt-in

tdeekens:
Yes. When creating the syncer we could pass options defaulting to current behavior but which can be overwritten. We could even warn if we think some opt-in options are a saner default. We could link back to the docs outlining why. This opt-ins could be per option or a more general thing like a strict mode. So sync actions by default doesnt run in it but you can { strictMode: true }. That could also be used to gracefull introduce breaking changes.

PR #543, opened 2018-04-20
feat(custom-objects-exporter): add new exporter package

### LINKED ISSUE #434: Custom Objects Export
### Description
One of our customers would like to export custom objects via Impex, which is currently not available yet. The customer wants to use the tool to handel important collections eg. 
 - Mapping tables between carrier codes and the different order statuses in CTP.
 - Tables of provinces / regions by country. 
 - Tables of cities related to the corresponding province / region.
 - Collections of additional data of our partners such as names of employees, photographs, certifications, experience, etc.

### AC
- custom objects can be exported via JSON using IMPEX
- this solution will be offered in JSON, not CSV
- a second step will be offering the custom object import (extra issue)

daern91:
#### Summary
PR adds package to export custom objects from CTP in JSON.

#### Todo

- Tests
    - [x] Unit
    - [x] Integration
- [x] Documentation

resolves #434


docs/cli/custom-objects-exporter.md
```diff
@@ -0,0 +1,87 @@
+# Custom Objects Exporter
+
+A package that helps with exporting [commercetools custom objects](https://docs.commercetools.com/http-api-projects-custom-objects.html) in `JSON` format from the [commercetools platform](https://docs.commercetools.com/).
```
adnasa: a package __which exports__...

docs/cli/custom-objects-exporter.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Configuration
+
+The constructor accepts two arguments:
+
+* A required object containing the following values:
+  * `apiConfig` (Object): `AuthMiddleware` options for authentication on the commercetools platform. (Required. See [here](https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html#named-arguments-options))
+  * `accessToken` (String): Access token to be used to authenticate requests to API. Requires scope of [`view_products`], [`view_orders`], [`view_customers`]
```
adnasa: maybe a link to the package or docs that helps dev to retrieve their tokens?

adnasa: why in brackets for `view_products`, btw?

integration-tests/cli/custom-objects-export.it.js
```diff
@@ -0,0 +1,83 @@
...
+    await createData(apiConfig, 'customObjects', createdCustomObjects)
+  }, 30000)
+
+  afterAll(async () => {
+    await clearData(apiConfig, 'customObjects')
+  })
+
+  describe('CLI basic functionality', () => {
+    it('should print usage information given the help flag', async () => {
+      const [stdout, stderr] = await exec(`${bin} --help`)
```
adnasa: maybe move these into a `beforeEach` for higher extensability?

integration-tests/cli/custom-objects-export.it.js
```diff
@@ -0,0 +1,83 @@
...
+      expect(stderr).toBeFalsy()
```
adnasa: I'd rather be explicit over `.toBeFalsy()` to find out the end value

```js
.toBe(undefined|null|false)
```

packages/custom-objects-exporter/readme.md
```diff
@@ -0,0 +1,11 @@
+# custom-objects-exporter
+
+Export [commercetools custom objects](https://docs.commercetools.com/http-api-projects-custom-objects.html) from the commercetools platform.
+
+More info here: https://commercetools.github.io/nodejs/cli/custom-objects-exporter.html
+
+## Install
+
+```bash
+npm i @commercetools/custom-objects-exporter -g
```
adnasa: nitpick: this is inconsistent to the command given in `docs/cli/custom-objects-exporter.md`

docs/cli/custom-objects-exporter.md
daern91: Good one 👍

docs/cli/custom-objects-exporter.md
daern91: Good idea, will add that. 
Regarding the brackets, I saw other packages in the monorepo used brackets so I wanted to keep it consistent.

packages/custom-objects-exporter/src/cli.js
```diff
@@ -0,0 +1,115 @@
+import fs from 'fs'
+import { getCredentials } from '@commercetools/get-credentials'
+import npmlog from 'npmlog'
```
tdeekens: Nice. Didn't know that package.

packages/custom-objects-exporter/src/cli.js
```diff
@@ -0,0 +1,115 @@
...
+
+resolveCredentials(args)
+  .then(credentials => {
+    const apiConfig = {
+      host: args.authUrl,
+      apiUrl: args.apiUrl,
+      projectKey: args.projectKey,
+      credentials,
+    }
+    const constructorOptions = {
```
tdeekens: Hm, isn't the logger also a contructor option? Should we name this according to where it ends up?

tdeekens: Like `exporterOptions`?

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,117 @@
...
+          libraryVersion: pkg.version,
+        }),
+        createHttpMiddleware({ host: this.apiConfig.apiUrl }),
+      ],
+    })
+
+    this.predicate = options.predicate
+
+    this.logger = {
+      error: () => {},
```
tdeekens: You could extract that - maybe in another PR even - into a `utils/silent-logger` and then just

```js
this.logger = {
   ...silentLogger
   ...logger
}
```

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,117 @@
...
+
+  static handleOutput(
+    outputStream: stream$Writable,
+    pipeStream: stream$Writable,
+    instance: Object
+  ) {
+    CustomObjectsExporter.fetchObjects(pipeStream, instance)
+      .then(() => {
+        instance.logger.info('Export operation completed successfully')
+        if (outputStream !== process.stdout) pipeStream.end()
```
tdeekens: I don't understand the semantics behind this check? What does it do?

packages/custom-objects-exporter/test/main.spec.js
```diff
@@ -0,0 +1,98 @@
+import streamtest from 'streamtest'
+import CustomObjectsExporter from '../src/main'
+
+describe('CustomObjectsExporter', () => {
+  const logger = {
+    error: () => {},
+    warn: () => {},
+    info: () => {},
+    verbose: () => {},
```
tdeekens: Again here you could then just `import { silentLogger } from '../utils`

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,117 @@
...
+      })
+      .catch((e: Error) => {
+        outputStream.emit('error', e)
+      })
+  }
+
+  static fetchObjects(output: stream$Writable, instance: Object): Promise<any> {
```
tdeekens: What's `instance` here? I got a bit confused by it being the `instance` of this class or something?

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,117 @@
...
+      request,
+      (data: Object): Promise<any> => {
+        if (data.statusCode !== 200)
+          return Promise.reject(
+            new Error(`Request returned error ${data.statusCode}`)
+          )
+        data.body.results.forEach((object: Buffer) => {
+          output.write(object)
+        })
+        const success = `Successfully exported ${data.body.count} custom object`
```
tdeekens: `successMsg`?

integration-tests/cli/custom-objects-export.it.js
daern91: This one actually returns an empty string. I think `.toBeFalsy()` is a bit prettier than `.toBe('')`, to be honest.

integration-tests/cli/custom-objects-export.it.js
daern91: Not sure I follow. They are checking the output from using different flags?

integration-tests/cli/custom-objects-export.it.js
adnasa: yeah I guess it's prettier. in the end, it's important to understand what the expected value can be and that is conveyed through tests.

without understandable tests, you have a hard-to-understand implementation.
I just mention this since there is no guarantee that we maintain this everyday in conjunction to easing onboarding new members to OSS

just food for thought. 😉

tdeekens: Depends on your intend. `toBeFalsy` actually means you implementation doesn't care if it's `null`, `undefined`, `''` etc. If you typings however don't even allow that the test ain't really inline with that.

docs/cli/custom-objects-exporter.md
wizzy25: [`view_products`], [`view_orders`], [`view_customers`] -> [`view_products`, `view_orders`, `view_customers`]

docs/cli/custom-objects-exporter.md
```diff
@@ -0,0 +1,87 @@
...
+  * `predicate` (String): Query string specifying (where) predicate. More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates) (Optional)
+* An optional logger object having four functions (`info`, `warn`, `error` and `verbose`)
```
wizzy25: I think it'll make more sense to use `debug` as the fourth log level instead of `verbose`. That's also what you pass in [here](https://github.com/commercetools/nodejs/pull/543/files#diff-df95a275d2c8c414f98590424a9d9b75R76)

integration-tests/cli/custom-objects-export.it.js
wizzy25: In this case, I think `.toBeFalsy()` makes more sense as we don't actually care what the error is, as long as it isn't a truthy value, it'll be fine

integration-tests/cli/custom-objects-export.it.js
```diff
@@ -0,0 +1,83 @@
...
+
+    it('should print the module version given the version flag', async () => {
+      const [stdout, stderr] = await exec(`${bin} --version`)
+      expect(stderr).toBeFalsy()
+      expect(stdout).toBe(`${version}\n`)
+    })
+  })
+
+  describe('export function', () => {
+    it(
```
wizzy25: This test seems like a lot of tests that could be further separated into different describe/it blocks... Would be nice if you could refactor this to be more granular

integration-tests/cli/helpers/custom-objects-export.data.js
```diff
@@ -0,0 +1,59 @@
...
+    container: 'Syrinx',
+    key: 'crystalKey',
+    value: {
+      paymentMethod: 'new',
+      paymentID: '3',
+    },
+  },
+]
+
+export const expectedCustomObjects = [
```
wizzy25: Instead of creating two arrays for created and expected, you can create only one, then assert that what is returned from the server contains the object you created. Then this wouldn't be needed

packages/custom-objects-exporter/readme.md
wizzy25: Please write the `npm install` command in full for posterity's sake ;-)

packages/custom-objects-exporter/src/cli.js
wizzy25: We're trying to move away this module to `pino`. Please use `pino` instead

packages/custom-objects-exporter/src/main.js
wizzy25: @tdeekens You can't call `.end()` on process streams (`process.stdout` and `process.stderr`) so without this check, passing a process stream as the output will throw

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,117 @@
...
+    this.logger.info('Starting Export')
+    const jsonStream = JSONStream.stringify()
+    jsonStream.pipe(outputStream)
+    CustomObjectsExporter.handleOutput(outputStream, jsonStream, this)
+  }
+
+  static handleOutput(
+    outputStream: stream$Writable,
+    pipeStream: stream$Writable,
+    instance: Object
```
wizzy25: Instead of passing the entire instance, I think it'll be better to pass only what you actually need

daern91:
Thanks a bunch @tdeekens, @adnasa and @wizzy25. I pushed another commit that hopefully fixes all comments. Could you please have another look?

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,135 @@
...
+
+  static handleOutput(
+    outputStream: stream$Writable,
+    pipeStream: stream$Writable,
+    client: Object,
+    projectKey: string,
+    predicate: ?string,
+    logger: LoggerOptions
+  ) {
+    CustomObjectsExporter.fetchObjects(
```
adnasa: ooooh yeah.. static call!

packages/custom-objects-exporter/src/main.js
```diff
@@ -0,0 +1,135 @@
...
+      ({ statusCode, body }: Object): Promise<any> => {
+        if (statusCode !== 200)
+          return Promise.reject(
+            new Error(`Request returned error ${statusCode}`)
+          )
+        body.results.forEach((object: Buffer) => {
+          output.write(object)
+        })
+        const successMsg = `Successfully exported ${body.count} custom object`
+        logger.debug(body.count > 1 ? `${successMsg}s` : successMsg)
```
adnasa: nitpick.. I would have done this differently.

```js
logger.debug(
  `Successfully exported ${body.count} custom %s`,
  body.count > 1 ? 'objects' : 'object'
)
```

point is, we are using the same message, only the value to the placeholder, or token (in this case, `object`) is different

sorry if I didn't see this earlier

PR #554, opened 2018-04-26
Add `lint-staged`, `jest-runners` and `commitlint`

tdeekens:
#### Summary

In general just modernizes the tool chain a little bit.

#### Description

Started working on feature and noticed that we can improve tooling a bit here. What I set out to do is what we have on other repos already:

1. Use jest-runners and the project feature where possible
   - This speeds up CI etc. We can run flow, eslint and test in jest
2. Add commitlint to lint commit messages
3. Add `lint-staged` to lint staged files (flow and eslint)


tdeekens:
I added @emmenko for awareness and maybe quick glimpse. I literally plugged things together in 15 minutes to show and raise awareness for how to improve things.

package.json
```diff
@@ -10,14 +10,19 @@
...
+    "lint:js": "jest --config jest.eslint.config.js",
+    "test": "cross-env NODE_ENV=development jest --config jest.test.config.js",
     "test:package": "cross-env NODE_ENV=development ./scripts/test-package",
     "pretest:integration": "./scripts/install-npm-packages-for-integration-tests",
-    "test:integration": "cross-env NODE_ENV=development jest --config integration-tests/.jestrc.json",
-    "format": "prettier --write '{packages,integration-tests,types}/**/*.js'",
+    "test:integration": "cross-env NODE_ENV=development jest --config integration-tests/jest.test.config.js",
+    "format": "npm run format:js & npm run format:md",
+    "format:js": "prettier --write '{packages,integration-tests,types}/**/*.js'",
+    "format:md": "prettier --parser markdown --write '**/*.md'",
```
daern91: Nice!

tdeekens:
How is the `types` folder treated. I am a bit confused by it. It lives outside `packages` as a sibling to it. Should it be a package too?

emmenko:
I didn't know where to put it initially. Do you know what the proper convention would be?

tdeekens:
> Do you know what the proper convention would be?

Should it be a package which all other packages require versioned? Feels safer to me as then it can evolve "independently" which out all packages needing the same type.

wizzy25:
> Should it be a package which all other packages require versioned? Feels safer to me as then it can evolve "independently" which out all packages needing the same type.

I think it wouldn't really matter much as it is in the same repo nonetheless. However, if there's a 'convention' of sorts, [Jest](https://github.com/facebook/jest) folder like here for example

tdeekens:
Ok. So no package. All fine by me. How do we solve the linting issue then? Seems to me like imports were alised or something. Ideas?

emmenko:
> How do we solve the linting issue then? Seems to me like imports were alised or something. Ideas?

The alias is defined in the `.babelrc` (look for the `module-resolver` plugin). It seems like only a few packages are failing though, so I'm not sure the alias is the problem. I can have a look later.

emmenko:
Oh, it seems that you @tdeekens changed the imports to relative imports last time you updated the deps 😅

https://github.com/commercetools/nodejs/commit/aa4db7bbbf9aeddb43ff1093cf248912d4ca4dfa#diff-4815dfe53e94b12bcc914cb7818a8e7fL8

image

I'll try to fix it, otherwise we can make it a package.

tdeekens:
> imports to relative imports last time you updated the 

What how? With a VSCode plugin? I can't remember doing that. Sorry :)

emmenko:
No worries, I'm looking into it

emmenko:
Alright, it's fixed now. I removed the babel resolver and used a custom eslint resolver for `types` (same as jest does). See latest commits.

/cc @tdeekens

tdeekens:
Thanks makes sense.

tdeekens:
Rebased out the conflicts.

tdeekens:
Enjoy proper tooling peeps!

PR #556, opened 2018-04-26
Remove `browser` field in `pkg` from all packages

tdeekens:
#### Summary

This removes the `browser` field in all packages.

#### Description

After checking packages such as react-router or Redux we noticed that none specified a browser field in the package. This solves an IE11 issue for any consumer of sdk-client too. Rollup can now consume `browser` fields itself (sounds bizarre) which makes it resolve `isomorphic-fetch` correctly which does not have a `module` field. Without is resolves to the node version of isomorphic-fetch which does not support cookies.

TL;DR: This fixes resolving of modules during build to make sdk-client pull in the correct isomorphic-fetch variation to work in IE11. The `sdk-client` recently got a config options in regards to cookies which breaks whenever used in the node build of `isomorprhic-fetch`. With IE11 being the only browser where the polyfill actually kicks in.

I don't really see this a breaking change for any consumer as all modern bundlers do not look at the browser field. We only bundle in UMD to allow direct importing via a <script>-tag from e.g. a CDN.

Kudos to @luhmann for debugging everything!

### DIFF
--- a/packages/sdk-middleware-auth/src/base-auth-flow.js
+++ b/packages/sdk-middleware-auth/src/base-auth-flow.js
@@ -1,6 +1,5 @@
 /* @flow */
 /* global fetch */
-import 'isomorphic-fetch'
 import type {
   MiddlewareRequest,
   Next,
@@ -34,10 +33,19 @@ function calculateExpirationTime(expiresIn: number): number {
 }
 
 function executeRequest(
-  { url, basicAuth, body, tokenCache, requestState, pendingTasks, response },
+  {
+    fetcher,
+    url,
+    basicAuth,
+    body,
+    tokenCache,
+    requestState,
+    pendingTasks,
+    response,
+  },
   next: Next
 ) {
-  fetch(url, {
+  fetcher(url, {
     method: 'POST',
     headers: {
       Authorization: `Basic ${basicAuth}`,
@@ -92,7 +100,8 @@ function executeRequest(
       })
     })
     .catch((error: Error) => {
-      response.reject(error)
+      if (response && typeof response.reject === 'function')
+        response.reject(error)
     })
 }
 
@@ -106,10 +115,18 @@ export default function authMiddlewareBase(
     pendingTasks,
     requestState,
     tokenCache,
+    fetch: fetcher,
   }: AuthMiddlewareBaseOptions,
   next: Next,
   userOptions?: AuthMiddlewareOptions | PasswordAuthMiddlewareOptions
 ) {
+  if (!fetcher && typeof fetch === 'undefined')
+    throw new Error(
+      '`fetch` is not available. Please pass in `fetch` as an option or have it globally available.'
+    )
+  if (!fetcher)
+    // eslint-disable-next-line
+    fetcher = fetch
   // Check if there is already a `Authorization` header in the request.
   // If so, then go directly to the next middleware.
   if (
@@ -146,6 +163,7 @@ export default function authMiddlewareBase(
   ) {
     executeRequest(
       {
+        fetcher,
         ...buildRequestForRefreshTokenFlow({
           ...userOptions,
           refreshToken: tokenObj.refreshToken,
@@ -163,6 +181,7 @@ export default function authMiddlewareBase(
   // Token and refreshToken are not present or invalid. Request a new token...
   executeRequest(
     {
+      fetcher,
       url,
       basicAuth,
       body,
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -8,9 +8,8 @@ import type {
   MiddlewareResponse,
 } from 'types/sdk'
 
-import 'isomorphic-fetch'
-import parseHeaders from './parse-headers'
 import getErrorByCode, { NetworkError, HttpError } from './errors'
+import parseHeaders from '../src/parse-headers'
 
 function createError({ statusCode, message, ...rest }): HttpErrorType {
   let errorMessage = message || 'Unexpected non-JSON error response'
@@ -64,28 +63,46 @@ export default function createHttpMiddleware({
     retryDelay = 200,
     maxDelay = Infinity,
   } = {},
+  fetch: fetcher,
 }: HttpMiddlewareOptions): Middleware {
+  if (!fetcher && typeof fetch === 'undefined')
+    throw new Error(
+      '`fetch` is not available. Please pass in `fetch` as an option or have it globally available.'
+    )
+
+  if (!fetcher)
+    // `fetcher` is set here rather than the destructuring to ensure fetch is
+    // declared before referencing it otherwise it would cause a `ReferenceError`.
+    // For reference of this pattern: https://github.com/apollographql/apollo-link/blob/498b413a5b5199b0758ce898b3bb55451f57a2fa/packages/apollo-link-http/src/httpLink.ts#L49
+
+    // eslint-disable-next-line
+    fetcher = fetch
+
   return next => (request: MiddlewareRequest, response: MiddlewareResponse) => {
     const url = host.replace(/\/$/, '') + request.uri
     const body =
       typeof request.body === 'string' || Buffer.isBuffer(request.body)
         ? request.body
-        : JSON.stringify(request.body)
+        : // NOTE: `stringify` of `null` gives the String('null')
+          JSON.stringify(request.body || undefined)
     const requestHeader = {
-      'Content-Type': 'application/json',
+      'Content-Type': ['application/json'],
       ...request.headers,
-      ...(body ? { 'Content-Length': Buffer.byteLength(body).toString() } : {}),
+      ...(body
+        ? { 'Content-Length': Buffer.byteLength(body).toString() }
+        : null),
     }
-    const requestObj: Object = new Request(url, {
+    const fetchOptions: Object = {
       method: request.method,
-      headers: new Headers(requestHeader),
+      headers: requestHeader,
       ...(credentialsMode ? { credentials: credentialsMode } : {}),
-      ...(body ? { body } : {}),
-    })
+      ...(body ? { body } : null),
+    }
     let retryCount = 0
     // wrap in a fn so we can retry if error occur
     function executeFetch() {
-      fetch(requestObj).then(
+      // $FlowFixMe
+      fetcher(url, fetchOptions).then(
         (res: Response) => {
           if (res.ok) {
             res.json().then((result: Object) => {
@@ -96,10 +113,10 @@ export default function createHttpMiddleware({
               }
               if (includeResponseHeaders)
                 parsedResponse.headers = parseHeaders(res.headers)
+
               if (includeOriginalRequest) {
                 parsedResponse.request = {
-                  ...requestObj,
-                  headers: parseHeaders(requestObj.headers),
+                  ...fetchOptions,
                 }
                 maskAuthData(parsedResponse.request, maskSensitiveHeaderData)
               }
--- a/.flowconfig
+++ b/.flowconfig
@@ -59,4 +59,4 @@ include_warnings=true
 emoji=true
 
 [version]
-^0.70.0
+^0.71.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## @commercetools/sdk-middleware-auth@4.0.0 (2018-05-10)
+
+#### :boom: Breaking Change
+
+The package is not bundled anymore with `isomorphic-fetch`. Whenever no global `fetch` implementation is availble you will have to pass in a `fetch` implementation as an argument to `createAuthMiddlewareForClientCredentialsFlow`, `createAuthMiddlewareForPasswordFlow`, `createAuthMiddlewareForRefreshTokenFlow`, `createAuthMiddlewareForAnonymousSessionFlow`. More information is also available in the [docs](https://github.com/commercetools/nodejs/blob/master/docs/sdk/api/sdkMiddlewareAuth.md).
+
+## @commercetools/sdk-middleware-http@4.0.0 (2018-05-10)
+
+#### :boom: Breaking Change
+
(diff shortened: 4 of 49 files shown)


tdeekens:
Just an attempt to structure the observations more eloquently again.

1. When building for ESM you can only bundle `isomorphic-fetch` for one target (node or browser)
   - Until now the node version was targeted as `isomorphic-fetch` does not have a `module` field
   - We before used `webpack` for bundling which had the same "issue"
2. By adding adding `browser: true` to Rollup we target the browser build of `isomorphic-fetch`
   - For that also have to remove all `browser` fields from out packages as otherwise bundling fails
   - However, the same "issue" persists only flipped: now the ESM build contains the browser version and won't run under node correctly
     - This may be fine depending if people intend to bundle for Node (e.g. with webpack) by using the `module` field
     - Normally they should resolve to the `main` field but bundlers understand ESM so the before mentioned is a valid option

In summary: 

`isomorphic-fetch` is not a build time dependency. When it runs in node, node will resolve to `main` when building a browser bundle (e.g. via webpack), webpack will go to the `browser` entry. 
When building for one target, you can not include both code for both environments: node and the browser. 
This begs the question as, with most polyfills, also the a fetch polyfill should be a peer dependency (which is a breaking change). The `sdk-client` would then throw if: a.) no `global.fetch` exists or b.) no `fetch` was passed as an option. The latter is to support ponyfills instead of polyfills.

Hope this makes sense @emmenko and @wizzy25.

emmenko:
I had to read this 3 times 🙃

> Without is resolves to the node version of isomorphic-fetch which does not support cookies.

👆 So this is the main problem if I understand correctly, right?

> tobi and me probably found a solution for the ie11-cookie-bug ... the underlying issue was that the node-sdk was built in a way that the browser-field in the package-json of the isomorphic-fetch-polyfill was ignored ... this led to an evaluation of the polyfill as always using the underlying node-sdk-lib, which has no concept of a cookie-jar ... so whenever the polyfill was used in a browser that does not have native fetch-implementation that would have failed since we moved to cookie-based-auth

Right, so this might explain the issue yes. Although I'm pretty sure this was happening on the Apollo queries. Does Apollo client have the same problem? 🤔

> This begs the question as, with most polyfills, also the a fetch polyfill should be a peer dependency (which is a breaking change).

The way you explained this, it seems that this is the only options we have. As long as we follow SemVer and document the breaking change, I don't see a problem of changing this.

And thanks a lot for digging into this 🙌

emmenko:
> Sorry for not being super clear. I literally wrote this up while we were still debugging it.

Ah no worries, I just meant that I had to wrap my head around this to get the point. Thanks for providing as much information as possible! 😄

> Depending on if it gets a fetch passed as an argument or what globally available fetch implementation of polyfill it consumes: yes

Yeah, the `apollo-link-http` takes the global fetch, but we can specifically pass one.

https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-http/src/httpLink.ts#L36-L37

So yeah, we should fix this problem the way you described it 👌

tdeekens:
> Yeah, the apollo-link-http takes the global fetch, but we can specifically pass one.

Yes that's what we should do too. You pass in a `fetch` or we assume there is a global fetch. If none is there we throw.

tdeekens:
> No because of 👇

Sorry for beig unclear. I meant we would have the same problem uf we would not pass unfetch. 

So after all we all agree to change sdk-client is the best here?

luhmann:
> So after all we all agree to change sdk-client is the best here?

I am right now reading up about peerDependencies especially in conjunction with rollup again as I am a bit fuzzy about how this would exactly solve the problem. But my lack of knowledge should not stop you. The proposed upgrade path of people having to update and merely add a dependency seems reasonable to me.

tdeekens:
> Ah, so it would cause isomorphic-fetch being loaded/bundled by the consumer of the node-sdk

Jup. From an app

```js
import fetch from 'unfetch';

createClient({ fetch });
```

from node

```js
import fetch from 'isomorphic-fetch';

createClient({ fetch });
```

in `createClient` we "just"

```js
const fetcher = args.fetch || fetch; // falling back to global

if (!fetcher) throw new Error('@commercetools/sdk-client: no `fetch` available. Please pass in `fetch` or have it globally available');
```

luhmann:
Aaaaaah! You even want to change the code 😃 and force people to change the way they initialize. Even cleaner than what I imagined!

tdeekens:
1. Change code: yes. 
2. Force them how to initialise: half half. Those with a globally available fetch will continue to work. Mostly node consumers will break as they have to pass node-fetch or iso-fetch

emmenko:
Yep, I agree with you guys 👌

emmenko:
@tdeekens should we do those breaking changes in this PR?

tdeekens:
Yes. I would like to do it here. Without this is breaking nonetheless but also worthless.

tdeekens:
This increased in scope in that it:

1. Changed all CLI packages to use `node-fetch` passing it to the respective middlewares
2. Validates and checks where needed that either a global `fetch` is available or a `fetch` is passed
3. Tweaks all tests (including integration tests) to pass in test
   - This adds the functionality to remove `nock` but that's another story (as `fetch` can be a `jest.fn` now

tdeekens:
@emmenko and @luhmann I think this is ready for review. I will link this into the mc-fe (note we should not forget the mc-be) to see how well it behaves and if it fixes IE11.

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -92,7 +100,8 @@ function executeRequest(
       })
     })
     .catch((error: Error) => {
-      response.reject(error)
+      if (response && typeof response.reject === 'function')
```
tdeekens: Seems like with node-fetch in some cases you actually can not and do not have to reject. At least that's what the tests say if you leave this check out.

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -106,10 +115,18 @@ export default function authMiddlewareBase(
...
+    fetch: fetcher,
   }: AuthMiddlewareBaseOptions,
   next: Next,
   userOptions?: AuthMiddlewareOptions | PasswordAuthMiddlewareOptions
 ) {
+  if (!fetcher && typeof fetch === 'undefined')
+    throw new Error(
+      '`fetch` is not available. Please pass in `fetch` as an option or have it globally available.'
+    )
+  if (!fetcher)
```
tdeekens: I am open to suggestions here. I stole this from apollo-link-http assuming that they know better than we.

packages/sdk-middleware-auth/src/base-auth-flow.js
tdeekens: With `node-fetch` it seems like you do not always have to now can reject. The tests showed this which is why I added the guard.

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -106,10 +115,18 @@ export default function authMiddlewareBase(
     pendingTasks,
     requestState,
     tokenCache,
+    fetch: fetcher,
   }: AuthMiddlewareBaseOptions,
   next: Next,
   userOptions?: AuthMiddlewareOptions | PasswordAuthMiddlewareOptions
 ) {
+  if (!fetcher && typeof fetch === 'undefined')
```
luhmann: Will this work in node too, if someone included a polyfill? 

Or should this test for `global.fetch` or `window.fetch`?

packages/sdk-middleware-http/src/http.js
```diff
@@ -64,28 +63,46 @@ export default function createHttpMiddleware({
...
     const url = host.replace(/\/$/, '') + request.uri
     const body =
       typeof request.body === 'string' || Buffer.isBuffer(request.body)
         ? request.body
-        : JSON.stringify(request.body)
+        : // NOTE: `stringify` of `null` gives the String('null')
+          JSON.stringify(request.body || undefined)
     const requestHeader = {
-      'Content-Type': 'application/json',
+      'Content-Type': ['application/json'],
```
luhmann: How I read the whatwg-fetch-spec passing an array actually does not seem to work. However it works: 

`await fetch(location.href, { headers: { 'Content-Type': ['application/json', 'application/javascript']}})`

leads to a concatenated string for `Content-Type` so 🤷‍♂️

tdeekens: Good to know. I guess the tests are focussed on `node-fetch` (which is what's passed under test).

packages/sdk-middleware-auth/src/base-auth-flow.js
tdeekens: To my knowledge it should. In the browser the default global is `window` in node it's `global`.

tdeekens:
I re-added the resolution to `jsnext` and the browser entries. As with the extracted fetching dependency this will continue to work just fine.

As a last note: the only breaking change with this is for Node.js users who do not have `isomorphic-fetch` polyfilling things by default in their apps. They will have to pass in `fetch`.

tdeekens:
@wizzy25 and @daern91 this is what we talked about yesterday.

packages/sdk-middleware-auth/src/base-auth-flow.js
emmenko: The `response.reject` I think is from the SDK middleware and it should always be defined

https://github.com/commercetools/nodejs/blob/d7fce290e3ff5c65b59637ee5b79faefad73e06f/packages/sdk-client/src/client.js#L66-L72

So I'm not sure this is necessary...

packages/sdk-middleware-auth/src/base-auth-flow.js
emmenko: I think it's fine yeah 👍

packages/sdk-middleware-http/src/http.js
emmenko: Hmm I'm confused on why we need this 🧐

packages/sdk-middleware-http/src/http.js
```diff
@@ -64,28 +63,46 @@ export default function createHttpMiddleware({
...
       ...request.headers,
-      ...(body ? { 'Content-Length': Buffer.byteLength(body).toString() } : {}),
+      ...(body
+        ? { 'Content-Length': Buffer.byteLength(body).toString() }
+        : null),
     }
-    const requestObj: Object = new Request(url, {
```
emmenko: Thanks 🙏

packages/sdk-middleware-http/src/http.js
```diff
@@ -64,28 +63,46 @@ export default function createHttpMiddleware({
...
       ...(credentialsMode ? { credentials: credentialsMode } : {}),
-      ...(body ? { body } : {}),
-    })
+      ...(body ? { body } : null),
+    }
     let retryCount = 0
     // wrap in a fn so we can retry if error occur
     function executeFetch() {
-      fetch(requestObj).then(
+      // $FlowFixMe
```
emmenko: I don't know this one 🤔

packages/sdk-middleware-http/src/http.js
tdeekens: How a list of headers for one key is mangled differs among the `fetch` implementations. `node-fetch` makes them an array by default. As a result we align our expectations with it. Does that make sense? If we'd use a native `fetch` in the browser a list of headers for a given key would lead to a concatenated string.

packages/sdk-middleware-http/src/http.js
tdeekens: It's otherwise not really type-able with Flow. At this point we know given the destructuring above and the setting of fetcher that it will always be defined or we would have thrown an error. Flow doesn't notice this. I have the same "issue" in parts with flopflip where I have to aggressively guard which we can't here. $FlowFixMe is like `eslint-ignore-next-line`.

packages/sdk-middleware-http/src/http.js
tdeekens: Ah I meant to comment on this too. It was immensely pointless to do this. 80% of all fetch implementation wrap the options into a Request anyway and the options.headers into a Headers.

packages/sdk-middleware-auth/src/base-auth-flow.js
tdeekens: I tried again to remove it. It leads to "ensure to fetch new token only once and keep track of pending tasks" to fail. 

It's an integration test which was running on Node.js with `isomorphic-fetch` having pulled in the `whatwg-fetch` polyfill before. Now it runs with `node-fetch`. I would like to keep the guard.

Regarding the "Initial response shape": I didn't touch that bit and would like to keep it if we're both not sure what it's for.

tdeekens:
> I would also update the changelog with the "breaking changes" for the related packages.

Happened in 9c6451b

I am now hitting the integration tests to find all places where we need to pass in a `fetch` implementation now.

tdeekens:
@emmenko or @wizzy25 can you check the last failing integration tests? I really do not know why they fail...

packages/sdk-middleware-http/src/http.js
emmenko: Ok but just to be clear: it works also with just a string right?

packages/sdk-middleware-http/src/http.js
emmenko: Thanks for the explanation

emmenko:
No idea...is it working on master?

tdeekens:
> No idea...is it working on master?

No. Then I will check again and figure out if I can find the issue.

daern91:
@tdeekens commit e98afb1 breaks the int-tests for me locally but re-running it on CI made it pass. All green atm...

Do you know why we changed the `Authorization` header to be capitalised in this PR?

tdeekens:
> Do you know why we changed the Authorization header to be capitalised in this PR?

Because now we use `node-fetch` which does not lowercase the headers passed in. We receive the same headers we pass in passed out now.

daern91:
Makes sense, great!

emmenko:
> anything you see missing?

Not that I can see.

tdeekens:
> Not that I can see.

Can you approve 💃

tdeekens:
I linked this into the MC and everything works. I am as confident as I can be that this does not have unanticipated side-effects.

PR #570, opened 2018-05-04
fix(product-export): throw error when no products are found

### LINKED ISSUE #553: bug(product-json-to-csv): Error with empty JSON
### Support reference number
https://jira.commercetools.com/browse/SUPPORT-2927

### Description
Problem here is when the predicate delivers an exported products file that is just an empty array the package will throw an error. 

### Expected Behavior
It should export an empty csv file instead.

daern91:
#### Summary

PR makes product-export throw error `No products found` when predicate is too strict or there simply isn't any products.

#### Description

We used to export empty JSON files which 
a) Doesn't make sense to clients as they would prefer to get an error instead of downloading an empty file.
b) Caused `json-to-csv` parser package to break since it is not built to handle empty JSON files.

Note that this is a breaking change so we need to make sure everyone is aware of the new behaviour.

resolves #553

### DIFF
--- a/packages/product-exporter/src/main.js
+++ b/packages/product-exporter/src/main.js
@@ -90,6 +90,9 @@ export default class ProductExporter {
       .process(
         request,
         ({ body: { results: products } }: ProcessFnResponse): Promise<*> => {
+          if (products.length < 1) {
+            return Promise.reject(Error('No products found'))
+          }
           this.logger.debug(`Fetched ${products.length} products`)
           ProductExporter._writeEachProduct(outputStream, products)
           this.logger.debug(
--- a/packages/product-exporter/src/main.js
+++ b/packages/product-exporter/src/main.js
@@ -90,14 +90,14 @@ export default class ProductExporter {
       .process(
         request,
         ({ body: { results: products } }: ProcessFnResponse): Promise<*> => {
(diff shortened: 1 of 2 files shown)


daern91:
@tdeekens quite small me thinks. If anyone is expecting an empty json file when attempting to export inexistent data they will just get the error instead. But, nevertheless, its still breaking.

tdeekens:
Yeah totally agree. Can be sold as an improvement or fix :)

packages/product-exporter/src/main.js
```diff
@@ -95,6 +95,9 @@ export default class ProductExporter {
           this.logger.debug(
             `${products.length} products written to outputStream`
           )
+          if (products.length < 1) {
+            return Promise.reject(Error('No products found'))
```
wizzy25: You should reject with a `new` error instance here

daern91: There's actually no need for that as they both do the same
> When Error is used like a function -- without new, it will return an Error object. Therefore, a mere call to Error will produce the same output that constructing an Error object via the new keyword would.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

PR #581, opened 2018-05-07
feat(discount-code-export): add support for export template

wizzy25:
affects: @commercetools/discount-code-exporter

Also define default column headers
add default language for normal export

#### Summary
This PR fixes a bug where certain columns were missing from the export

#### Description
This PR adds the possibility to export discount codes with a template.

Without a template, certain default columns are always exported, and custom fields are not included hence the **breaking change**

Right now, these fields are : `name`, `description`, `code`, `cartDiscounts`,`cartPredicate`,`groups`,`isActive`,`validFrom`,`validUntil`,`references`,`maxApplications`,`maxApplicationsPerCustomer`


docs/cli/discount-code-exporter.md
```diff
@@ -9,11 +9,15 @@ The constructor accepts two arguments:
...
+  * `language` (String): Language used for localised fields if no template is given (Optional. Default: `'en'`)
   * `accessToken` (String): Access token to be used to authenticate requests to API. Requires scope of [`view_orders`]
   * `delimiter` (String): CSV delimiter (Optional. Default: `','`)
   * `multiValueDelimiter` (String): CSV delimiter used in multivalue fields (Optional. Default: `';'`)
   * `exportFormat` (String): Export format ['csv', 'json'] (Optional. Default: 'json')
   * `predicate` (String): Query string specifying (where) predicate. More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates) (Optional)
+  * `headerFields` (Array<String>): An array of column names the exported CSV file should contain. This headerFields array should contain the required columns of the CSV file(Optional. If omitted, a default set of column fields is used. Currently, these fields are: `name`, `description`, `code`, `cartDiscounts`,`cartPredicate`,`groups`,`isActive`,`validFrom`,`validUntil`,`references`,`maxApplications`,`maxApplicationsPerCustomer`.
```
tdeekens: `file(Optional` -> `file (Optional`

tdeekens: If we use the term `headerFields` in other packages all good. I felt `fields` might already be enough.

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -57,70 +57,146 @@ describe('DiscountCodeExport', () => {
...
+          expect(DiscountCodeExport.setupHeaders([], 'de')).toMatchSnapshot()
+        })
       })
-      const outputStream = streamtest.v2.toText((error, result) => {
-        const expectedResult = [sampleCode]
-        expect(JSON.parse(result)).toEqual(expectedResult)
-        done()
+
+      describe('when null', () => {
+        test('should return return passed headers', () => {
```
tdeekens: return ~return~

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -57,70 +57,146 @@ describe('DiscountCodeExport', () => {
...
-        code: 'discount-code',
-        name: { en: 'some-discount-name' },
-        cartDiscounts: [{ id: 'cart-discount-1' }, { id: 'cart-discount-2' }],
-      }
-      codeExport._fetchCodes = jest.fn().mockImplementation(jsonStream => {
-        jsonStream.write(sampleCode)
-        return Promise.resolve()
+    describe('without passed headers', () => {
+      describe('when empty array', () => {
+        test('should return return passed headers', () => {
```
tdeekens: return ~return~

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -57,70 +57,146 @@ describe('DiscountCodeExport', () => {
...
+          expect(DiscountCodeExport.setupHeaders([], 'de')).toMatchSnapshot()
```
tdeekens: Nitpick: but for this I would not use a snapshot. I sense it'd be easier to have `expect(DiscountCodeExport.setupHeaders([], 'de')).toEqual([])` as then the expected value is co-located. It's not like it will change or contains values and the nice CLI diffing of snapshotting it provides any benefit.

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -57,70 +57,146 @@ describe('DiscountCodeExport', () => {
...
+  describe('::run', () => {
+    describe('successful export', () => {
+      let sampleCode
+      beforeEach(() => {
+        sampleCode = {
+          code: 'discount-code',
+          name: { en: 'some-discount-name' },
+          cartDiscounts: [{ id: 'cart-discount-1' }, { id: 'cart-discount-2' }],
+        }
+        codeExport._fetchCodes = jest.fn().mockImplementation(csvStream => {
```
tdeekens: Why `jest.fn().mockImplementation`? I know the docs state so but if we never `mockImplementation` later on wouldn't a `jest.fn((csvStream => {}))` be enough? Even then we could `mockImplementation` later?

packages/discount-code-exporter/test/main.spec.js
```diff
@@ -57,70 +57,146 @@ describe('DiscountCodeExport', () => {
...
+              {
+                apiConfig: {
+                  projectKey: 'test-project-key',
+                },
+                exportFormat: 'csv',
+                headerFields: ['code', 'name.en', 'cartDiscounts', 'groups'],
+              },
+              logger
+            )
+            codeExport._fetchCodes = jest.fn().mockImplementation(csvStream => {
```
tdeekens: I know we talked about this and it's unrelated to this PR but it feels more and more that the `fetchCodes` should be a module in itself. It would make testing way easier. The `DiscountCodeExport` could just `import { fetchCodes } from './fetch-codes'` but the test would just `jest.mock('./fetch-codes')` while also `import { fetchCodes } from './fetch-codes'`. It can then just use `fetchCodes.mockImplementation`.

packages/discount-code-exporter/test/main.spec.js
wizzy25: I'm actually not expecting an empty array to be returned in this case, but yeah. snapshotting doesn't help here

packages/discount-code-exporter/test/main.spec.js
wizzy25: True. I see a lot of sense in that. But like you rightly said, this isn't the PR for it. Nevertheless I think it's something that should be done at some point

packages/discount-code-exporter/src/main.js
```diff
@@ -66,14 +69,40 @@ export default class DiscountCodeExport {
...
+          'code',
+          'cartDiscounts',
+          'cartPredicate',
+          'groups',
+          'isActive',
+          'validFrom',
+          'validUntil',
+          'references',
+          'maxApplications',
+          'maxApplicationsPerCustomer',
```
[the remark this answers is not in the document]
wizzy25: @daern91 Yeah. That makes sense

wizzy25:
Refactored :)
@daern91 @tdeekens

PR #592, opened 2018-05-15
feat(sync-actions): configure empty string as not-a-value

adnasa:
#### Summary

- add configuration to `createSyncProductTypes` to treat empty string as not a value
- follow up to https://github.com/commercetools/nodejs/pull/541

#### Description

So at the project I work on, we still use work-arounds to treat empty string, `undefined` and `null` as not a value in our forms. This is something that we wish is already supported on the `nodejs/sync-actions`.

Unfortunately, on #541 we learned that the changes in that PR is a breaking change across all sync-action types and can potentially break clients using the `nodejs/sync-actions` and still has the use case of __empty string being a value__

This time I suggest it to be an __opt-in__

For now, I added support only on product-types. We can add support to other types as a separate step essentially when it is needed. There won't be much repetition/redundancy, except on the entry point and on the call of `buildBaseAttributesActions`

#### Todo

* [ ] Figure out a better name to `shouldOmitEmptyString`
* Tests
  * [x] Unit

### DIFF
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -21,7 +21,7 @@ export const baseActionsList = [
   { action: 'changeDescription', key: 'description' },
 ]
 
-export function actionsMapBase(diff, previous, next, config) {
+export function actionsMapBase(diff, previous, next, config = {}) {
   return buildBaseAttributesActions({
     diff,
     actions: baseActionsList,
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -28,7 +28,6 @@ export function actionsMapBase(diff, previous, next, config) {
     oldObj: previous,
     newObj: next,
     shouldOmitEmptyString: config.shouldOmitEmptyString,
-    // omitUnecessaryUpdates: config.shouldOmitEmptyString,
   })
 }
 
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -21,12 +21,14 @@ export const baseActionsList = [
   { action: 'changeDescription', key: 'description' },
 ]
 
-export function actionsMapBase(diff, previous, next) {
+export function actionsMapBase(diff, previous, next, config) {
   return buildBaseAttributesActions({
     diff,
     actions: baseActionsList,
     oldObj: previous,
     newObj: next,
+    shouldOmitEmptyString: config.shouldOmitEmptyString,
+    // omitUnecessaryUpdates: config.shouldOmitEmptyString,
   })
 }
 
--- a/packages/sync-actions/src/product-types.js
+++ b/packages/sync-actions/src/product-types.js
@@ -46,7 +46,7 @@ function createProductTypeMapActions(
 
 export default (
   actionGroupConfig: Array<ActionGroup>,
-  config: Object
+  config: Object = {}
 ): SyncAction => {
   const mapActionGroup = createMapActionGroup(actionGroupConfig)
   const doMapActions = createProductTypeMapActions(mapActionGroup, config)
--- a/packages/sync-actions/src/product-types.js
+++ b/packages/sync-actions/src/product-types.js
@@ -13,7 +13,8 @@ function createProductTypeMapActions(
   mapActionGroup: (
     type: string,
     fn: () => Array<UpdateAction>
-  ) => Array<UpdateAction>
+  ) => Array<UpdateAction>,
+  config: Object
 ): (diff: Object, next: Object, previous: Object) => Array<UpdateAction> {
   return function doMapActions(
     diff: Object,
@@ -23,7 +24,7 @@ function createProductTypeMapActions(
     const allActions = []
     allActions.push(
       mapActionGroup('base', (): Array<UpdateAction> =>
-        productTypeActions.actionsMapBase(diff, previous, next)
+        productTypeActions.actionsMapBase(diff, previous, next, config)
       ),
       mapActionGroup('attributes', (): Array<UpdateAction> =>
         productTypeActions.actionsMapAttributes(
@@ -43,9 +44,12 @@ function createProductTypeMapActions(
   }
 }
 
-export default (config: Array<ActionGroup>): SyncAction => {
-  const mapActionGroup = createMapActionGroup(config)
-  const doMapActions = createProductTypeMapActions(mapActionGroup)
+export default (
+  actionGroupConfig: Array<ActionGroup>,
+  config: Object
+): SyncAction => {
+  const mapActionGroup = createMapActionGroup(actionGroupConfig)
+  const doMapActions = createProductTypeMapActions(mapActionGroup, config)
   const buildActions = createBuildActions(diffpatcher.diff, doMapActions)
   return { buildActions }
 }
--- a/packages/sync-actions/src/product-types.js
+++ b/packages/sync-actions/src/product-types.js
@@ -1,6 +1,11 @@
 /* @flow */
 import flatten from 'lodash.flatten'
-import type { SyncAction, UpdateAction, ActionGroup } from 'types/sdk'
+import type {
+  SyncAction,
+  UpdateAction,
+  ActionGroup,
+  SyncActionConfig,
+} from 'types/sdk'
 import createBuildActions from './utils/create-build-actions'
 import createMapActionGroup from './utils/create-map-action-group'
 import * as productTypeActions from './product-types-actions'
@@ -14,7 +19,7 @@ function createProductTypeMapActions(
     type: string,
     fn: () => Array<UpdateAction>
   ) => Array<UpdateAction>,
-  config: Object
+  config: SyncActionConfig
 ): (diff: Object, next: Object, previous: Object) => Array<UpdateAction> {
   return function doMapActions(
     diff: Object,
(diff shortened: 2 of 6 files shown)


packages/sync-actions/src/utils/common-actions.js
```diff
@@ -42,7 +61,7 @@ export function buildBaseAttributesActions({ actions, diff, oldObj, newObj }) {
       const patched = diffpatcher.patch(clone(before), delta)
       return { action: item.action, [actionKey]: patched }
     })
-    .filter(action => action)
+    .filter(action => !isNil(action))
```
adnasa: Not part of the PR's purpose, but I chose to be explicit here

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -11,17 +17,30 @@ import * as diffpatcher from './diffpatcher'
...
  * @param  {Object} options.newObj - the new representation of the object
+ * @param {Boolean} options.shouldOmitEmptyString - a flag to determine if we should treat an empty string a NON-value
  */
-export function buildBaseAttributesActions({ actions, diff, oldObj, newObj }) {
+export function buildBaseAttributesActions({
+  actions,
+  diff,
+  oldObj,
+  newObj,
+  shouldOmitEmptyString,
```
adnasa: I welcome suggestions on good names to this.
Please use the thread below

packages/sync-actions/src/product-types.js
```diff
@@ -13,7 +13,8 @@ function createProductTypeMapActions(
   mapActionGroup: (
     type: string,
     fn: () => Array<UpdateAction>
-  ) => Array<UpdateAction>
+  ) => Array<UpdateAction>,
+  config: Object
```
adnasa: I wanted to use `types/sdk` here, but I get flow errors, so I went back to just `Object`
I could use some help to introduce the following:

```
export const SyncActionConfig = {
  ourFlag: Boolean,
}
```

adnasa:
what the... tests are breaking. Looking into that right away

packages/sync-actions/src/product-types.js
wizzy25: What lint errors did you get here?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -21,12 +21,14 @@ export const baseActionsList = [
...
 
-export function actionsMapBase(diff, previous, next) {
+export function actionsMapBase(diff, previous, next, config) {
   return buildBaseAttributesActions({
     diff,
     actions: baseActionsList,
     oldObj: previous,
     newObj: next,
+    shouldOmitEmptyString: config.shouldOmitEmptyString,
+    // omitUnecessaryUpdates: config.shouldOmitEmptyString,
```
wizzy25: I think you want to remove this comment ;)

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -2,6 +2,12 @@ import isNil from 'lodash.isnil'
 import clone from './clone'
 import * as diffpatcher from './diffpatcher'
 
+export function getIsAnEmptyValue(value, shouldOmitEmptyString) {
```
wizzy25: Not fixated on this, but it may be a good idea to pass a default value

packages/sync-actions/src/utils/common-actions.js
wizzy25: Two hard problems in Computer Science: cache invalidation and *naming things* 😄

packages/sync-actions/src/product-types-actions.js
adnasa: maybe I wasn't clear earlier, but I left that comment there as a suggestion to naming.
I would need help in finding a good name for this since `shouldOmitEmptyString` is just... bad 😄

packages/sync-actions/src/utils/common-actions.js
tdeekens: Are you sure that filters out the exact same things?

adnasa:
@tdeekens 

> We intend to keep the current behaviour forever?

I'm going to assume that you are referring to the _overall behaviour of the nodejs package_:
NO, I don't want that, and I think we should __stop being so scared__ of introducing breaking changes and __start leveraging transparency__ to communicate breaking change. My problem here is that in my experience with `nodejs`, it's inflexible as a product to introduce new behaviour.
But at least here, we are stepping in the right direction. Make it configurable.

> We will never intend to migrate towards your proposed handling through the config?

I'm going to assume that you are referring to _having this behaviour across all types_ in the `sync-actions`: Yes, we should? let's start with product-types, evaluate __the approach__, and once approved I'm more than happy to add them to all types

tdeekens:
Alright. Thanks for the explanation. Given what you said, and I don't hold strong opinions on this, why don't we:

1. Add a `{ strictMode: true }` (or named differently) to the product type syncer now
   - It will do what your config does now. It additionally `console.warns` whenever an empty string is found.

The idea is that users can opt-into what we plan to deprecate in the future early. Also do see what that means for them. The idea is a bit inspired from React with the `<StrictMode>`. This is also why the name could be different.

We can communicate the `strictMode` as an option which maybe even groups soon to be deprecated behaviour and runs the syncer in a, how we intend it to be, future proof way.

packages/sync-actions/src/utils/common-actions.js
adnasa: We seem to always return `undefined` above whenever we aren't supposed to generate an update-action

packages/sync-actions/src/utils/common-actions.js
tdeekens: This `getIsBoolean` naming convention is driving me nuts :)

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -2,6 +2,12 @@ import isNil from 'lodash.isnil'
...
+  const isEmptyString = v => Boolean(typeof v === 'string' && v.trim() === '')
```
tdeekens: As another thread: We could also change the signature so that it receives what it's meant to treat as an empty value. `isAnEmptyValue(value, [undefined, null]` per default and with `shouldOmitEmptyString` it just receives `isAnEmptyValue(value, [undefined, null, '']`. What do you think?

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -11,17 +17,30 @@ import * as diffpatcher from './diffpatcher'
...
       const key = item.key // e.g.: name, description, ...
       const actionKey = item.actionKey || item.key
       const delta = diff[key]
       const before = oldObj[key]
       const now = newObj[key]
-      const isNotDefinedBefore = isNil(oldObj[key])
-      const isNotDefinedNow = isNil(newObj[key])
+      const isNotDefinedBefore = getIsAnEmptyValue(
+        oldObj[key],
+        shouldOmitEmptyString
```
tdeekens: With my comment above this could become: `shouldOmitEmptyString ? [undefined, null, ''] : [undefined, null]`.

packages/sync-actions/src/utils/common-actions.js
adnasa: hmmm... yes I get the idea, but I wonder what that signature would look like:

On the top of my head, that could be

```js
const isEmptyValue = (value, emptyValues) => emptyValues.includes(value)
```

but we still would need the trimming, since the user could potentially pass in `' '` as a way to trick the update-actions previously.

adnasa: I actually prefer the original approach

tdeekens: No, the user can't specify that. The user would not specify the `emptyValues`. A consumer would still only specify `shouldOmitEmptyString` which translates to a static `[undefined, null, '']`.

packages/sync-actions/src/utils/common-actions.js
adnasa: no we don't. we return `undefined` above

packages/sync-actions/src/product-types.js
adnasa: forget it, I effed up

packages/sync-actions/src/utils/common-actions.js
adnasa: > No, the user can't specify that. 

no, they don't specify the config, but they do specify the value? that is what I was referring to with _since the user could potentially pass in ' ' as a way to trick the update-actions previously._

tdeekens: Sorry for still failing to get my point across. We don't change the user facing configuration options

1. We keep the config as is with `shouldOmitEmptyString`
2. We however don't pass around the `shouldOmitEmptyString` so deeply instead the `getIsAnEmptyValue` becomes

```js
const normalizeValue = value => typeof value === 'string' ? value.trim() : value;
const createIsEmptyValue = emptyValues => value => 
  emptyValues.some(emptyValue => emptyValue === normalizeValue(value))
```

In this line https://github.com/commercetools/nodejs/pull/592/files#diff-d544c8548498969be35940fcda6c3386R36

We can then there do a

```js
const defaultEmptyValues = [null, undefined];
const isEmptyValue = createIsEmptyValue(options.shouldOmitEmptyString ? emptyValues.concat('') : emptyValues)
```

To be that would be easier to understand which is subjective obviously. Again as previously stated I am fine either way. So feel free to pick or drop.

adnasa: > We however don't pass around the shouldOmitEmptyString so deeply instead the getIsAnEmptyValue becomes

okay, now I get it 😄 
I like your suggested approach better. update coming up!

adnasa: df7aa27dd06bc5cb9ce95cfb03976e3975f809ce

PR #593, opened 2018-05-17
fix(sync-actions/product-types): product types doesn't generate update actions for enum changes

adnasa:
#### Summary

- generates update-actions for changes in enum
- removes `get*` and use simpler assignment

#### Description

for the moment, the `sync-actions/products` assumes the following shape of the attribute when the type is enum;

```
{
  type: {
    name: 'enum'
  }
  values: []
}
```

which results `sync-actions` not generating appropriate update actions when there is a change in `values` of enums.

##### Correct shape

```js
{
  type: {
    name: 'enum'
    values: []
  }
}
```

#### Todo

* Tests
  * [x] Unit

### DIFF
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -50,21 +50,20 @@ const getIsAnAttributeDefinitionBaseFieldChange = diff =>
   diff.isSearchable
 
 function actionsMapEnums(attributeType, attributeDiff, previous, next) {
-  const getAddActionName = type =>
-    type === 'enum' ? 'addPlainEnumValue' : 'addLocalizedEnumValue'
-  const getChangeEnumOrderActionName = type =>
-    type === 'enum'
+  const addEnumActionName =
+    attributeType === 'enum' ? 'addPlainEnumValue' : 'addLocalizedEnumValue'
+  const changeEnumOrderActionName =
+    attributeType === 'enum'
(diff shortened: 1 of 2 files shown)


packages/sync-actions/src/product-types-actions.js
```diff
@@ -188,13 +186,13 @@ export function actionsMapAttributes(
...
         )
-      } else if (diffValue.values) {
+      } else if (diffValue.type.values) {
         actions.push(
           ...actionsMapEnums(
             extractedPairs.oldObj.type.name,
-            diffValue,
-            extractedPairs.oldObj,
-            extractedPairs.newObj
+            diffValue.type,
```
tdeekens: This is the actual fix, correct?

adnasa: yes this is the actual fix.
Also, regarding your concern about test, I did adjust `createAttributeDefinition` to match the right shape.

packages/sync-actions/test/product-types-sync-enums.spec.js
```diff
@@ -6,11 +6,8 @@ const createTestProductType = custom => ({
   ...custom,
 })
 
-const createAttributeDefinition = custom => ({
-  type: {
-    name: custom.name,
-  },
-  ...custom,
+const createAttributeDefinition = type => ({
```
dferber90: Do we need this function anymore? At this point it seems simpler to just inline the objects.

PR #601, opened 2018-05-21
feat(personal-data-erasure): add new customer erasure package

### LINKED ISSUE #580: GDPR tool
### Support reference number

please see CTP-1229 and CTP-1230 for all information

ideally 25th May 2018

@daern91 @cneijenhuis

daern91:
#### Summary

As the new GDPR rules come into play May 25th we need to offer tools to delete all data existing on a customer if requested. It was proposed to do this as an open source package and this PR is intended to provide just those tools.

#### Description

Me and @cneijenhuis worked together to put this together the other day real quick. It is not yet finished but I would love to get your feedback on it so far. Since I am on vacation the rest of this week it would help a lot in order to wrap it up quickly.

More info in below links:
https://jira.commercetools.com/browse/CTP-1229
https://jira.commercetools.com/browse/CTP-1230
https://deploy-preview-963--commercetools-documentation.netlify.com/release-notes.html#releases-2018-05-09-data-erasure

#### Todo

* Tests
  * [x] Unit
  * [x] Integration
* [x] Documentation
* [ ] Flow types
* [x] Logging
* [x] Refactor for readability 

resolves #580


packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+      customersUri,
+      ordersUri,
+      cartsUri,
+      paymentsUri,
+      shoppingsListsUri,
+      reviewsUri,
+    ]
+
+    // todo change to process instead of execute
+    // in case of more than 500 results
```
butenkor: This quite important TODO. Will this be a part of this PR?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+    getUris.forEach(uri => {
+      const request = GDPRTool.buildRequest(uri.uri, 'GET')
+
+      this.client
+        .execute(request)
+        .then(async payload => {
+          if (payload.statusCode !== 200) return Promise.reject(payload)
```
butenkor: Will it also reject if status code is 404 which seems acceptable?

tdeekens: Nope not to my knowledge unless the `client.execute` implements that deliberately. `fetch` (as per spec) rejects only when a network error is encountered not when a 5xx or 4xx comes along.

integration-tests/cli/helpers/utils.js
```diff
@@ -60,3 +61,23 @@ export function createData(apiConfig, entityName, data) {
...
+  const service = createRequestBuilder({
+    projectKey: apiConfig.projectKey,
+  })[entityName]
+
+  const request = {
+    uri: service.build(),
+    method: 'GET',
+  }
+
+  return client.execute(request)
```
tdeekens: Sorry but how is this getting the id only?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+    const shoppingsListsUri = requestBuilder.shoppingLists
+      .where(`customer(id = "${customerId}")`)
+      .perPage(500)
+      .build()
+    const reviewsUri = requestBuilder.reviews
+      .where(`customer(id = "${customerId}")`)
+      .perPage(500)
+      .build()
+
+    const getUris = [
```
tdeekens: This ain't a getter maybe it's `urisOfResourcesToDelete`?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+
+          const messagesUri = requestBuilder.messages
+            .where(reference)
+            .perPage(500)
+            .build()
+
+          const request = GDPRTool.buildRequest(messagesUri, 'GET')
+          const messages = await this.client.execute(request)
+
+          results = [...messages.body.results, ...results]
```
tdeekens: I fail to understand what gets combined here? The nested but flattened response of a `Promise.all` before and what does the `messages` request return?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+
+      this.client
+        .execute(request)
+        .then(async payload => {
+          if (payload.statusCode !== 200) return Promise.reject(payload)
+          GDPRTool.delete(payload, uri.builder)
+
+          return Promise.resolve()
+        })
+        .catch(error => Promise.reject(error))
```
tdeekens: Why do you need to catch and reject the same error again? Can't we just omit the catch block in this case?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+      ],
+    })
+
+    this.logger = {
+      ...silentLogger,
+      ...options.logger,
+    }
+  }
+
+  getData(customerId) {
```
tdeekens: `fetchDeletableResources`?

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+          const messages = await this.client.execute(request)
+
+          results = [...messages.body.results, ...results]
+        }
+        return Promise.resolve(results)
+      })
+      .catch(error => Promise.reject(error))
+  }
+
+  deleteData(customerId) {
```
tdeekens: I think it's easier if you go with `deleteAll` and `deleteOne` below.

packages/gdpr-tool/package.json
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "@commercetools/gdpr-tool",
```
emmenko: Nitpick: either `gdpr-tools` or  simply `gdpr`.

In case we want to make this a CLI, I would suggest naming this `gdpr-cli`

packages/gdpr-tool/src/main.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+  static buildDeleteRequests(result, builder) {
+    let deleteUri = builder
+      .byId(result.id)
+      .withVersion(result.version)
+      .build()
+
+    // todo add config option to URI builder
+    deleteUri += '&dataErasure=true'
+    return GDPRTool.buildRequest(deleteUri, 'DELETE')
```
tdeekens: Why is all of this static on the prototype and not `new`ed? wouldn't it be `new GDPRTool(config)` (or whatever we call it) and then `this.buildDeletionRequest` (if renamed)`?

packages/gdpr-tool/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "version": "1.0.0-development",
```
emmenko: Nitpick: I never saw this kind of version. Why not simply `1.0.0`?

packages/gdpr-tool/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "description": "Export and delete all data related to a single customer",
```
emmenko: Is that the only use case for this library? 🤔

packages/gdpr-tool/readme.md
```diff
@@ -0,0 +1,9 @@
+# gdpr-tool
+
+Export and delete all customer data from the commercetools platform.
+
+## Install
+
+```bash
+npm install @commercetools/gdpr-tool --global
```
emmenko: It's not a CLI, so it should not be installed globally

packages/gdpr-tool/src/main.js
daern91: Yup, added now! :-)

packages/gdpr-tool/src/main.js
daern91: Nice find. Added a check for `404`'s

packages/gdpr-tool/src/main.js
daern91: Good question. `messages` returns the response from this
https://docs.commercetools.com/http-api-projects-messages.html

We build the predicate to all messages [here](https://github.com/commercetools/nodejs/blob/45d0f7902d0cd43c8e92d281eba81e22b56ccf88/packages/gdpr-tool/src/main.js#L89) since they can be linked to not only the `customerID`, but also the `orderId`, `cartId` etc etc.

So we then combine all the results(customer,cart,orders, etc) with all messages.

packages/gdpr-tool/src/main.js
daern91: Ahh thanks a lot, I didnt think of that 👍

packages/gdpr-tool/src/main.js
daern91: Hmm, I was thinking it would be smarter to have them all on the prototype instead of duplicating(?) them all onto the instance. 

I'm not sure of the convention here so is happy to change it. Do you know how its usually structured between the two?

packages/gdpr-tool/package.json
daern91: Yup.. Maybe we could even fit this inside of another package in the monorepo?

packages/gdpr-tool/package.json
daern91: Good idea. Will wrap it into CLI 👍

packages/gdpr-tool/package.json
daern91: By using `1.0.0-development` it will be bumped to `1.0.0` during npm release. No matter if we choose `major`, `minor` or `patch`. 

Im not 100% sure but I think if I named it `1.0.0` here it would be bumped to `1.0.1` if I release as `patch`, for example.

packages/customer-erasure/test/main.spec.js
```diff
@@ -168,5 +188,36 @@ describe('CustomerErasure', () => {
...
+    describe('with status code 500', () => {
+      beforeEach(() => {
+        const payload = {
+          statusCode: 500,
+          body: {
+            results: [],
+          },
+        }
+        customerErasure.client.process = jest.fn(async (request, callback) => {
+          await callback(payload)
```
tdeekens: Lovely. Just bizarre to see an callback be awaited somehow.

daern91:
Thanks a lot @tdeekens, @butenkor, @adnasa and @emmenko. Tobi guessed them both right on what I wanted feedback on... 

PR has now been updated with CLI and docs. Only thing missing is flow types but as I’m on vacation I suggest we do that in another PR next week. Just to get this out before the 25th.

Could you please have another look, and possibly approve, now?

emmenko:
image

😄

docs/cli/customer-erasure.md
```diff
@@ -0,0 +1,93 @@
+# Customer Erasure
+
+A package which deletes or exports [commercetools customer data](#list-of-customer-data) in `JSON` format from the [commercetools platform](https://docs.commercetools.com/).
```
emmenko: This is a bit weird: you can either delete or export (so no deletion) customer data. Why is the package called "customer erasure"?

integration-tests/cli/customer-erasure.it.js
```diff
@@ -0,0 +1,153 @@
...
+  review,
+  customLineItem,
+} from './helpers/customer-erasure.data'
+
+let projectKey
+if (process.env.CI === 'true') projectKey = 'custom-objects-import-int-tests'
+else projectKey = process.env.npm_config_projectkey
+
+describe('customer erasure', () => {
+  const bin = './integration-tests/node_modules/.bin/customer-erasure'
```
emmenko: Nitpick: `yarn run customer-erause` or `$(npm bin)/customer-erasure`

integration-tests/cli/helpers/utils.js
```diff
@@ -40,7 +40,7 @@ export function clearData(apiConfig, entityName) {
   })
 }
 
-export function createData(apiConfig, entityName, data) {
+export function createData(apiConfig, entityName, data, id) {
```
emmenko: Nitpick: I suggest to use named arguments instead of positional arguments. Makes it easier to add/remove arguments.

```js
createData({ apiConfig, entityName, data, id })
```

packages/customer-erasure/bin/customer-erasure.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+/* eslint-disable */
+/*
+  This file exists because we use Babel to transpile the JS but when testing
```
emmenko: General question: do we still need babel?

packages/customer-erasure/src/cli.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+const errorHandler = errors => {
+  if (Array.isArray(errors)) errors.forEach(logError)
+  else logError(errors)
+
+  process.exitCode = 1
+}
+
+const resolveCredentials = _args => {
```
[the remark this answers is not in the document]
emmenko: Yeah, having done some Reason recently, using `_` as a prefix means that the variable should be _ignored_.

packages/customer-erasure/src/cli.js
```diff
@@ -0,0 +1,148 @@
...
+    process.stderr.write(`ERR: ${errorFormatter.render(error)}`)
+  else process.stderr.write(`ERR: ${error.message || error}`)
+}
+
+const errorHandler = errors => {
+  if (Array.isArray(errors)) errors.forEach(logError)
+  else logError(errors)
+
+  process.exitCode = 1
+}
```
emmenko: General question: what happens if there is an error in the middle of the process? For example, if I get an network error and I retry the command, would it still complete the "deletion" or would there be problems because of the interrupted process?

packages/customer-erasure/src/main.js
```diff
@@ -0,0 +1,217 @@
...
+import {
+  createAuthMiddlewareForClientCredentialsFlow,
+  createAuthMiddlewareWithExistingToken,
+} from '@commercetools/sdk-middleware-auth'
+import { createUserAgentMiddleware } from '@commercetools/sdk-middleware-user-agent'
+import flatten from 'lodash.flatten'
+import silentLogger from './utils/silent-logger'
+import pkg from '../package.json'
+
+// todo add flow types
```
emmenko: A bit sad to see this as a "nice-to-have" thing and not a core part of the development.

packages/customer-erasure/src/main.js
```diff
@@ -0,0 +1,217 @@
...
+    const urisOfResourcesToDelete = [
+      customersUri,
+      ordersUri,
+      cartsUri,
+      paymentsUri,
+      shoppingsListsUri,
+      reviewsUri,
+    ]
+
+    return Promise.all(
```
emmenko: Nitpick:

```js
const allData = await Promise.all(...);
```

packages/customer-erasure/src/main.js
```diff
@@ -0,0 +1,217 @@
...
+          return Promise.resolve()
+        })
+      })
+    )
+  }
+
+  _deleteOne(payload, builder) {
+    const results = payload.body.results
+    if (results.length > 0) {
+      results.forEach(async result => {
```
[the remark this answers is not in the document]
emmenko: The `this.client.execute(deleteRequest)` is a promise, so that part should be awaited.

Actually it should probably be a `Promise.all(results.map(...))`

integration-tests/cli/helpers/utils.js
tdeekens: If this is used across the library I don't see it being an issue for this PR.

packages/customer-erasure/bin/customer-erasure.js
tdeekens: I think babel is used for all/most CLIs? Maybe something to remove at some point in another step depending on the engine/node version we require?

integration-tests/cli/helpers/utils.js
emmenko: Of course, it's more of a general thing (hence the nitpick)

docs/cli/customer-erasure.md
tdeekens: Not sure what other names would be good. I just tried to avoid GDPR. What about `Customer Aggregator` or `Customer Collator`. I am find with `Customer Erasure ` as I think it's the package's main purpose and CLI tools are inherently hard to name. The fact that data needs to be gathered in order to be deleted and deletion is an option (as in it can only be exported) seems obvious.

daern91: Ahh, well said, totally agree.

emmenko: Hmm why not simply `customer-data-exporter`? And you have the option to delete it

packages/customer-erasure/src/cli.js
```diff
@@ -0,0 +1,148 @@
...
+  })
+  .option('accessToken', {
+    describe: 'CTP client access token.',
+  })
+  .option('projectKey', {
+    alias: 'p',
+    describe: 'API project key.',
+    demand: true,
+  })
+  .option('deleteAll', {
```
emmenko: Given this option, we should prompt the user for confirmation, as it’s a non-reversible action! ⚠️

packages/customer-erasure/src/cli.js
daern91: That should work just fine as we let `404`s resolve and only call delete if data exists [here](https://github.com/commercetools/nodejs/pull/601/files#diff-39c284b79054f260b4baecab4f766173R179)

packages/customer-erasure/src/cli.js
daern91: Ahh, great idea!

docs/cli/customer-erasure.md
daern91: I don't have a super strong opinion on this but I think using `Erasure` should be fine as it is the main feature clients have been asking for and why we decided to create this package in the first place. 

Even though the deletion part is, indeed, an option in the CLI just to make it a bit safer.

integration-tests/cli/customer-erasure.it.js
daern91: Above didnt work out for me when running locally so I kept the path like this for now.

integration-tests/cli/helpers/utils.js
daern91: Good idea. Will keep in mind for future PRs :-)

packages/customer-erasure/src/cli.js
daern91: yargs does not have native prompt so I added the functionality by using [prompt-confirm](https://www.npmjs.com/package/prompt-confirm)

docs/cli/customer-erasure.md
cneijenhuis: I guess `Customer` isn't a perfect match. It is about personal data, which is partially stored in the `Customer` object, but also in `Order` etc.
I'd prefer `Personal Data Foobar` over `Customer Foobar`.

I also agree we shouldn't hardcode GDPR into the name. There may be other countries outside the EU/GDPR that have similar rules.

The law-ish terms are `Retrieval` and `Erasure`, so if we want someone to be able to pattern-match from GDPR to this tool, something like `Personal Data Retrieval and Erasure` would be a good name.

But I don't have a strong opinion, whatever fits well among other CLI tools is also fine with me :)

tdeekens: `personal-data-erasure` would be fine by me. I still think erasure is the main purpose of this and the retrieval being a side effect. If someone searches for "how can I delete my private stuff" I don't think `personal-data-exporter` is an obvious name for it. I don't like the term `personal-data-management` as everything is management in software, as is data. 

Long discussions around naming again:

1. Either keep as is
2. Rename to (unless other suggestions): `personal-data-erasure`

Feel free to suggest other names.

emmenko: Yeah, better like this 👌

tdeekens:
Alright. Just for all @daern91 will be back in the office Tuesday. It's not needed to get this done before. I clarified that. 

What should be done then, to my understanding, is only renaming the pkg (again sorry) to `personal-data-erasure`.

packages/customer-erasure/src/cli.js
emmenko: I usually use `prompts` but, yeah I guess it's fine :)

packages/customer-erasure/src/cli.js
emmenko: I'm just asking to avoid having dead references.

And also to make sure that the script is idempotent, meaning that it can be executed multiple times.

emmenko:
I would also like to see flowtype annotations 😇

tdeekens:
> I would also like to see flowtype annotations

Yes, but that @daern91 can tackle as a follow-up as mentioned somewhere.

daern91:
Absolutely fantastic, thanks a lot for the help everyone!!

PR #602, opened 2018-05-22
feat(sync-actions): Add `SyncActionConfig` to all types

adnasa:
#### Summary

- follow up to https://github.com/commercetools/nodejs/pull/592
- adds `SyncActionConfig` to all sync-action types.

* Tests
  * [x] Unit

I didn't add test to all types just yet. I will do that in this PR as we speak, but it shouldn't stop this PR from being reviewed


packages/sync-actions/src/customers.js
```diff
@@ -9,38 +14,50 @@ import * as diffpatcher from './utils/diffpatcher'
...
-  return function doMapActions(diff, newObj, oldObj /* , options */) {
+function createCustomerMapActions(
+  mapActionGroup: Function,
+  config: SyncActionConfig
+): (diff: Object, newObj: Object, oldObj: Object) => Array<UpdateAction> {
+  return function doMapActions(
+    diff: Object,
+    newObj: Object,
+    oldObj: Object /* , options */
+  ): Array<UpdateAction> {
```
adnasa: I did this re-factoring, because my editor was screaming with warnings. I figured why not just do it as we go

adnasa:
@dferber90 @tdeekens any chance I get a review here?
thanks in advance 😛

packages/sync-actions/src/cart-discounts-actions.js
```diff
@@ -14,11 +14,12 @@ export const baseActionsList = [
   { action: 'changeStackingMode', key: 'stackingMode' },
 ]
 
-export function actionsMapBase(diff, oldObj, newObj) {
+export function actionsMapBase(diff, oldObj, newObj, config = {}) {
```
tdeekens: Why not `config: SyncActionConfig = {}`?

packages/sync-actions/src/tax-categories.js
```diff
@@ -1,34 +1,49 @@
...
       )
     )
-
     return flatten(allActions)
   }
 }
 
-export default (config: Array<ActionGroup>): SyncAction => {
+export default (
+  actionGroupConfig: Array<ActionGroup>,
```
tdeekens: Ok this was before but that's confusing: the actionGroupConfig is an list of actionGroup.

packages/sync-actions/src/tax-categories.js
```diff
@@ -1,34 +1,49 @@
...
+  config: SyncActionConfig
```
tdeekens: Maybe rename to `syncActionConfig` it's a bit confusing when you read on and see the createTaxCategoriesMapAction receive "just config".

packages/sync-actions/src/tax-categories.js
adnasa: I would have used `actionsGroups`, but that was taken 🚐

packages/sync-actions/src/cart-discounts-actions.js
adnasa: `flow` notation hasn't been added to this file at the time of writing.
The changes I've made where I added type notations where on files where `flow` was already added, and just needed "cleaning up" to eliminate all warnings.

I would want to say that we should add `flow` to all files but I think we should do that another time

tdeekens: All cool. Thanks for clarifying.

adnasa:
just one last commit 3e64943 where I `config -> syncActionConfig`

packages/sync-actions/src/tax-categories.js
dferber90: I'd welcome this renaming 👍

adnasa:
@dferber90 done, 46a1d0a
will hit that button when everything goes green! ✅

PR #612, opened 2018-05-30
feat(api-req-builder): add \`fullDataErasure\` option

### LINKED ISSUE #611: add `fullDataErasure()` to api-request-builder
### Description

We now have the [dataErasure option](https://deploy-preview-963--commercetools-documentation.netlify.com/release-notes.html#releases-2018-05-24-data-erasure) in the API and we need to add it to the `api-request-builder` in order to not have to write code like [this](https://github.com/commercetools/nodejs/blob/master/packages/personal-data-erasure/src/main.js#L211)

```
let deleteUri = builder
  .byId(result.id)
  .withVersion(result.version)
  .build()

deleteUri += '&dataErasure=true'
```

### Expected Behavior

We should have a function like `fullDataErasure(true)` that adds this option to the uri.

```
let deleteUri = builder
  .byId(result.id)
  .withVersion(result.version)
  .fullDataErasure(true)
  .build()
```

daern91:

### Description

We now have the [dataErasure option](https://deploy-preview-963--commercetools-documentation.netlify.com/release-notes.html#releases-2018-05-24-data-erasure) in the API and we need to add it to the `api-request-builder` in order to not have to write code like [this](https://github.com/commercetools/nodejs/blob/master/packages/personal-data-erasure/src/main.js#L211)

```
let deleteUri = builder
  .byId(result.id)
  .withVersion(result.version)
  .build()

deleteUri += '&dataErasure=true'
```

PR adds a `fullDataErasure()` function that adds this option to the uri.

```
let deleteUri = builder
  .byId(result.id)
  .withVersion(result.version)
  .fullDataErasure(true)
  .build()
```

#### Todo

* Tests
  * [x] Unit
* [x] Documentation
  
closes #611

### DIFF
--- a/packages/api-request-builder/src/build-query-string.js
+++ b/packages/api-request-builder/src/build-query-string.js
@@ -106,7 +106,7 @@ export default function buildQueryString(
 
   if (version) queryString.push(`version=${version}`)
 
-  if (dataErasure) queryString.push(`dataErasure=${String(dataErasure)}`)
+  if (dataErasure) queryString.push(dataErasure)
 
   return queryString.join('&')
 }
--- a/docs/sdk/api/apiRequestBuilder.md
+++ b/docs/sdk/api/apiRequestBuilder.md
@@ -75,14 +75,14 @@ const uri = service.channels.withVersion(2).build()
 
 You can also append the `dataErasure` option to the uri when making a delete request if you want to make sure all related data is deleted. For example, regarding the GDPR, this means that all personal data related to the particular object, including invisible data, is erased. [More info here](https://docs.commercetools.com/release-notes#releases-2018-05-24-data-erasure)
 
-This can be done by passing the required `boolean` to the `.fullDataErasure()` method.
+This can be done by using the `.withFullDataErasure()` method.
 
 ```js
 const service = createRequestBuilder(options)
 const deleteUri = service.payments
   .byId(12345)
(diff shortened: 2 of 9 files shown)


packages/api-request-builder/src/build-query-string.js
```diff
@@ -105,5 +106,7 @@ export default function buildQueryString(
 
   if (version) queryString.push(`version=${version}`)
 
+  if (dataErasure) queryString.push(`dataErasure=${String(dataErasure)}`)
```
daern91: Using `String()` since linter yells about not coercing booleans otherwise.

`Cannot coerce `dataErasure` to string because boolean [1] should not be coerced.`

tdeekens: As a side note: magic boolean arguments are kind of discouraged. 

1. From a call site you never know what it does

Instead of `withErasure(true)` it would be `withEresure({ shouldErase: true })`. 

That obviously makes the API more clunky and shows a slight flaw. Maybe we can omit the argument all together. And hence also the check `if(shouldErease)`?

tdeekens:
`withFullDataErasure` is also fine. I was just more trying to follow the `with` naming pattern. As discussed the Boolean arg can be omitted.

daern91:
@tdeekens Great ideas. Refactored now!

PR #614, opened 2018-05-31
feat(personal-data-erasure): add flow types and refactoring

### LINKED ISSUE #613: Update `personal-data-erasure` package
Add flow types to `personal-data-erasure` package and use new `.withFullDataErasure()` method.

daern91:
#### Summary

As promised in earlier PR. Here is the added flow types and also changing `deleteUri` to use new `.withFullDataErasure()` method instead of adding it like this `deleteUri += '&dataErasure=true'`

closes #613


packages/personal-data-erasure/src/main.js
```diff
@@ -106,92 +134,111 @@ export default class PersonalDataErasure {
...
-        return Promise.reject(
-          Error(`Request returned status code ${response.statusCode}`)
-        )
+  async _getAllMessages(request: ClientRequest): Promise<Messages> {
+    const messages = await this.client.process(
+      request,
+      (response: ClientResult): Promise<any> => {
+        if (response.statusCode !== 200 && response.statusCode !== 404)
+          return Promise.reject(
+            Error(`Request returned status code ${response.statusCode}`)
```
adnasa: hmm.. I could've caught this earlier.
Doesn't the API return a `message` as well? doesn't it make sense to display that together with the `statusCode`

types/personalDataErasure.js
```diff
@@ -0,0 +1,199 @@
+/* @flow */
+
+/* Logger */
+export type LoggerOptions = {
+  error: Function,
```
emmenko: Nitpick: Instead of typing this as `Function`, it’s better to specify the full signature with the return value of `unit`.

types/personalDataErasure.js
```diff
@@ -0,0 +1,199 @@
...
+  id: string,
+  version: number,
+  sequenceNumber: number,
+  resource: Object,
+  resourceVersion: number,
+  type: string,
+  customer?: Customer,
+  createdAt: string,
+  lastModifiedAt: string,
+}>
```
emmenko: Might be better to extract this into its own type `Message`

types/sdk.js
```diff
@@ -292,6 +292,22 @@ export type ServiceBuilderDefinition = {
...
+  page: Function,
+  perPage: Function,
+  byId: Function,
+  byKey: Function,
+  byCustomerId: Function,
+  byCartId: Function,
+  expand: Function,
+  build: Function,
+  parse: Function,
+}
```
emmenko: Same thing here with the function signatures

packages/personal-data-erasure/src/main.js
```diff
@@ -106,92 +134,111 @@ export default class PersonalDataErasure {
...
+      resourcesToDelete.map(
+        (resource: {
+          uri: string,
+          builder: ServiceBuilderInstance,
+        }): Promise<any> => {
+          const request = PersonalDataErasure.buildRequest(resource.uri, 'GET')
+
+          return this.client.process(
+            request,
+            (response: Object): Promise<any> => {
```
emmenko: Isn’t this `Promise<AllData>`?

packages/personal-data-erasure/src/main.js
```diff
@@ -201,22 +248,26 @@ export default class PersonalDataErasure {
     }
   }
 
-  static buildDeleteRequests(result, builder) {
-    let deleteUri = builder
+  static buildDeleteRequest(
+    result: Object,
```
emmenko: We should expand this type

packages/personal-data-erasure/src/main.js
daern91: Im afraid that doesn't work as we resolve the promise without data like this.
`return Promise.resolve()`

Setting it to `Promise<AllData>` then returns this error instead

```js
Cannot return `Promise.resolve()` because undefined [1] is incompatible with `AllData` [2] in type argument `R` [3].

   packages/personal-data-erasure/src/main.js:227:22
   227|               return Promise.resolve()
                             ^^^^^^^^^^^^^^^^^ [1]

References:
   packages/personal-data-erasure/src/main.js:220:41
   220|             (response: Object): Promise<AllData> => {
                                                ^^^^^^^ [2]
   /private/tmp/flow/flowlib_3e57fb1b/core.js:605:24
   605| declare class Promise<+R> {
                               ^ [3]
```

emmenko: Then it should just be `Promise<void>` 😉

packages/personal-data-erasure/src/main.js
emmenko: Hmm well, I see that in the `_deleteOne` function the `buildDeleteRequest` is called with `result` being typed as `ClientResult`.

PR #641, opened 2018-06-22
chore(scripts): run flow through eslint to get warnings

daern91:
#### Summary

Updates dependencies and runs Flow through ESLint to get the warnings.


emmenko:
I'm confused now 🙃

tdeekens:
We are too the description didn't elaborate enough.

1. Flow runs through jest-runner-flowtype
2. The `flowconfig` contains `include_warnings` to be true

- When running flow through either the jest runner or the `flow-bin` warnings are not shown
- When running flow through ESLint however the warnings are shown

It was meant to be a task for @Babanila to fix as many warnings as possible for that they need to see them at least somewhere. @daern91 say this as an option.

This is also why I wrote that we should figure out if the behaviour we observed above is normal.

emmenko:
Hmm I see. But it's not that we are running flow now, we are just running lint with the flow plugin, right? Does this mean that we don't directly use flow anymore?
Is this also a "known problem"?

tdeekens:
> Does this mean that we don't directly use flow anymore

1. We can and we do on CI [ref](https://github.com/commercetools/nodejs/blob/master/package.json#L17)
2. I think "nobody" used `npm run flow` so far. ESLint will also report flow errors and warnings just as flow-bin to my understanding

emmenko:
Good, that for the info! 👌

PR #647, opened 2018-06-25
perf(deps): remove bluebird dependency

daern91:
#### Summary

Bluebird is a [quite big](https://bundlephobia.com/result?p=bluebird@3.5.1) library and its not really needed in most of our packages where native Promises work just as good. PR removes this dependency from `discount-code-importer`, `state-importer`, `csv-parser-price` and `price-exporter` packages.

For example, this change reduces `state-importer` dependency size from **211.1** to **133.9** kB (minified) (~37% decrease!)

#### Todo

* Tests
  * [x] Unit
  * [x] Integration


packages/discount-code-importer/src/main.js
```diff
@@ -97,6 +96,14 @@ export default class DiscountCodeImport {
     return `code in ("${predicateArray.join('", "')}")`
   }
 
+  static promiseMapSerially(functions: Array<Function>) {
```
daern91: Made this one to replace bluebird's [mapSeries](http://bluebirdjs.com/docs/api/promise.mapseries.html) in order to execute every promise sequentially.

packages/discount-code-importer/src/main.js
```diff
@@ -139,58 +147,60 @@ export default class DiscountCodeImport {
...
   _createOrUpdate(newCodes: CodeDataArray, existingCodes: CodeDataArray) {
-    return Promise.map(newCodes, (newCode: CodeData) => {
-      const existingCode = _.find(existingCodes, ['code', newCode.code])
-      if (existingCode)
-        return this._update(newCode, existingCode)
-          .then(response => {
-            if (response && response.statusCode === 304)
-              this._summary.unchanged += 1
-            else this._summary.updated += 1
+    return Promise.all(
```
daern91: In all these cases we will simply use native `Promise.all` instead of bluebird's [Promise.map](http://bluebirdjs.com/docs/api/promise.map.html). No other changes just a lot of noise from prettier since we pass in the mapped array to `Promise.all`

tdeekens: Are we sure that the behaviour for `Bluebird.map` and `Promise.all` is the same. Meaning that both behave the same whenever one Promise in the list is rejected that:

1. The over all rejection is the same
2. The rest of the list is still executed _or_ aborted?

daern91: Sure thing, please have a look at [this CodeSandbox](https://codesandbox.io/s/km98xpq3yr). Just change the script src in index.html to see it behaving the same.

PR #654, opened 2018-07-05
fix(req-builder): ensure that built uri is always the same and defaul…

### LINKED ISSUE #653: API request builder defaulting to different staged values
### Description
When using the [api-request-builder](https://commercetools.github.io/nodejs/sdk/api/apiRequestBuilder.html) to query [productprojections](https://docs.commercetools.com/http-api-projects-productProjections.html#query-productprojections) we set the default `staged` value to **false** when running once (by not adding it at all), and then to **true** on all subsequent calls to the `.build()` method. 

This doesn't seem like the intended behaviour since we get different results. So, my question is, which one do we actually want to default to? 

Personally, I think defaulting to the first behaviour is the most coherent since it will just follow the API. This way people can opt-in by using the `.staged(true)` method and we won't have to update all of our tests.

Please have a look at the current behaviour below

```js
const a = require("@commercetools/api-request-builder");

const requestBuilder = a.createRequestBuilder({
  projectKey: "test"
});
const productProjectionsService = requestBuilder.productProjections;

console.log(productProjectionsService.build());
console.log(productProjectionsService.build());
console.log(productProjectionsService.build());
// /test/product-projections
// /test/product-projections?staged=true
// /test/product-projections?staged=true
```

Ping @emmenko @adnasa @tdeekens

daern91:
…t staged to false to align with API

#### Description

PR fixes bug where multiple calls to the `.build()` method would return different uri's. 
(first call would default `staged` to **false** and all following ones to **true**)
More info [here](https://github.com/commercetools/nodejs/issues/653).

This will also remove the default addition of `staged=true` to created URIs so we instead align completely with the API. ([which will default to false](https://docs.commercetools.com/http-api-projects-products-search#search-productprojections))

After discussion with relevant teams, both in PS and MC, there seemed to be an agreement that this is the best way forward. 

However, input from PO would also be good. Especially as this could be breaking. @Oehmi?

For more info, please read discussion here #653 

cc @butenkor @adnasa @tdeekens 

resolves #653


packages/api-request-builder/src/default-params.js
```diff
@@ -77,16 +76,13 @@ export function setDefaultParams() {
...
   if (this.features.includes(features.search)) {
-    this.params.staged = getDefaultSearchParams().staged
     this.params.pagination = getDefaultSearchParams().pagination
     this.params.search = getDefaultSearchParams().search
   }
 
   if (this.features.includes(features.queryLocation))
     this.params.location = getDefaultQueryParams().location
 
-  if (this.features.includes(features.projection)) this.params.staged = true
```
tdeekens: Just so I follow:

1. Before `staged` was `true` by default
2. Then we check if we search to set it to `getDefaultSearchParams().staged`
   - Which is also `true` as per [ref](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-params.js#L44)
3. Then we check if we're looking at the projection and set it to `true`

I am just wondering if there is a deliberate product need buried here. From old days.

packages/api-request-builder/src/query-projection.js
```diff
@@ -6,7 +6,7 @@
  * (for current / published)
  * @return {Object} The instance of the service, can be chained.
  */
-export function staged(isStaged: boolean = true): Object {
+export function staged(isStaged: boolean = false): Object {
```
tdeekens: While we're here. Maybe `shouldFetchStaged` might be a better name? _Note_: of the parameters not the `param`.

packages/api-request-builder/src/default-params.js
daern91: Yeah this is how I perceived the workflow too. No idea about any possible old need or not though.

@Oehmi ?

adnasa:
actually @yacx should be the one to give this a go, really shame that he is on leave and we take this decision on a product level. @Oehmi sorry for throwing this on your lap.

I propose the following strategy, especially if we get a ✅ from a PO

1. go ahead with this
2. Discuss this problem openly in detail, and what consequences this changes could lead to, in the same fashion we did for [`sync-actions`](https://github.com/commercetools/nodejs/blob/master/docs/sdk/api/syncActions.md#L569). Either a blog post, or a new section in `req-builder` documentation would be good.
3. add reference to the discussion in the release notes in addition to marking it as __breaking change__

daern91:
Sounds good to me, added documentation [here](https://github.com/commercetools/nodejs/pull/654/commits/9cddad43cf0dddecc6c21e2ba9fddedf204dea1e) and will link to this PR and issue discussion in release notes. 

Anything else you can think of before we can merge this one, @adnasa, @butenkor, @tdeekens?

adnasa:
> ..and will link to this PR and issue discussion in release notes.

I believe linking to the original issue/disucussion #653 would be move valuable. Other than that, I'm ready to give this a __go__

PR #655, opened 2018-07-06
MCD-1236: adding return info actions to order's sync actions

NickDevG:
#### Summary
Adding return info actions to order's sync actions

#### Description
[Jira ticket](https://jira.commercetools.com/browse/MCD-1236)


packages/sync-actions/src/order-actions.js
```diff
@@ -37,3 +38,56 @@ export function actionsMapDeliveries(diff, oldObj, newObj) {
...
+
+export function actionsMapReturnsInfo(diff, oldObj, newObj) {
+  const returnInfoDiff = diff.returnInfo
+  if (!returnInfoDiff) return []
+
+  const handler = createBuildArrayActions('returnInfo', {
+    [CHANGE_ACTIONS]: (oldSReturnInfo, newReturnInfo) => {
+      const updateActions = Object.keys(diff.returnInfo).reduce(
+        (itemActions, key) => {
+          const { items = {} } = diff.returnInfo[key]
```
daern91: Hmm, why not use `returnInfoDiff` here and on line 48?

packages/sync-actions/src/order-actions.js
```diff
@@ -37,3 +38,56 @@ export function actionsMapDeliveries(diff, oldObj, newObj) {
...
+          if (Object.keys(items).length > 0) {
+            return [
+              ...itemActions,
+              ...Object.keys(items).reduce((actions, index) => {
+                if (items[index].shipmentState) {
```
daern91: Im not sure if this is a legit case, however, this will fail if we want to generate both an `setReturnShipmentState` and `setReturnPaymentState`. For example,

```js
const before = {
  returnInfo: [
    {
      returnTrackingId: "touched-item",
      items: [
        {
          id: "test-1",
          shipmentState: "returned",
          paymentState: "initial"
        }
      ]
    }
  ]
}

const now = {
  returnInfo: [
    {
      returnTrackingId: "touched-item",
      items: [
        {
          id: "test-1",
          shipmentState: "backInStock",
          paymentState: "refunded"
        }
      ]
    }
  ]
}

// will always only return the `setReturnShipmentState`
[
  {
    action: "setReturnShipmentState",
    returnItemId: "test-1",
    shipmentState: "backInStock"
  }
]

```

NickDevG: Nice catch. It is possible to modify both. Thanks for pointing that out 💪

NickDevG:
@daern91 Sorry for the late response. I was so focused on what I was doing that I didn't realized you've commented on the PR

packages/sync-actions/src/order-actions.js
```diff
@@ -37,3 +38,56 @@ export function actionsMapDeliveries(diff, oldObj, newObj) {
...
+
+export function actionsMapReturnsInfo(diff, oldObj, newObj) {
+  const returnInfoDiff = diff.returnInfo
+  if (!returnInfoDiff) return []
+
+  const handler = createBuildArrayActions('returnInfo', {
+    [CHANGE_ACTIONS]: (oldSReturnInfo, newReturnInfo) => {
+      const updateActions = Object.keys(returnInfoDiff).reduce(
+        (itemActions, key) => {
+          const { items = {} } = diff.returnInfo[key]
```
daern91: nitpick: I think we could also use `returnInfoDiff` variable here, so
`const { items = {} } = returnInfoDiff[key]`

packages/sync-actions/test/order-sync.spec.js
```diff
@@ -21,64 +21,321 @@ describe('Actions', () => {
...
+        shippingInfo: {
+          deliveries: [],
+        },
+      }
+      const now = {
+        shippingInfo: {
+          deliveries: [
+            {
+              items: [{ id: 'li-1', qty: 1 }, { id: 'li-2', qty: 2 }],
+              parcels: [
```
adnasa: `deliveries`, `parcels` and `address` are all optional for the `addDelivery` action, and I would suggest that the test should reflect that, by breaking the test(s) into testing `parcels`, `deliveries` and `address` separately.


ref: https://docs.commercetools.com/http-api-projects-orders.html#add-delivery

packages/sync-actions/test/order-sync.spec.js
```diff
@@ -21,64 +21,321 @@ describe('Actions', () => {
...
+            items: [
+              {
+                id: 'test-1',
+                shipmentState: 'returned',
+                paymentState: 'initial',
+              },
+              {
+                id: 'test-2',
+                shipmentState: 'returned',
+                paymentState: 'refunded',
```
adnasa: I would add a comment that this is changed, and that this is what you are actually testing... otherwise it is hard to detect/predict what the `actions` should look like

packages/sync-actions/src/order-actions.js
```diff
@@ -37,3 +38,56 @@ export function actionsMapDeliveries(diff, oldObj, newObj) {
...
+
+  const handler = createBuildArrayActions('returnInfo', {
+    [CHANGE_ACTIONS]: (oldSReturnInfo, newReturnInfo) => {
+      const updateActions = Object.keys(returnInfoDiff).reduce(
+        (itemActions, key) => {
+          const { items = {} } = returnInfoDiff[key]
+          if (Object.keys(items).length > 0) {
+            return [
+              ...itemActions,
+              ...Object.keys(items).reduce((actions, index) => {
```
adnasa: wouldn't it be nicer if this was `Object.values` alternatively `Object.entries`? then a `items[index]...` wouldn't be needed

packages/sync-actions/src/order-actions.js
```diff
@@ -37,3 +38,56 @@ export function actionsMapDeliveries(diff, oldObj, newObj) {
...
+                let itActions = []
```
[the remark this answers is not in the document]
NickDevG: We can have both, payment and shipment changes at the same time, so we can't return inside every block as @daern91 pointed out in [this comment](https://github.com/commercetools/nodejs/pull/655#discussion_r200927490)

packages/sync-actions/test/order-sync.spec.js
NickDevG: This is not mine, I think is marked as new because of auto-formatting. Do you want me to look into it?

packages/sync-actions/src/order-actions.js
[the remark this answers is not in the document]
adnasa: interesting... then let's change approach...
the way you have your if-blocks now are sensitive to order. 
if you use push, then this can easily be re-factored later

```js

const x = []
if (shipment) x.push({})
if (payment) x.push({})

// instead of 

let x = []
if (shipment) x = [{}]
if (payment) x = [...x, {}]
```

packages/sync-actions/test/order-sync.spec.js
adnasa: aaaah... I seee.
nevermind then :)

packages/sync-actions/src/order-actions.js
NickDevG: I'm using the key to get the updated prop from the `newReturnInfo`, while I'm iterating over the `returnInfoDiff` item's key

NickDevG:
@adnasa when you have some time, please take another look 🙏

NickDevG:
@daern91 @adnasa can any of you please merge this? I don't have write access

adnasa:
will do

PR #670, opened 2018-07-19
chore(docs): add 'todo' cmd and auto generate todo docs

daern91:
affects: @commercetools/custom-objects-importer

#### Description

After reading the todo discussion I thought it would be a good idea to also use `leasot` in the entire monorepo. 

This PR proposes not only the `todo` command to easily list all `TODO`s but also auto-generates a `.md` file that we can link to. The idea is that it can be good so the `TODO`s don't get forgotten and also as an easy way for new contributors to find easier tasks.

Any opinions on this, @tdeekens @adnasa @Babanila?

closes #671


package.json
```diff
@@ -33,6 +33,8 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && ./scripts/publish-docs",
     "flow": "eslint packages",
+    "todo": "leasot 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib'",
```
tdeekens: I am confused by the integration. Don't you want to do this in lint-staged on any js file change and just auto git add things?

package.json
```diff
@@ -33,6 +33,8 @@
...
+    "todo-staged": "leasot -x --reporter markdown 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib' > TODO.md && git add TODO.md",
```
tdeekens: This is a strange command. I think this should just goto the lint-staged.config.js

package.json
daern91: Ahh, good question. I want to run through the whole repo in order to build a correct `TODO.md`. So running it on any js file change inside `lint-staged.config.js` would run the script multiple times, no?

tdeekens: Nope. It runs it given that a given set of currently staged files matching you glob. Ref: https://github.com/okonet/lint-staged#packagejson-example

tdeekens: If you want to run it against anything then a normal pre-commit hook would do. Husky is very nice for that. https://github.com/typicode/husky

tdeekens: I think generally 

1. You need to run it only once for all files
2. Then have a hook for js files with lint-staged

daern91: Im not following what the second step would do here?

Anyways, I shortened the scripts so they make more sense here https://github.com/commercetools/nodejs/pull/670/commits/bd2ad8a9190b16f55720720c42f7429596f78a35

So basically, `npm run todo` will just log out all errors to the terminal.
`todo-staged` is being run in the Husky `precommit` hook after `lint-staged` in order to create the `.md` with all the todo's we currently have.

package.json
```diff
@@ -33,8 +33,8 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && ./scripts/publish-docs",
     "flow": "eslint packages",
-    "todo": "leasot 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib'",
-    "todo-staged": "leasot -x --reporter markdown 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib' > TODO.md && git add TODO.md",
+    "todo": "leasot 'packages/*/src/**.js'",
```
tdeekens: I'd prefix with and name `build:todos`

package.json
```diff
@@ -33,6 +33,8 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && ./scripts/publish-docs",
     "flow": "eslint packages",
+    "todo": "leasot 'packages/*/src/**.js'",
```
adnasa: wouldn't this be `todo:build` or `build:todo`?

otherwise the command is strange

package.json
```diff
@@ -33,8 +33,8 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && ./scripts/publish-docs",
     "flow": "eslint packages",
-    "todo": "leasot 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib'",
-    "todo-staged": "leasot -x --reporter markdown 'packages/**/*.js' --ignore '**/node_modules','**/dist','**/lib' > TODO.md && git add TODO.md",
+    "todo": "leasot 'packages/*/src/**.js'",
+    "todo-staged": "leasot -x --reporter markdown 'packages/*/src/**.js' > TODO.md && git add TODO.md",
```
tdeekens: Maybe `stage:todos` and rename the usage from above. Furthermore maybe it's a `scripts/stage-todos`.

package.json
tdeekens: Different thread: what's the difference between `todo` and `todo-staged` and why doesn't one use the other?

daern91: Its weird.

`todo` is meant to just be run in the terminal to quickly print out all the TODO's.
`leasot 'packages/*/src/**.js' -x`

If we extend that command in todo-staged so it becomes

`npm run todo -- --reporter markdown > TODO.md`

it actually includes all below in the piping to file:

> @ todo /Users/*****/*****/nodejs
> leasot 'packages/*/src/**.js' -x "--reporter" "markdown"

When running it by itself, without extending original `todo`, I dont see this problem.

package.json
```diff
@@ -33,8 +33,8 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && ./scripts/publish-docs",
     "flow": "eslint packages",
-    "todo": "leasot 'packages/*/src/**.js'",
-    "todo-staged": "leasot -x --reporter markdown 'packages/*/src/**.js' > TODO.md && git add TODO.md",
+    "build:todos": "leasot -x 'packages/*/src/**.js'",
+    "publish:todos": "leasot -x --reporter markdown 'packages/*/src/**.js' > TODOS.md",
```
tdeekens: Publish? We don't publish anything? We just generate them. 😕

daern91: hah, friggin naming... `build:todos-list`? `build:todos-file`? `docs:todos`?

tdeekens: generate:todos just dumps them to stdout
build:todos adds a file

It isn't docs really.

PR #671, opened 2018-07-19
Integrate Todo generation into lint-staged

tdeekens:
#### Summary

More in inline comments. It's a PR against your base.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -1,0 +1,9 @@
+> @ generate:todos /Users/tdeekens/Development/nodejs
+> leasot -x --reporter markdown 'packages/\*/src/\*\*.js'
+
+### TODOs
+
+| Filename                                           | line # | TODO                                                                                        |
+| :------------------------------------------------- | :----: | :------------------------------------------------------------------------------------------ |
+| packages/api-request-builder/src/create-service.js |  129   | this can lead to invalid URIs as getIdOrKey can return                                      |
+| packages/custom-objects-importer/src/main.js       |  121   | remove `FlowFixMe` when [this](https://github.com/facebook/flow/issues/5294) issue is fixed |
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -11,5 +11,7 @@ module.exports = {
     // For that reason, we move the `--onlyChanged` flag next to it.
     'npm run lint:js -- --reporters=jest-silent-reporter --onlyChanged',
     'flow focus-check',
+    'npm run build:todos',
+    'git add TODOS.md',
   ],
 }
--- a/package.json
+++ b/package.json
@@ -10,10 +10,12 @@
     "bootstrap": "lerna bootstrap",
     "check-updates": "./scripts/check-updates",
     "clean": "rm -rf coverage; lerna exec -- rimraf lib dist",
-    "precommit": "lint-staged && npm run todo-staged",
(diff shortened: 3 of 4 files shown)


TODOS.md
```diff
@@ -0,0 +1,9 @@
+> @ generate:todos /Users/tdeekens/Development/nodejs
```
tdeekens: Just renamed the file. It contains multiple todos so why not name it like that :)

lint-staged.config.js
```diff
@@ -11,5 +11,7 @@ module.exports = {
     // For that reason, we move the `--onlyChanged` flag next to it.
     'npm run lint:js -- --reporters=jest-silent-reporter --onlyChanged',
     'flow focus-check',
+    'npm run build:todos',
+    'git add TODOS.md',
```
tdeekens: Adding is something you, I think, want to do in a lint-staged script cause there you know if you actually have to.

package.json
```diff
@@ -10,10 +10,12 @@
     "bootstrap": "lerna bootstrap",
     "check-updates": "./scripts/check-updates",
     "clean": "rm -rf coverage; lerna exec -- rimraf lib dist",
-    "precommit": "lint-staged && npm run todo-staged",
+    "precommit": "lint-staged",
```
tdeekens: precommit is lint-staged. the rest should/could go there.

package.json
```diff
@@ -10,10 +10,12 @@
...
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "commit": "git-cz",
     "build": "lerna run --parallel build",
+    "generate:todos": "leasot -x --reporter markdown 'packages/*/src/**.js'",
```
tdeekens: This is your printing script. It's just dumps them to stdout.

package.json
```diff
@@ -10,10 +10,12 @@
...
+    "build:todos": "npm run generate:todos > TODOS.md",
```
tdeekens: Here we generate and dump it into the file.

TODOS.md
daern91: Wait isn’t this the problem I had too? Now we are checking in not so useful data in line #1 and #2

lint-staged.config.js
daern91: Ahh got it. I thought this was some sort of hook that got run on every file change, individually. But it actually just runs once?

package.json
daern91: Is this ok looking in terminal though? Sorry am on phone so can’t check. I mean since reporter is markdown instead of default table.

TODOS.md
daern91: ‘	+> @ generate:todos /Users/tdeekens/Development/nodejs
’

lint-staged.config.js
tdeekens: Yup. The staged files are passed as an argument but we dont do anything with them so this should be fine. Give it a try. Otherwise I misread the lint-staged docs.

package.json
tdeekens: Ah well yeah. Then may generate and build do not use one another and duplicate the globa little  If not otherise possible.

lint-staged.config.js
daern91: ahhh makes total sense. Thanks! 😍

PR #673, opened 2018-07-20
Add set valid from and until action for discounts

islam3zzat:
#### Summary
In this PR we add support we combine both  `setValidFrom` and `setValidUntil` to `setValidFromAndUntil` when appropriate 

- Add utility function that combines `validFromAction` and `validUntilAction` if both exist
- Apply `combineValidityActions` to discounts

#### Description

When we apply list of actions, and one of them fails, all fail, which is great.
But in some situation like the following example
**Draft**
```js
{
  validFrom: '01.01.2019',
  validUntil: '10.01.2019'
}
```
**Updated values**
```js
{
  validFrom: '15.01.2019',
  validUntil: '20.01.2019'
}
```
this will generate two update actions
```
[
  {setValidFrom: '15.01.2019'},
  {setValidUntil: '20.01.2019'}
]
```
And when applying the first action _setValidFrom_ we'll have this middle state 
```js
{
  validFrom: '15.01.2019',
  validUntil: '10.01.2019'
}
```
which will fail because `validFrom` should not exede `validUntil`
and this was the motivation for adding `setValidFromAndUntil` action
https://docs.commercetools.com/http-api-projects-productDiscounts.html#set-valid-from-and-until
Which allows us to update `validFrom` and `validUntil` in one step

#### Todo

* Tests
  * [*] Unit
  * [ ] Integration
  * [ ] Acceptance
* [ ] Documentation

### DIFF
--- a/packages/sync-actions/src/utils/combine-validity-actions.js
+++ b/packages/sync-actions/src/utils/combine-validity-actions.js
@@ -1,0 +1,19 @@
+export default function combineValidityActions(actions = []) {
+  const [setValidFromAction, setValidUntilAction] = actions.filter(
+    item => item.action === 'setValidFrom' || item.action === 'setValidUntil'
+  )
+  if (setValidFromAction && setValidUntilAction) {
+    return [
+      ...actions.filter(
+        item =>
+          !(item.action === 'setValidFrom' || item.action === 'setValidUntil')
+      ),
+      {
+        action: 'setValidFromAndUntil',
+        validFrom: setValidFromAction.validFrom,
+        validUntil: setValidUntilAction.validUntil,
+      },
+    ]
+  }
+  return actions
+}
--- a/packages/sync-actions/src/utils/combine-validity-actions.js
+++ b/packages/sync-actions/src/utils/combine-validity-actions.js
@@ -1,13 +1,14 @@
+const validityActions = ['setValidFrom', 'setValidUntil']
+
+const isValidityActions = actionName => validityActions.includes(actionName)
+
 export default function combineValidityActions(actions = []) {
-  const [setValidFromAction, setValidUntilAction] = actions.filter(
-    item => item.action === 'setValidFrom' || item.action === 'setValidUntil'
+  const [setValidFromAction, setValidUntilAction] = actions.filter(item =>
+    isValidityActions(item.action)
   )
   if (setValidFromAction && setValidUntilAction) {
     return [
-      ...actions.filter(
-        item =>
-          !(item.action === 'setValidFrom' || item.action === 'setValidUntil')
-      ),
+      ...actions.filter(item => !isValidityActions(item.action)),
       {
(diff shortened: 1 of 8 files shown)


packages/sync-actions/src/utils/combine-validity-actions.js
```diff
@@ -0,0 +1,19 @@
+export default function combineValidityActions(actions = []) {
+  const [setValidFromAction, setValidUntilAction] = actions.filter(
+    item => item.action === 'setValidFrom' || item.action === 'setValidUntil'
+  )
+  if (setValidFromAction && setValidUntilAction) {
+    return [
+      ...actions.filter(
+        item =>
+          !(item.action === 'setValidFrom' || item.action === 'setValidUntil')
```
adnasa: `!['setValidUntil', 'setValidFrom'].includes(item.action)`

packages/sync-actions/src/utils/combine-validity-actions.js
```diff
@@ -0,0 +1,19 @@
+export default function combineValidityActions(actions = []) {
+  const [setValidFromAction, setValidUntilAction] = actions.filter(
+    item => item.action === 'setValidFrom' || item.action === 'setValidUntil'
```
adnasa: `['setValidUntil', 'setValidFrom'].includes(item.action)`

packages/sync-actions/src/utils/combine-validity-actions.js
```diff
@@ -0,0 +1,19 @@
+export default function combineValidityActions(actions = []) {
+  const [setValidFromAction, setValidUntilAction] = actions.filter(
```
adnasa: I could refactor this a bit

```
const combinedSetValidDatesAction = [
  'setValidFrom',
  'setValidTo'
]

const isOneOfCombined = actionName => combinedSetValidDatesAction.includes(actionName)
```

Then you will notice that this whole function could be a bit simpler

adnasa:
your unit-tests are failing for some odd reason

packages/sync-actions/test/utils/combine-validity-actions.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    })
+    it('should not compine into `setValidFromAndUntil` and keep `validFromAction`', () => {
+      expect(combinedActions).toMatchObject([otherAction, validFromAction])
+    })
+  })
+  describe('when only `setValidUntil` is available', () => {
+    beforeEach(() => {
+      combinedActions = combineValidityActions([otherAction, validUntilAction])
+    })
+    it('should not compine into `setValidFromAndUntil` and keep `validUntilAction`', () => {
```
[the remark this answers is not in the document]
adnasa: keep `compine`, I like it better

tdeekens:
Alright. I'll take care from here. Will merge, release and integrate into the mc-fe. Thanks all.

PR #681, opened 2018-07-26
fix(middleware-auth): use task's own next function when processing pending tasks

lauraluiz:
#### Summary

When a request enters the auth-middleware, it will always continue by calling its own `next` function. That's the behaviour to be expected. But there was one single case in which this was not true: when a token is already being fetched by another request and the task needs to wait in the `pendingTasks` stack while it arrives. In that situation, once the token had been fetched, all pending tasks were calling the `next` method from the task that fetched the token, instead of calling their own `next` function.

#### Description

Instead of the previous behaviour, now each pending task is calling their own `next` function.

Context: this issue led to some unexpected behaviour when the `next` function contains global vars or uses Promises to capture the token to use it with another library, as it was my case. Even though I'm aware my case is not a conventional one, I still think the previous behaviour should be considered a bug. But regarding my case in particular (capturing the token with a Promise to add it to an Apollo request), I'm welcome to alternative solutions ;)

#### Todo

* Tests
  * [ ] Unit
  * [ ] Integration
  * [ ] Acceptance
* [ ] Documentation


packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -146,7 +143,7 @@ export default function authMiddlewareBase(
     return
   }
   // Keep pending tasks until a token is fetched
-  pendingTasks.push({ request, response })
+  pendingTasks.push({ request, response, next })
```
adnasa: A comment explaining this could be useful for the next dev looking into this

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -82,7 +79,7 @@ function executeRequest(
                 // Assign the new token in the request header
                 const requestWithAuth = mergeAuthHeader(token, task.request)
                 // console.log('test', cache, pendingTasks)
-                next(requestWithAuth, task.response)
+                task.next(requestWithAuth, task.response)
```
adnasa: Same here

lauraluiz:
@adnasa done! Let me know if I should improve the description somehow.

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -143,6 +144,9 @@ export default function authMiddlewareBase(
     return
   }
   // Keep pending tasks until a token is fetched
+  // Save next function as well, to call it once the token has been fetched, which prevents
+  // unexpected behaviour in a context in which the next function uses global vars
```
adnasa: 👍

PR #695, opened 2018-08-07
fix(product-json-to-csv): fix not shown zero value in csv export

Babanila:
affects: @commercetools/product-json-to-csv

#### Summary

Fix the zero values not shown in product export of CSV

#### Description
- Fixing of the values '0' not shown in the csv export as shown in the json format.

- This PR closes [this](https://github.com/sphereio/sphere-node-product-csv-sync/issues/219)


packages/product-json-to-csv/package.json
```diff
@@ -47,6 +47,7 @@
     "highland": "2.13.0",
     "json2csv": "3.11.5",
     "lodash": "4.17.10",
+    "lodash.isnil": "^4.0.0",
```
daern91: No need to install this dependency since the whole `lodash` package is installed above on line 49.

packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -140,7 +140,7 @@ export default class ProductMapping {
               // check if product is masterVariant
               if (product.productType)
                 product.productType.attributes.forEach((attribute: Object) => {
-                  if (!acc[attribute.name]) acc[attribute.name] = ''
+                  if (isNil(acc[attribute.name])) acc[attribute.name] = ''
```
daern91: Nice one 👍 👏 

For future reference I would also add a comment here, in the PR, to let readers know why this change is made.

.flowconfig
```diff
@@ -59,4 +59,4 @@ include_warnings=true
 emoji=true
 
 [version]
-^0.74.0
+0.77.0
```
daern91: 👍

packages/product-json-to-csv/test/map-headers.spec.js
```diff
@@ -19,7 +19,7 @@ describe('mapHeaders', () => {
         'name.en',
         'description.en',
         'slug.en',
-        'attribute',
+        'attributes',
```
daern91: I'm wondering why we're removing singular `attribute` here? Is it never used and was just a typo?

packages/product-json-to-csv/test/map-headers.spec.js
```diff
@@ -122,10 +122,10 @@ describe('mapHeaders', () => {
       })
     })
 
-    test('contains `attribute`', () => {
+    test('contains `attributes`', () => {
```
daern91: Same question here, does it mean that singular `attribute` is not a thing?

packages/product-json-to-csv/test/map-headers.spec.js
Babanila: According to the [docs](https://docs.commercetools.com/http-api-projects-products.html#productvariant), it is `attributes`.
`attribute` was declared in the code but was never called or used except in the test.
Guess it may be a typo error.

packages/product-json-to-csv/test/map-headers.spec.js
Babanila: Same as above.

packages/product-json-to-csv/package.json
adnasa: for the future; be careful to pin the version that you install

packages/product-json-to-csv/src/map-headers.js
```diff
@@ -42,6 +42,11 @@ export default function mapHeaders(headers: Array<string>): Array<HeaderObj> {
         label: 'images',
         value: 'variant.images',
       }
+    if (header === 'attributes' || header === 'variant.attributes')
```
adnasa:
```js
['attributes', 'variant.attributes'].includes(header)
```

1. easy to read
2. easy to extend
3. easy to extract for testing purposes
4. did I mention that it is easy to read?

packages/product-json-to-csv/src/map-headers.js
```diff
@@ -42,6 +42,11 @@ export default function mapHeaders(headers: Array<string>): Array<HeaderObj> {
...
+      return {
+        label: 'attributes',
+        value: 'variant.attributes',
```
adnasa: wait... should we always return this no matter what the value of `header` is?

packages/product-json-to-csv/src/map-product-data.js
adnasa: I would additionally add why _mutation_ is necessary here.

packages/product-json-to-csv/test/map-headers.spec.js
adnasa: `attributes` is the right key name.

packages/product-json-to-csv/src/map-product-data.js
Babanila: `if (!acc[attribute.name]) acc[attribute.name] = ''` will return empty string for `undefined`, `null` and `'0'`, that is why the '0' value is not shown in the csv product export.

packages/product-json-to-csv/test/map-headers.spec.js
```diff
@@ -258,10 +258,10 @@ describe('mapHeaders', () => {
       })
     })
 
-    test('contains `attribute`', () => {
+    test('contains `variant.attributes`', () => {
       expect(mapHeaders(headerFields)[14]).toEqual({
-        label: 'attribute',
-        value: 'attribute',
+        label: 'attributes',
+        value: 'variant.attributes',
```
adnasa: general question;

does `variant` entail `masterVariant` as well?

Babanila: Yes

packages/product-json-to-csv/src/map-product-data.js
adnasa: `isNil('0')` doesn't return `true` does it?

image

packages/product-json-to-csv/package.json
Babanila: Removed.
👍

packages/product-json-to-csv/src/map-headers.js
Babanila: Just following the code structure, but I can make another PR for the refactoring.

https://github.com/commercetools/nodejs/blob/a6eca3d2619a6087e11b8301a42f4a9d3aeb5644/packages/product-json-to-csv/src/map-headers.js#L35-L49

packages/product-json-to-csv/src/map-product-data.js
Babanila: Sorry my bad, I meant zero value (0).

packages/product-json-to-csv/test/map-product-data.spec.js
```diff
@@ -181,11 +185,11 @@ describe('::ProductMapping', () => {
                 name: 'color',
                 value: {
                   label: {
-                    it: 'blanco',
-                    de: 'weiss',
-                    en: 'white',
+                    it: 0,
```
junajan: The zero value was already tested using that `lengthOfProduct` attribute so no need to test it also here in the `lenum` attribute.

packages/product-json-to-csv/test/__snapshots__/map-product-data.spec.js.snap
```diff
@@ -45,7 +46,10 @@ Array [
   },
   Object {
     "article": "sample 089 WHT",
-    "color": "white",
+    "color.key": 0,
```
junajan: Why did this change? I would preserve the behavior from the [previous module](https://github.com/sphereio/sphere-node-product-csv-sync/blob/master/src/coffee/exportmapping.coffee) which was mapping lenums like this:
```
attributeName: item key
attributeName.en: English label
``` 
WDYT @Babanila @daern91 ?

adnasa: agreed.
This shouldn't change

image

in `map-product-data.spec.js`

adnasa: good catch!

packages/product-json-to-csv/test/map-product-data.spec.js
Babanila: Ok, thanks

daern91:
Thanks for the update, this looks good now IMO

packages/product-json-to-csv/src/map-product-data.js
Babanila: `if (isNil(acc[attribute.name])) acc[attribute.name] = ' '` was used because it is short and compact and it replaces undefined and null with empty string.

Would you prefer the below ?
`  if ( acc[attribute.name] === undefined ||  acc[attribute.name] === null )
                    acc[attribute.name] = ' ' `

PR #701, opened 2018-08-16
Add channels syncer to sync actions package

tdeekens:
#### Summary

The channels API ([ref](https://docs.commercetools.com/http-api-projects-channels#channels)) is not yet supported by the `sync-actions` package. This PR adds a syncer for them. It includes:

1. Base actions (name, description, etc)
2. Custom fields
3. Roles

It is very much aligned with previous syncers.


packages/sync-actions/src/channels-actions.js
```diff
@@ -0,0 +1,38 @@
+import { buildBaseAttributesActions } from './utils/common-actions'
+import createBuildArrayActions, {
+  ADD_ACTIONS,
+  REMOVE_ACTIONS,
+} from './utils/create-build-array-actions'
+
+export const baseActionsList = [
+  { action: 'setKey', key: 'key' },
```
daern91: Shouldn't this one be `changeKey`?
https://docs.commercetools.com/http-api-projects-channels.html#update-channel

adnasa: this should be [`changeKey`](https://docs.commercetools.com/http-api-projects-channels.html#change-key)

it makes sense as well, since `key` is a required field on the draft

packages/sync-actions/src/channels-actions.js
```diff
@@ -0,0 +1,20 @@
+import { buildBaseAttributesActions } from './utils/common-actions'
+
+export const baseActionsList = [
+  { action: 'setKey', key: 'key' },
```
adnasa: my comment got overidden, this is supposedly `changeKey`

adnasa:
you got two Swedes now nagging about that changeKey 😆

tdeekens:
Thanks @adnasa and @daern91. In commit 25240d47dc70a9d775973ef8c5ca6dead3bd6adc I changed it to `changeKey`. Please have another look.

packages/sync-actions/src/channels-actions.js
```diff
@@ -0,0 +1,20 @@
+import { buildBaseAttributesActions } from './utils/common-actions'
+
+export const baseActionsList = [
+  { action: 'changeKey', key: 'key' },
+  { action: 'changeName', key: 'name' },
+  { action: 'changeDescription', key: 'description' },
+  { action: 'setAddress', key: 'address' },
+  { action: 'setGeoLocation', key: 'geoLocation' },
+  { action: 'setRoles', key: 'roles' },
```
adnasa: This is simple enough.
Maybe update the sync-actions doc highlighting that we do not support `addRoles` and `removeRoles` for now?

tdeekens: I don't see why. They're always a `Array<String>` so there is very little benefit in having add/remove for now. To me it's behind the API and we can iterate on that if customers have 10k channel lists.

tdeekens:
> Just for the sake of completion, we could just add examples into syncActions.md. 😍

Thanks for insisting. Did in one of the later commits. Didn't know about the doc. Also added a note there about the `setRoles` action.

Thanks both for the review 💯. Will release.

montezume:
Thanks guys 😄

PR #709, opened 2018-08-21
fix(personal-data-erasure): fix failing int tests

daern91:
#### Summary

Tests were flaky because we bizarrely tried to delete all data twice from the same project. This would most of the time, weirdly enough, work since we are running the tests in parallel. However, sometimes it would fail since we can not delete what doesn't exist. Duh!

#### Description

PR restructures test file to set up and tear down the project correctly so we have new, fresh, data for each deletion. PR also removes carts from `resourcesToDelete` as it belongs to `customer` and gets deleted that way. (Hence, the tests are still passing).
https://docs.commercetools.com/http-api-projects-carts.html#carts

#### Todo

* Tests
  * [x] Unit
  * [x] Integration

### DIFF
--- a/integration-tests/cli/personal-data-erasure.it.js
+++ b/integration-tests/cli/personal-data-erasure.it.js
@@ -17,7 +17,6 @@ import {
 let projectKey
 if (process.env.CI === 'true') projectKey = 'custom-objects-import-int-tests'
 else projectKey = process.env.npm_config_projectkey
-projectKey = 'custom-objects-import-int-tests'
 
 describe('personal data erasure', () => {
   jest.setTimeout(15000) // 15 second timeout
--- a/integration-tests/cli/personal-data-erasure.it.js
+++ b/integration-tests/cli/personal-data-erasure.it.js
@@ -20,6 +20,8 @@ else projectKey = process.env.npm_config_projectkey
 projectKey = 'custom-objects-import-int-tests'
 
 describe('personal data erasure', () => {
+  jest.setTimeout(15000) // 15 second timeout
+
   const bin = './integration-tests/node_modules/.bin/personal-data-erasure'
 
   describe('CLI basic functionality', () => {
@@ -49,7 +51,7 @@ describe('personal data erasure', () => {
       debug: () => {},
     }
 
-    beforeAll(async () => {
+    const setupProject = async () => {
       const credentials = await getCredentials(projectKey)
       apiConfig = {
         host: 'https://auth.sphere.io',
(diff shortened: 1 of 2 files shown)


integration-tests/cli/personal-data-erasure.it.js
```diff
@@ -17,9 +17,10 @@ import {
 let projectKey
 if (process.env.CI === 'true') projectKey = 'custom-objects-import-int-tests'
 else projectKey = process.env.npm_config_projectkey
-projectKey = 'custom-objects-import-int-tests'
 
 describe('personal data erasure', () => {
+  jest.setTimeout(15000) // 15 second timeout
```
daern91: Setting this here instead of manually for each test.

integration-tests/cli/personal-data-erasure.it.js
```diff
@@ -89,63 +90,77 @@ describe('personal data erasure', () => {
...
       personalDataErasure = new PersonalDataErasure({ apiConfig }, logger)
     })
 
-    describe('export function', () => {
-      it('should get data on the CTP', async () => {
-        const data = await personalDataErasure.getCustomerData(customerId)
+    describe('functional usage', () => {
+      beforeAll(async () => {
+        await setupProject()
+      }, 30000) // 30 second timeout as this might take a while
```
daern91: this only needs to be called once per `describe` block, no need for an `beforeEach` since it takes quite a while to set up.

integration-tests/cli/personal-data-erasure.it.js
```diff
@@ -89,63 +90,77 @@ describe('personal data erasure', () => {
...
+              setTimeout(async () => {
+                data = await personalDataErasure.getCustomerData(customerId)
+                resolve()
+              }, ms)
+            )
+
+          await personalDataErasure.deleteAll(customerId)
+
+          // wait 3s for DB to finish deletion
+          await fetchDataAfterTimeout(3000)
```
daern91: open for other suggestions here than using `setTimeout`

packages/personal-data-erasure/src/main.js
```diff
@@ -203,7 +199,6 @@ export default class PersonalDataErasure {
     const resourcesToDelete = [
       customersResource,
       ordersResource,
-      cartsResource,
```
daern91: This one would also cause errors since it gets deleted inside the `customerResource` already. Sometimes it would be fast enough to delete it before deletion of `customer` was finished, sometimes not.

PR #720, opened 2018-08-28
fix(product-json-to-csv): add a missing dependency

### LINKED ISSUE #719: [product-json-to-csv]: Add a missing dependency
### Support reference number
SUPPORT-3190

### Description
When importing product mapping class from [product-json-to-csv](https://github.com/commercetools/nodejs/blob/master/packages/product-json-to-csv/src/map-product-data.js) I am getting this error:
```
Error: Cannot find module 'babel-runtime/core-js/object/entries'
```

Which says that there is a missing dependency in the module.

junajan:
#### Summary

Fixes #719


junajan:
Hi @daern91  .. I updated my PR here 296dcd1

PR #730, opened 2018-08-29
fix(product-json-to-csv): fix exporting zero products

### LINKED ISSUE #729: [product-json-to-csv]: Bug when doing a full export of zero products
### Support reference number
SUPPORT-3830

### Description
When doing a full export of no products we still try to wait (using `single-emit` package) until all product streams are finished but we don't have any so the package throws an error
```
Error: The constructor expects a collection of items
```
This error comes from this line:
https://github.com/commercetools/nodejs/blob/master/packages/product-json-to-csv/src/writer.js#L84

junajan:
#### Summary
Fixes #729 

#### Description
We should check whether we have any streams (saving products) and if not, we can zip an empty folder right away.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/product-json-to-csv/test/writer.spec.js
+++ b/packages/product-json-to-csv/test/writer.spec.js
@@ -262,6 +262,36 @@ describe('Writer', () => {
       writer.archiveDir(null, outputStream, logger)
     })
 
+    test('throw an error when streams fail', done => {
+      const tempFile = tmp.fileSync({ postfix: '.zip' })
+      const tmpDir = tmp.dirSync({ unsafeCleanup: true })
+      const outputStream = fs.createWriteStream(tempFile.name)
+      const failedStream = fs.createWriteStream(tempFile.name)
+      // throw error while ending stream
+      failedStream.end = () => {
+        failedStream.emit('error', new Error('test error'))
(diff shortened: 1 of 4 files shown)


packages/product-json-to-csv/test/writer.spec.js
```diff
@@ -163,17 +163,51 @@ describe('Writer', () => {
...
+            entries += 1
             if (entry.path === 'products/product-type-1.csv')
               entry.pipe(verifyCsv1)
             else if (entry.path === 'products/product-type-2.csv')
               entry.pipe(verifyCsv2)
+
+            if (entries === 2) {
+              expect(entries).toEqual(2)
+              tempFile.removeCallback()
+              done()
```
daern91: This is painful to the eyes, we should make time asap to make a better implementation.

junajan: Agree .. I will investigate how does it work in that unzip module and try to fix it there

junajan:
Hello @daern91  .. please check my latest commit da8eb65

PR #755, opened 2018-09-04
chore(deps): update deps

daern91:
#### Summary

Updates `rollup` and `jsondiffpatch`

#### Description

Rollup also seem to have included some new install-scripts so we could not longer ignore those with `yarn --ignore-scripts`. Instead, PR removes `bootstrap` and `build` as separate commands from CI as they are already inside the `postinstall` script [here](https://github.com/commercetools/nodejs/blob/master/package.json#L9) anyways.

Finally, there is an issue with `jsondiffpatch` with minified UMD builds when using es6 modules. I will create an issue for that later. Fix for now is to simply use require statements. 

#### Todo

- Tests
  - [x] Unit
  - [x] Integration
- [x] Documentation

resolves #742
resolves #748


packages/sync-actions/package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "fast-equals": "1.6.0",
-    "jsondiffpatch": "0.2.4",
+    "jsondiffpatch": "0.3.11",
```
adnasa: oooh... i don't trust this lib enough to allow a minor upgrade anymore tbh..

adnasa: Looking at the repo, there is not a changelog either. :/

daern91: Hmm, fair point. Only thing I can find is that there are no breaking changes in this update. `0.3.x` only dropped support for old node versions. 
More reading here: https://github.com/benjamine/jsondiffpatch/issues/213

`0.2.4` dates back to September 2016... I'd personally prefer to go through with the update or find a new, better maintained, library to use instead.

adnasa: > I'd personally prefer to go through with the update or find a new, better maintained, library to use instead

👍 
if there is time to evaluate other options, we definitely should do that.

daern91: Sounds good. How about we go through with this one and create a new issue for replacing the library? Then make sure to get it into the planning moving forward.

PR #760, opened 2018-09-06
refactor(sync-actions): remove product-types

adnasa:
#### Summary

- removes product types sync-actions

#### Description

Product Type sync actions was implemented half-baked. The current implementation of the product-types sync-actions (and even with the help of `jsondiffpatch`) can not calculate appropriate sync-actions for attributes and enum values as they are replaced/edited (all values) and removed.

I've written a breaking unit test, to which appropriate update-actions are supposed to be calculated for enum values and it turns out the new version of attribute + enum are introduce edge cases that `sync-actions` can not cover
https://github.com/commercetools/nodejs/compare/adnasa-unit-test-enum-entire-product-type

#### Further problems to discover

if we were to keep `sync-actions/product-types`, we have to add new features to appropriate generate update-actions when there are change to attribute name and changes to the attribute enum value key.

When either an attribute changes its name, or an enum value changes its key, we simply lose the identifier to the `original -> next` version of resource, which would make it impossible to correct a appropriate update-actions.

one option is apply the concept of `hints`, how this could turn is that; __besides__ having just base actions built into the sync-action package of the resource, for other action we could provide **further** `previous, next` values along with a type to the package:

```js
const productTypeSync = createProductTypeSync()
const previous = {
	name: 'previous',
}
const next = {
	name: 'next',
}

const updateActions = productTypeSync.buildActions(previous, next, {
  // hints
  nestedEntityChanges: {
    attributeDefinitions: [
      {
        previous: { name: 'previous-attribute-name' },
        next: { name: 'next-attribute-name' }
      }
    ],
    attributeEnumValues: [
      {
        previous: { key: 'previous-enum-key' },
        next: { key: 'next-enum-key' }
      },
      {
        previous: { key: 'previous-enum-key-2' },
        next: undefined
      }
    ]
  }
})

console.log(updateActions)
// [
//   // this is calculated by `buildBaseAttributesActions`
//   {
//     action: 'changeName',
//     name: 'next-attribute-name'
//   },

//   // these are calculated separately, only through `nestedEntityChanges` hints
//   // because hint are __explicit__, we don't worry about identifiers.
//   {
//     action: 'changeAttributeName',
//     attributeName: 'next-attribute-name'
//   },
//   {
//     action: 'changeEnumKey',
//     key: 'next-enum-key'
//   },
//   {
//     action: 'removeEnumValues'
//     attributeName: 'next-attribute-name'
//     keys: ['previous-enum-key-2'],
//   }
// ]
```

**DO NOT MERGE, this is still used by a client**
I mark this as a **breaking change**, since even though that this is a relatively new package to the `sync-actions`, some other projects could have started using this.


adnasa:
Closed this now and opened #789 instead, which introduces hints approach

Issue #759, opened 2018-09-06
Redesign sync-actions


daern91:
Current sync-actions are very limited in many ways and need to be redesigned.

adnasa:
duplicate of #758 ?

Issue #758, opened 2018-09-06
Replace `jsondiffpatch` lib


daern91:
### Description

This library used inside our [sync-actions](https://github.com/commercetools/nodejs/blob/72f30f7e17d8c134752fb8f9947fa3993a5cc3f6/packages/sync-actions/package.json#L38) package is not very stable or well-maintained. For example there is no changelog at all so every update is at risk of breaking something unexpected. 

We should replace it with something more stable.

PR #789, opened 2018-09-10
refactor(sync-actions/product-types): apply hints for nested value changes. 

adnasa:
#### Summary

- supercedes https://github.com/commercetools/nodejs/pull/760
- removes old implementation for update-action generation for `attributes` and `enums`
- introduces __hints__ approach to appropriately calculate update-actions for `attributes` and `enums`
- _slightly_ related to https://github.com/commercetools/nodejs/issues/759

#### Description

Everything discussed in `syncActions.md`

#### Todo

- Tests
  - [x] Unit
  - ~~Integration~~
  - ~~Acceptance~~
- [x] Documentation


packages/sync-actions/src/product-types.js
```diff
@@ -67,7 +58,14 @@ export default (
     mapActionGroup,
     syncActionConfig
   )
-  const buildActions = createBuildActions(diffpatcher.diff, doMapActions)
+  const onBeforeApplyingDiff = null
+  const buildActions = createBuildActions(
+    diffpatcher.diff,
+    doMapActions,
+    onBeforeApplyingDiff,
+    { withHints: true }
```
adnasa: This could have been part of `syncActionConfig`. However that would mean _leaking_ the underlying API of `createBuildActions`  outwards

adnasa:
One thing to add beyond what is proposed in the PR, is that _one could argue_ that there is possibility to move this logic into `onBeforeApplyingDiff`. However, this would bring us back to square one of implicity, and we might again have trouble calculating appropriate update actions since the identifier (when change) and can't be found..

For now, I don't fully understand why `lint` on CI fails. Running this locally gave me all ✅ 🤔

packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -54,7 +54,12 @@ function injectMissingPriceIds(nextVariants, previousVariants) {
   })
 }
 
-export default function createBuildActions(differ, doMapActions, onBeforeDiff) {
+export default function createBuildActions(
+  differ,
+  doMapActions,
+  onBeforeDiff,
+  syncActionConfig = {}
```
adnasa: this is not `syncActionConfig` anymore

docs/sdk/api/syncActions.md
```diff
@@ -486,6 +486,105 @@ client.execute(productTypesRequest)
 .catch(error => ...)
 ```
 
+#### Difference to sync-actions for other resources
+
+Unlike other resources (e.g `createSyncProducts`), `createSyncProductTypes` requires that you apply hints to calculcate update actions for nested values such as `attributes` and `enumValues`. Since a [change in the API](https://docs.commercetools.com/release-notes.html#releases-2018-03-09-product-type-rename-name-and-enum-key), the [previous implementation was not capable for calculating appropiate update-actions](https://github.com/commercetools/nodejs/pull/760#issue-213684712) when an enum-value has changed its `key`, or when an attribute has changed its `name`. The `key` of an enum value was used as its identifier to calculate correct update-actions. When the intention of a change was `changeEnumKey`, it is discernable for `sync-actions` to appropriately calculate that for you. Same goes for attribute values.
```
yanns: `s/calculcate/calculate`

adnasa: `s/calculcate/calcutta`? :bowtie:

adnasa:
I think it is important to review this an _approach_/_API-to-hints_ point of view... code can always be cleaned up as we go along with the review 😉

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
```diff
@@ -0,0 +1,484 @@
...
+              nestedValuesChanges: {
+                attributeDefinitions: [attributeDefinitionDraftItem],
+              },
+            }
+          )
+        })
+        it('should match snapshot', () => {
+          expect(updateActions).toMatchSnapshot()
+        })
+        it('should generate `removeAttributeDefinition` update-action', () => {
```
daern91: Love the structure of these tests

packages/sync-actions/src/product-types-actions.js
```diff
@@ -50,107 +47,6 @@ const getIsAnAttributeDefinitionBaseFieldChange = diff =>
...
-            attributeName: next.name,
-            action: 'removeEnumValues',
-            keys: removedKeys,
-          },
-        ]
-      : []),
-  ]
-}
-
 export function actionsMapAttributes(
```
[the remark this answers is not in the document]
adnasa: 307301a ho hoo hooooooo

types/sdk.js
```diff
@@ -341,6 +341,7 @@ export type SyncAction = {
 }
 export type SyncActionConfig = {
   shouldOmitEmptyString: boolean,
+  withHints: boolean,
```
adnasa: This is supposed to be removed..

packages/sync-actions/src/product-types-actions.js
```diff
@@ -31,189 +23,245 @@ export function actionsMapBase(diff, previous, next, config = {}) {
...
+  actionDefinition
+) => {
+  const isEmpty = value =>
+    [isNil, nextValue => nextValue === ''].some(operator =>
+      operator(typeof value === 'string' ? value.trim() : value)
+    )
+  return Object.entries(actionDefinition).reduce(
+    (nextUpdateActions, [field, actionFieldDefinition]) => {
+      if (isEmpty(previous[field]) && isEmpty(next[field]))
+        return nextUpdateActions
```
daern91: We need a test for this return.

packages/sync-actions/src/product-types-actions.js
```diff
@@ -31,189 +23,245 @@ export function actionsMapBase(diff, previous, next, config = {}) {
...
+      if (!isEmpty(previous[field]) && isEmpty(next[field]))
+        return [...nextUpdateActions, actionFieldDefinition]
```
daern91: This too

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
```diff
@@ -0,0 +1,492 @@
...
+            })
+            attributeEnumDraftItemToBeRemoved2 = createAttributeEnumDraftItem({
+              previous: { key: 'enum-key-2', label: 'enum-label-2' },
+              next: undefined,
+              hint: {
+                attributeName: 'attribute-enum-with-2-enum-values-to-remove',
+                isLocalized: false,
+              },
+            })
+            attributeEnumDraftItemToBeChanged = createAttributeEnumDraftItem({
```
daern91: Im confused here. How can this become a `changePlainEnumValueLabel` when there is no `previous`?

packages/sync-actions/test/product-types-sync-base.spec.js
```diff
@@ -4,7 +4,7 @@ import { baseActionsList } from '../src/product-types-actions'
 
 describe('Exports', () => {
   test('action group list', () => {
-    expect(actionGroups).toEqual(['base', 'attributes'])
+    expect(actionGroups).toEqual(['base'])
```
daern91: why removing `'attributes'` here?

adnasa: because attributes should not be part of the `actionGroups` which I assume are the target keys for `doMapActions`

https://github.com/commercetools/nodejs/blob/13f0f3a0a2343d0852011ecf71418f93172a16af/packages/sync-actions/src/product-types.js#L37-L47

daern91:
I think my main question is how this will work practically for users. In the other packages we simply just diff the `now` and `before` so a user can just dump their data into the function. 

Is there any way we do this to hints too?

adnasa:
> I think my main question is how this will work practically for users. In the other packages we simply just diff the now and before so a user can just dump their data into the function.

I think I've covered this extensively with documentation that unlike other implementations, we accept only hints for nested values on product-types.

This is why I asked that this is reviewed as an approach and topic and if this is something we can proceed with, then we can apply the same approach to other sync-actions __providing that they have the same problem and needs__, and not having to re-think the underlying API with no time to figure out use cases ref #758 

short answers: no, it's not "practical", and I wasn't trying to comply to how other sync-actions work for the moment, because that doesn't solve the problem. This is the more intuitive approach that I propose.

> Is there any way we do this to hints too?

I don't understand? the point of hints, is that they are provided by the consumer, not specifically calculated internally in `product-types` before passing it down to the underlying `createBuildActions`..

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
daern91: Makes sense. These `previous` and `next` is something I'd have to figure out myself, as a user, then? They won't be generated?

daern91: Got it. Yeah I'd guess just some communication problems. But its starting to get clear for me now. 

So, in essence, we can say that proposed PR will calculate all `base actions` and let the user calculate and provide it for `attributes`/`enums` themselves? Do I understand correctly then?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -31,189 +23,245 @@ export function actionsMapBase(diff, previous, next, config = {}) {
...
+          updatedAttributeDefinition.previous,
+          updatedAttributeDefinition.next,
           {
-            attributeName: next.name,
-            action: 'removeEnumValues',
-            keys: removedKeys,
-          },
-        ]
-      : []),
+            label: {
```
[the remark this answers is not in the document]
adnasa: > Note: `createSyncProductTypes` does not support `changeAttributeName` nor `changeEnumKey` for the moment, but this is something we will add in, in the near future.

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
```diff
@@ -0,0 +1,492 @@
...
+          it('should not generate any update-actions', () => {
+            expect(updateActions).toEqual([])
+          })
+        })
+        describe('with changes', () => {
+          beforeEach(() => {
+            attributeDefinitionDraftItem = createAttributeDefinitionDraftItem({
+              next: {
+                type: { name: 'boolean' },
+                name: 'next-attribute-name',
```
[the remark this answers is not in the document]
adnasa: > Note: `createSyncProductTypes` does not support `changeAttributeName` nor `changeEnumKey` for the moment, but this is something we will add in, in the near future.

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
```diff
@@ -0,0 +1,492 @@
...
+            nestedValuesChanges: {
+              attributeDefinitions: [attributeDefinitionDraftItem],
+            },
+          }
+        )
+      })
+      it('should match snapshot', () => {
+        expect(updateActions).toMatchSnapshot()
+      })
+      it('should generate `addAttributeDefinition` update-action', () => {
```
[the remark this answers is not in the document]
adnasa: no they are not equal. A snapshot can break (because another action), where as what matters here is  that `addAttributeDefinition` always is present.

Also when a snapshot break, it's not uncommon that devs just updates the snap and move on with potentially letting a break slide

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
adnasa: exactly. if they want their nested values calculated, they either...

1. calculate it themselves, which they have to anyways since I intended to have product-types removed entirely
2. use hints to help the sync-actions calculate appropriate actions

cc @junajan

daern91:
@junajan, regarding backward compatibility I think @adnasa made a good case to why it is not possible here: https://github.com/commercetools/nodejs/pull/760. 

The other option would be to delete this package completely since it was not working correctly from the start.

With this in mind I'm in favour of moving forward with this PR. Just need to make sure to

1. Make sure clients are aware of these changes and we document it, properly and _extensively_, when releasing (as breaking).
2. Clean up code, comments and add missing tests.

adnasa:
quick update.
Last I talked to @junajan, he is on board on this API change that I proposed.
but before he shall review again, he wants the time to try out this following idea

> This configuration object should be optional otherwise all customers will have to create their own product-type-sync which will generate this configuration object.

Other than that, this PR is up for code review, as I have addressed concerns besides nitpicks

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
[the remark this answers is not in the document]
junajan: If you just update the snapshot without investigating it then it is completely pointless and this test does not have to be there at all. 

And if it does not generate the `addAttributeDefinition` than it will break also snapshot so that's why I thought that this test is not needed.

junajan:
Hello @daern91 @adnasa .. we already discussed my concerns about this PR that it will force users to create their own solution for calculating changes and hints configuration that can be later provided to this tool which will translate it to CTP API update actions.

All [other syncers](https://commercetools.github.io/nodejs/sdk/api/syncActions.html) (for categories, products, ...) take just two objects containing `prev` and `current` and calculate update actions from it. So I thought it would be nice to give the user some default behavior which would generate hints config automatically based on some predefined update strategy (we would have to document possible impacts it can have). With this, we would avoid the breaking change (if the default update strategy works like the current version of productType sync).

But because it is more complicated to create this tool which would calculate hints config and it would require more time.. I am ok with releasing this change as it is now and creating a feature request which would be processed later if there are people who would benefit from it.

packages/sync-actions/src/product-types-actions.js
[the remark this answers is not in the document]
adnasa: I'm not really sure what you really mean with your question @junajan,
but for `set*`, we don't need to pass a value if `previous` had a value and `next` doesn't.

so this is completely valid:

```js
{
	action: 'setInputTip',
	attributeName: 'x'
	// no `inputTip`, means that we are un-setting the value
}
```

and for `change*` update-actions, this would _break_ in a sense that the API would throw an error. The convention is that `change*` is an indication that this is a __required__ field and can not be unset.

and we would end up with following, which is a valid update-action however with an invalid value

```js
{
	action: 'changeDescription',
	attributeName: 'x'
	// no `description`, indicates that users has attempted to either pass in an empty string, null, or undefined.
	// the API would throw an error here anyways, __even__ if we were to set a `description` key with an empty string.
}
```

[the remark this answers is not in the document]
adnasa: > But where do you add the actual new value for that new field? 

if you are referring to `add*` update-actions, I have an `addAttributeDefinitions` update actions defined in L124

packages/sync-actions/src/product-types-actions.js
```diff
@@ -31,189 +23,245 @@ export function actionsMapBase(diff, previous, next, config = {}) {
...
+// `buildBasAttributesActions` generates update-actions with help of `diff`,
+// which is an object consisting of flags which indicates different operations.
+// `generateBaseFieldsUpdateActions` only generate based on `previous` and `next`.
+export const generateBaseFieldsUpdateActions = (
+  previous,
+  next,
+  actionDefinition
+) => {
+  const isEmpty = value =>
+    [isNil, nextValue => nextValue === ''].some(operator =>
```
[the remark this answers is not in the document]
adnasa: This is confusing...

```js
const isEmpty = value => isNil(value) || String(value).trim() === ''
```

wouldn't work since you are defining `isEmpty` as a _single_ function which is `value => isNil(value)`, where as an empty string is also an important operation to run on the value

packages/sync-actions/test/product-types-sync-attribute-hints.spec.js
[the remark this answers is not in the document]
adnasa: _stay tuned!_

packages/sync-actions/src/product-types-actions.js
adnasa: there is already unit-test for this

enum

https://github.com/commercetools/nodejs/blob/adnasa-apply-hints/packages/sync-actions/test/product-types-sync-attribute-hints.spec.js#L54-L70

attribute
https://github.com/commercetools/nodejs/blob/adnasa-apply-hints/packages/sync-actions/test/product-types-sync-attribute-hints.spec.js#L323-L342

adnasa:
@daern91 @junajan I added test now for `generateBaseFieldsUpdateActions` as well to complete this 2a0a1f0

docs/sdk/api/syncActions.md
```diff
@@ -486,6 +486,109 @@ client.execute(productTypesRequest)
...
+```js
+const productTypeSync = createProductTypeSync()
+const previous = {
+  name: 'previous',
+}
+const next = {
+  name: 'next',
+}
+
+const updateActions = productTypeSync.buildActions(next, previous, {
```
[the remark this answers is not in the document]
junajan: Resolving this as discussed here https://github.com/commercetools/nodejs/pull/789#issuecomment-423114878

packages/sync-actions/src/product-types-actions.js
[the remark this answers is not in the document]
junajan: I tried it and it work as expected:
```js
> const isNil = value => value === null || value === undefined;
undefined
> const isEmpty = value => isNil(value) || value === '';
// const isEmpty = function (value) { return isNil(value) || value === '' }
undefined
> isEmpty(null)
true
> isEmpty(undefined)
true
> isEmpty('')
true
> isEmpty(1234)
false
> isEmpty("abcd")
false
```
For me it looks much simpler than looping through functions but it is your code so if you preffer your solution, go for it.

packages/sync-actions/src/product-types-actions.js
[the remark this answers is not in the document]
junajan: Duplicate of [this comment](https://github.com/commercetools/nodejs/pull/789#discussion_r217004354) so this can be ignored

packages/sync-actions/src/product-types-actions.js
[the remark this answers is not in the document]
junajan: I see it now .. With this `if (!isEmpty(previous[field]) && isEmpty(next[field]))` you are using this action only when user removed the value (it was in `prev` but is not in `next` anymore) so you send an update action without actual value which removes value also on API. 👍

PR #790, opened 2018-09-10
feat(babel): migrate to babel@7

daern91:
#### Summary

PR migrates from **babel@6** to **@7**.

#### Description

One thing to notice here is the solution on how to use the config file. Babel 7 removes support for `.babelrc` inside of monorepos ([ref](https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading)).

We will now use `babel.config.js` instead ([ref](https://babeljs.io/docs/en/config-files#project-wide-configuration)). However, please see the following quote:

> The primary downside of this project-wide config is that, because it relies on the working directory, it can be more painful to use in monorepos if the working directory is not the monorepo root. For example, if you have
> ```
>babel.config.js
>packages/
>  mod1/
>    package.json
>    src/index.js
>  mod2/
>    package.json
>    src/index.js
> ```
> and the individual packages are responsible for running their builds (and their working directory for Babel is the individual packages), the babel.config.js file will not be automatically loaded, and users will be required to set the path to it manually.

So, PR also changes the `build script` across all packages. This might not be the ideal solution so I'd love to get any thoughts and input on it. Would be great if we could find another way.


package.json
```diff
@@ -38,20 +38,18 @@
...
+    "@babel/plugin-proposal-export-default-from": "7.0.0",
+    "@babel/plugin-proposal-export-namespace-from": "7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "7.0.0",
+    "@babel/preset-env": "7.0.0",
+    "@babel/preset-flow": "7.0.0",
     "@commitlint/cli": "7.1.1",
     "@commitlint/config-conventional": "7.1.1",
-    "babel-cli": "6.26.0",
-    "babel-core": "6.26.3",
+    "babel-core": "7.0.0-bridge.0",
```
daern91: This one is still needed for `babel-jest` below ([ref](https://jestjs.io/docs/en/getting-started#using-babel)).

lerna.json
```diff
@@ -1,7 +1,8 @@
 {
-  "lerna": "2.7.1",
+  "lerna": "2.11.0",
```
daern91: We can not yet update to lerna 3.x because of `cz-lerna-changelog` [not being updated](https://github.com/atlassian/cz-lerna-changelog/issues/16)

packages/csv-parser-discount-code/package.json
```diff
@@ -28,7 +28,7 @@
     "lib"
   ],
   "scripts": {
-    "build": "babel src --out-dir lib"
+    "build": "babel src --out-dir lib --config-file '../../babel.config.js'"
```
daern91: Other ideas would be appreciated here. I do prefer this over having a config file inside of every package but it's definitely not ideal.

lerna.json
emmenko: Hmm, why are we still using that?

packages/csv-parser-discount-code/package.json
emmenko: I think it's fine like this

lerna.json
daern91: It's used by `commitizen` when running the commit script.

https://github.com/commercetools/nodejs/blob/083e5590c50f28b5d50dc0d904903aad3a7ccb21/package.json#L15

emmenko: I would get rid of it. I think no one is using that command...?

babel.config.js
```diff
@@ -0,0 +1,48 @@
...
+        targets: { node: '6' },
+        modules: 'commonjs',
+        useBuiltIns: 'usage',
+        include: ['es7.object.entries'],
+      }
+    default:
+      break
+  }
+  return {}
+}
```
tdeekens: Nicely done.

jest.transform.js
```diff
@@ -0,0 +1,3 @@
+const babelConfig = require('./babel.config')
+
+module.exports = require('babel-jest').createTransformer(babelConfig)
```
tdeekens: Yep. Have the same on mc-fe and flopflip. This js good and does he job.

lerna.json
daern91: Sounds good, will do. I'll keep commitizen itself tho as it's amazing for on-boarding students and introducing them to angular commit convention.

tdeekens: Ask around in the team. I am not aware of anyone using it anymore. It was good introducing the style but should not be needed anymore. I would recommend removing it in amothrr PR.

package.json
tdeekens: Yup.

packages/csv-parser-discount-code/package.json
tdeekens: Yes this is fine. To my knowledge the idea of a babel.config over an rc without being breaking is to be more explicit. Meaning you pass it, you import it etc.

packages/state-importer/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Import states to the commercetools platform",
   "main": "lib/main.js",
   "scripts": {
-    "build": "babel src --out-dir lib"
+    "build": "babel src --out-dir lib --config-file '../../babel.config.js'"
```
tdeekens: You could maybe also pass it as an env var fro, the host project. BABEL_CONFIG=„blub“ and use it here via interpolation. I may have seen that somewhere. Not sure it serves a bug benefit however as it never moves etc.

rollup.config.js
```diff
@@ -32,11 +33,15 @@ const config = {
...
       include: ['node_modules/**'],
     }),
-    babel({
-      babelrc: true,
-      exclude: ['node_modules/**'],
-      runtimeHelpers: true,
-    }),
+    babel(
+      Object.assign(
+        {
```
tdeekens: I thought you could spread here. Not sure what the nvmrc says in the repo these days 🎃

babel.config.js
```diff
@@ -0,0 +1,48 @@
...
+      return {
+        targets: { node: '6' },
+        modules: 'commonjs',
+        useBuiltIns: 'usage',
+        include: ['es7.object.entries'],
+      }
+    default:
+      break
+  }
+  return {}
```
junajan: Just a nitpick: this `return {}` can be in `default` branch instead of `break`.

rollup.config.js
junajan: I believe that this file is not getting transpiled by babel and as we support node from version v6 we should not use any of new syntax here. Also, the spread operator based on ES6 works only with arrays.

lerna.json
adnasa: commitizen itself is so drilled into my spine that it is part of my muscle memory now.
However, if you feel that you want to keep this to ease "onboarding", or in some extent ensuring consistency in commit messages, then don't listen to us 👍

rollup.config.js
daern91: Yup, it doesn't. 

I tried ES7 Object spread first but it failed on Node6 :(

babel.config.js
daern91: Ahh, good find

PR #793, opened 2018-09-12
chore(renovate): ignore lerna dependency update

Babanila:
#### Summary

Dependency update ignore

#### Description
Updating dependency lerna from v2.11.0 to v3.3.1 will cause a `BREAKING CHANGE` details [here](https://github.com/lerna/lerna/blob/master/CHANGELOG.md), this affect the use of commitizen on momorepo.

An issues was created at [atlassian/cz-lerna](https://github.com/atlassian/cz-lerna-changelog/issues/16) for the breaking change.

**The changes includes the following:**
- Drop support for node v4.
- Add universal hosted git URL support. 
- Convert command lifecycle from callbacks to Promises.
- Preserve package.json structure during bootstrap mangling.
- publish : lerna.json bootstrapConfig and publishConfig namespaces are replaced with - --command.bootstrap and command.publish, respectively.
- package : The Package class no longer provides direct access to the JSON object used to construct the instance.
- add: lerna add now only supports adding one dependency at a time.
etc.

Closes #772

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -729,7 +729,7 @@
 
 "@semantic-release/commit-analyzer@^2.0.0":
   version "2.0.0"
-  resolved "https://registry.yarnpkg.com/@semantic-release/commit-analyzer/-/commit-analyzer-2.0.0.tgz#924d1e2c30167c6a472bed9f66ee8f8e077489b2"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -729,7 +729,7 @@
```
daern91: Also, just a nitpick. This file should not have to be updated in this PR.

PR #808, opened 2018-09-17
refactor: use tc39 optional chaining

daern91:
#### Summary

After the babel 7 migration finished we also have the option to use a bunch of new proposals. PR introduces support for [optional chaining](https://github.com/tc39/proposal-optional-chaining) which is moving into `stage-3` as of now.

To give a few examples of where this can be useful PR also refactors a couple lines as an example.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration


daern91:
Closing this one for now as of problems with new syntax and `flow-remove-types plugin`

PR #813, opened 2018-09-19
fix(sync-actions): only return price id from matching variant

daern91:
As @butenkor found in this comment https://github.com/sphereio/sphere-product-import/issues/128#issuecomment-422759154, current implementation of `injectMissingPriceIds` might lead to problems where we inject the id of a price from the wrong variant.

PR changes this to only check for prices with matching variant ID instead.


packages/sync-actions/src/utils/create-build-actions.js
```diff
@@ -19,33 +19,28 @@ function arePricesStructurallyEqual(oldPrice, newPrice) {
...
+  return price ? price.id : null
 }
 
 function injectMissingPriceIds(nextVariants, previousVariants) {
   return nextVariants.map(newVariant => {
     const { prices, ...restOfVariant } = newVariant
 
     if (!prices) return restOfVariant
+    const oldVariant = previousVariants.find(
+      previousVariant => previousVariant.id === newVariant.id
```
junajan: Hi @daern91 it looks really good... the only thing I am thinking about is if it is a correct approach that we are trying to match variants by their `ID`. There are also `SKU` and `KEY` which can be used if they are present. cc @butenkor

daern91: Im not 100% sure here either, but it seems like `id` is the only **required** value of the three? https://docs.commercetools.com/http-api-projects-products.html#productvariant

butenkor: I am not sure how you can match by variant ID as it is an internally generated ID by CTP, not known to external system which delivers updated/new data. Only fields which are known to both sides are SKU and variant key so those are the only candidates. I tend to use SKU as it is unique as key and has to be provided if you maintain inventories (SKU will be also mandatory later if we have dedicated price endpoint). Variant ID is also neither mandatory on create: https://docs.commercetools.com/http-api-projects-products.html#productvariantdraft
nor update as you can update by SKU too: https://docs.commercetools.com/http-api-projects-products.html#add-price

daern91: Ahh, that is a really good point. I guess we could just check if there is a match with any of the three, then?

Something like
```
const oldVariant = previousVariants.find(
  previousVariant => {
    previousVariant.id === newVariant.id ||
    previousVariant.key === newVariant.key ||
    previousVariant.sku === newVariant.sku 
  }
)
```

butenkor: Seems ok 👍

junajan: sounds good to me too

PR #835, opened 2018-09-24
fix(product-json-to-xlsx): add missing CLI parameter

### LINKED ISSUE #834: [product-json-to-xlsx] Add missing parameter - onlyMasterVariants
### Support reference number
In the new XLSX product mapper, we are missing `onlyMasterVariants` CLI parameter

junajan:
#### Summary
Fixes #834 

Note: This feature is tested already in `product-json-to-csv`

### DIFF
--- a/packages/product-json-to-xlsx/src/cli.js
+++ b/packages/product-json-to-xlsx/src/cli.js
@@ -94,6 +94,11 @@ Required scopes: ['view_products']`,
       'Define if product attributes like name should be added to each variant row.',
     type: 'boolean',
   })
+  .option('onlyMasterVariants', {
+    describe: 'Export only masterVariants from products.',
+    type: 'boolean',
+    default: false,
+  })
   .option('language', {
     alias: 'l',
     default: 'en',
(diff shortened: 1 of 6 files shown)


packages/product-json-to-xlsx/src/cli.js
```diff
@@ -201,6 +206,7 @@ Promise.all([getHeaders(args), resolveCredentials(args)])
       fillAllRows: args.fillAllRows,
       language: args.language,
       multiValueDelimiter: args.multiValueDelimiter,
+      onlyMasterVariants: args.onlyMasterVariants,
```
daern91: Maybe we should also add this here? https://github.com/commercetools/nodejs/blob/c7daa22c81415497cfc8041360ade51e4dd1d58a/packages/product-json-to-xlsx/test/main.spec.js#L19

junajan: It has a default value `onlyMasterVariants = false` so it worked even without it.
I added it to be more clear + it wanted also to bump flow version 0d744ef

PR #840, opened 2018-09-27
fix(auth): encode username and password for password flow

### LINKED ISSUE #839: Password flow doesn't accept `+` char in email
### Support reference number

https://jira.commercetools.com/browse/SUPPORT-4005

### Description

User email needs to be URI encoded

daern91:
#### Summary

PR encodes user email and password which the platform requires. More reading in these two threads ([1](https://jira.commercetools.com/browse/SUPPORT-4005) and [2](https://jira.commercetools.com/browse/CTP-1652)).

#### Todo

- Tests
  - [x] Integration

resolves #839

### DIFF
--- a/integration-tests/sdk/auth.it.js
+++ b/integration-tests/sdk/auth.it.js
@@ -7,7 +7,7 @@ import {
 } from '@commercetools/sdk-middleware-auth'
 import { createHttpMiddleware } from '@commercetools/sdk-middleware-http'
 import fetch from 'node-fetch'
-import { clearData, createData } from './../cli/helpers/utils'
+import { clearData, createData } from '../cli/helpers/utils'
 
 let projectKey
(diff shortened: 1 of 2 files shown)


junajan:
should we encode password too?

daern91:
@junajan Why not, I guess. It seemed to work without it already, with that new complex password, but I'll add it anyways.

lauraluiz:
I wonder how, though. It should not be working both with urlencoding and without it. At least I cannot make sense of it 🤔 
When you changed the email and added the `+`, before urlenconding it, was it failing?

daern91:
You are completely right, @lauraluiz :-) 

I added a `+` to the password inside the int test and then it did, in fact, fail for me locally. So I updated the PR with encoded password and a `+` inside password now.

PR #894, opened 2018-10-16
feat(category-exporter): add new exporter package

Babanila:
#### Summary

PR adds package to export categories from CTP in JSON.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration
- [x] Documentation


packages/category-exporter/src/main.js
```diff
@@ -0,0 +1,97 @@
...
+      .process(
+        { uri: this.buildURI(), method: 'GET' },
+        (payload: Object): Promise<any> => {
+          if (payload.statusCode !== 200)
+            return Promise.reject(
+              new Error(`Request returned error ${payload.statusCode}`)
+            )
+          const results = payload.body.results
+          const JSONResults = JSON.stringify(results)
+          outputStream.write(JSONResults)
```
daern91: This solution will fail if we try to export a lot of categories as it will write each array individually. It will look something like below

```
[
  ...categoryObjects
]
[
  ...categoryObjects
]
[
  ...categoryObjects
]
```

One solution would be to simply destruct all of the arrays and write all objects into a single one instead. Another one would be to pipe them all through something like [JSONStream](https://github.com/dominictarr/JSONStream).

packages/category-exporter/test/main.spec.js
```diff
@@ -0,0 +1,144 @@
...
+          body: {
+            results: [],
+          },
+        }
+        categoryExport.client.execute = jest
+          .fn()
+          .mockReturnValue(Promise.resolve(results))
+      })
+
+      test('should write to outputStream', done => {
```
daern91: I dont think `should write to outputStream` is the best description here. Maybe something like `should throw error`?

packages/category-exporter/test/main.spec.js
```diff
@@ -0,0 +1,144 @@
...
+          body: {
+            results: [{ id: 'id1', name: 'foo' }, { id: 'id2', name: 'bar' }],
+          },
+        }
+        categoryExport.client.execute = jest
+          .fn()
+          .mockReturnValue(Promise.resolve(results))
+      })
+
+      test('should write to outputStream', done => {
```
daern91: Same here, needs better description.

packages/category-exporter/test/main.spec.js
```diff
@@ -0,0 +1,144 @@
...
+          done()
+        })
+        categoryExport.run(outputStream)
+      })
+    })
+  })
+
+  describe('::buildURI', () => {
+    describe('without predicate', () => {
+      test('should build default uri', () => {
```
daern91: Hmm maybe `should build uri` makes more sense?

types/categoryExporter.js
```diff
@@ -0,0 +1,30 @@
...
+  apiUrl?: string,
+}
+
+export type CategoryExporterOptions = {
+  apiConfig: ApiConfigOptions,
+  accessToken?: string,
+  predicate?: string,
+  logger?: LoggerOptions,
+  batchSize?: number,
+  continueOnProblems?: boolean,
```
daern91: `batchSize` and `continueOnProblems` are not part of the code?

integration-tests/cli/category-exporter.it.js
```diff
@@ -0,0 +1,77 @@
...
+    // Get test credentials
+    const credentials = await getCredentials(projectKey)
+    apiConfig = {
+      host: 'https://auth.sphere.io',
+      apiUrl: 'https://api.sphere.io',
+      projectKey,
+      credentials,
+    }
+
+    await clearData(apiConfig, 'categories')
```
adnasa: Isn't it enough to clear data only on `afterAll`?

packages/category-exporter/package.json
```diff
@@ -0,0 +1,49 @@
...
+        "@commercetools/get-credentials": "^2.1.7",
+        "@commercetools/sdk-client": "1.5.8",
+        "@commercetools/sdk-middleware-auth": "4.0.6",
+        "@commercetools/sdk-middleware-http": "4.0.4",
+        "@commercetools/sdk-middleware-user-agent": "1.2.10",
+        "node-fetch": "2.2.0",
+        "pino": "^5.7.0",
+        "pino-pretty": "^2.2.1",
+        "pretty-error": "^2.1.1",
+        "yargs": "^12.0.2"
```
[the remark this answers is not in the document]
Babanila: Ok, thanks

integration-tests/cli/category-exporter.it.js
Babanila: It is just a safety measure, to be sure that no data from the beginning.

packages/category-exporter/test/main.spec.js
```diff
@@ -56,6 +56,7 @@ describe('CategoryExporter', () => {
 
       test('should write to outputStream', done => {
         const outputStream = streamtest.v2.toText((error, data) => {
+          console.log('object')
```
daern91: Please remove this one

Babanila: Ok

PR #897, opened 2018-10-16
fix(product-json-to-csv): fix mapping of boolean values #896

### LINKED ISSUE #896: [product-json-to-csv/xlsx] Map properly boolean values
### Support reference number

SUPPORT-4134

### Description

Old XLSX export had mapped boolean values as `false` and `true`
New XLSX export maps boolean as an empty string for a false value and `TRUE` for true value.

junajan:
#### Summary

Fixes #896 

#### Description

In the old `sphere-node-product-csv-sync` module we were mapping booleans to string `true` and `false` - https://github.com/sphereio/sphere-node-product-csv-sync/blob/master/src/coffee/exportmapping.coffee#L204

This PR adds the same behavior also here.

*NOTE* - after this is done, we should bump the version of `product-json-to-csv` module in `product-json-to-xlsx` which uses this mapping when generating XLSX exports.
#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation


daern91:
Thx @junajan, could you please look into the `codecov/patch`? 80% sounds a bit too low comparing to the ~99% target.

PR #903, opened 2018-10-24
fix(product-json-to-csv): allow to set multiple  languages for mapping

### LINKED ISSUE #900: [product-json-to-csv] Some attributes are not exported
### Support reference number
SUPPORT-4134

### Description
When doing a product export of products which have `ltext` or `set of ltext` attribute, first exported products can have this attribute missing so we won't add possible language versions to the CSV header.

Next exported products will have this attribute missing too because the header was already written to a CSV file so it can't be changed with additional attributes (eg `color.de`, `color.en`). 

In the old `csv-sync` module which was originally used for exporting products, we were able to provide a list of languages and it generated ltext header names for given languages.
https://github.com/sphereio/sphere-node-product-csv-sync/blob/master/src/coffee/exportmapping.coffee#L55

We should do the same logic also here.

junajan:
#### Summary
Fixes #900

#### Description
This PR allows the user to set multiple languages (where the first provided is taken as a `mainLanguage`). Languages will then be used to generate default empty localized values for `ltext/lenum` attributes.

For example, if the product has a `ltext` attribute `color` and languages are `en,de`, final CSV will contain: `color.en,color.de`

This is done because we use the first product to create a CSV/XLSX header and if the first product does not have all languages, the CSV header will not have all header names and all products which come later will lose values for those extra languages.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration
  - [ ] Acceptance
- [x] Documentation


junajan:
Hello @adnasa @daern91  .. thanks for you review.
I added also a small change which generates default (aka empty) lang values for name/slug/description fields. I also tested it in `product-import` so it can handle for example an import of an empty slug.

PR #907, opened 2018-10-24
chore: require minimum NPM 5.6 and use updated gitbook-cli repo

daern91:
#### Summary

When running `Lerna@3` or above we need to use `NPM@5.6.0` or above. `gitbook-cli` uses `5.1.0` and thus, breaks our publishing process.

Since (https://github.com/GitbookIO/gitbook-cli) does not seem to be super active, I propose we use [this](https://github.com/daern91/gitbook-cli) forked version until the PR with updated deps (hopefully) gets merged. Forked version uses `NPM@5.10.0` which is stable and works.

EDIT: Since install was failing I changed PR to use [Selective dependency resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) instead.

EDIT2: Both approaches above failed because lerna/npm-ci yelled about missing lock files. As I don't want to add them please do not merge this one until we've found a proper solution.


daern91:
@junajan fixed now by adding `--no-ci` to `lerna bootstrap` command. More reading [here](https://github.com/lerna/lerna/issues/1449)

PR #914, opened 2018-10-26
feat(sdk-auth): add several new features to sdkAuth module

### LINKED ISSUE #909: [sdk-auth] Add missing sdk-auth features
When adding the new `sdk-auth` module to Impex I found out that there are some missing features which we want to use in Impex in order to log in user, retrieve tokens without the need to provide `projectKey` and some other minor things:
 - make projectKey not mandatory (not used in all auth flows)
 - differentiate passwordFlows for client (/oauth/token) and customer (/oauth/{project-key}/customers/token) 
 - add refreshToken=false option to passwordFlow (so API won't generate refresh token if not needed)
 - add possibility to override projectKey/scopes
 - add login flow (/login with accessToken and user email&password)

junajan:
#### Summary
Resolves #909 

#### Description
This PR adds these new features to sdkAuth:
 - `customFlow` - one can set a custom config properties for every auth call
 - `clientPasswordFlow` - log in with `username&password` against `/oauth/token` endpoint
 - set custom config in every call

#### Todo

- Tests
  - [x] Unit
  - [x] Integration
  - [ ] Acceptance
- [x] Documentation


docs/sdk/api/sdkAuth.md
```diff
@@ -24,10 +24,11 @@ Creates an auth client to handle authorization against the commercetools platfor
 #### Named arguments (options)
 
 1.  `host` _(String)_: the host of the OAuth API service
-2.  `projectKey` _(String)_: the key of the project to assign the default scope to
-3.  `credentials` _(Object)_: the client credentials for authentication (`clientId`, `clientSecret`)
-4.  `scopes` _(Array)_: a list of [scopes](https://docs.commercetools.com/http-api-authorization.html#scopes) (default `manage_project:{projectKey}`) to assign to the OAuth token
-5.  `fetch` _(Function)_: A `fetch` implementation which can be e.g. `node-fetch` or `unfetch` but also the native browser `fetch` function. Only needs be be passed if not globally available (e.g. through `isomorphic-fetch`)
+2.  `projectKey` _(String)_: the key of the project to assign the default scope to (optional).
+3.  `disableRefreshToken` _(boolean)_: whether the API should generate a refresh token
```
lauraluiz: Out of curiosity, how did you find out about this parameter? I cannot find it in the platform's documentation 😮

packages/sdk-auth/src/auth.js
```diff
@@ -81,19 +81,31 @@ export default class SdkAuth {
     return Buffer.from(`${clientId}:${clientSecret}`).toString('base64')
   }
 
+  static _getScopes(scopes: ?Array<string>, projectKey: ?string) {
+    return scopes
+      ? scopes.join(' ')
+      : [constants.MANAGE_PROJECT, projectKey].filter(Boolean).join(':') // generate a default scope manage_project:projectKey
```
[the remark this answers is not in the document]
junajan: This is because if you don't provide a `projectKey` it will build an incorrect scope name.
In some cases you can provide scopes also like this:
```
scope=view_users_and_organizations view_projects
```
And API will create a token with these limitations for all your projects.

docs/sdk/api/sdkAuth.md
junajan: I was walking through some already created projects where I found it.
If you set `disableRefreshToken` to `true`, it will add `&refresh_token=false` to form params and then API won't generate refresh token.

docs/sdk/api/sdkAuth.md
```diff
@@ -24,10 +24,11 @@ Creates an auth client to handle authorization against the commercetools platfor
...
+4.  `credentials` _(Object)_: the client credentials for authentication (`clientId`, `clientSecret`)
+5.  `scopes` _(Array)_: a list of [scopes](https://docs.commercetools.com/http-api-authorization.html#scopes) (default `manage_project:{projectKey}`) to assign to the OAuth token
+6.  `fetch` _(Function)_: A `fetch` implementation which can be e.g. `node-fetch` or `unfetch` but also the native browser `fetch` function. Only needs be be passed if not globally available (e.g. through `isomorphic-fetch`)
```
lauraluiz: *Only needs TO be passed

PR #949, opened 2018-11-07
feat(csv-parser-price): handle import of Money customField in object format

### LINKED ISSUE #948: [csv-parser-price] Support mapping objects in price custom fields
### Support reference number
SUPPORT-4251

### Description
When importing prices we ignore the custom fields containing objects

For example this CSV:
```csv
variant-sku,value.currencyCode,value.centAmount,customType,customField.kiloPrice.centAmount,customField.kiloPrice.currencyCode
SKU-support-4251-2,EUR,100,price-withKiloPrice,3000,GBP
```
Is mapped as:
```json
{
	"prices": [{
		"prices": [{
			"custom": {
				"fields": {},
				"type": {
					"id": "24992e40-25a0-40da-a42a-cdf9d4c49fd9"
				}
			},
			"value": {
				"centAmount": 100,
				"currencyCode": "EUR"
			}
		}],
		"sku": "SKU-support-4251-2"
	}]
}
```

junajan:
#### Summary
Resolves #948 

#### Description
When exporting Money objects in custom fields through price exporter it is now possible to import them back.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation


packages/csv-parser-price/src/map-custom-fields.js
```diff
@@ -118,8 +120,30 @@ export default (function MapCustomFields() {
...
+        // map Number properties to integer
+        (val, key) => (numberKeys.includes(key) ? parseInt(val, 10) : val)
+      )
+
+      return {
+        data,
+      }
+    }
+
+    // handle money value provided as a string
```
daern91: Question, do we ever use money value as a string anymore? I thought this was deprecated? https://docs.commercetools.com/http-api-types#money

junajan: For example in `product-export/import` we serialize prices to string while mapping them to CSV/XLSX.

But for price export/import .. I don't know. I would leave it here for a while just to be sure we cover this in case there is some tool using this string format.

packages/csv-parser-price/src/map-custom-fields.js
```diff
@@ -118,8 +120,30 @@ export default (function MapCustomFields() {
...
+   * Takes a value and returns a CTP money object
+   * @param value String "USD 123" or object {centAmount: '123', currencyCode: 'USD'}
+   * @returns {{ data: CTP price object, error: Error message if any }}
+   */
   function mapMoney(value) {
     const result = {}
+
+    // handle money value provided as an object
+    if (typeof value === 'object' && value.centAmount && value.currencyCode) {
+      const numberKeys = ['centAmount', 'fractionDigits']
```
daern91: I think we need to add `preciseAmount` here in order to also support `HighPrecisionMoney `

https://docs.commercetools.com/http-api-types#highprecisionmoney

packages/csv-parser-price/src/map-custom-fields.js
daern91: Sounds good to me!

packages/csv-parser-price/src/map-custom-fields.js
daern91: Maybe also a test for it so we test both [Money](https://docs.commercetools.com/http-api-types#money) and [HighPrecisionMoney](https://docs.commercetools.com/http-api-types#highprecisionmoney)?

junajan: Hi @daern91 .. thx for your comment. I added `preciseAmount` together with a test here e29b4ad

PR #985, opened 2018-11-21
feat(resource-deleter): new resource deleter package

### LINKED ISSUE #699: Add new `deleter` package
### Description

We have a [sphere-deleter](https://github.com/mmoelli/sphere-deleter) package which should be migrated to the commercetools monorepo instead of being an old coffeescript repo.

This will mostly be used for support and testing purposes but is often used, therefore it would be great if we had it inside the monorepo for easy access and extension.

### Requirements

Package should have the same features as old package and also include option to unpublish products.

Babanila:
#### Summary

PR adds new resource-deleter package to CT monorepo.

#### Description

resource-deleter package is design to  delete resource on commercetools platform.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration
- [x] Documentation
 
References #699


docs/cli/resource-deleter.md
```diff
@@ -0,0 +1,119 @@
...
+
+The constructor accepts two arguments:
+
+- A required object containing the following values:
+  - `apiConfig` (Object): `AuthMiddleware` options for authentication on the commercetools platform. (Required. See [here](https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html#named-arguments-options))
+  - `accessToken` (String): [Access token] to be used to authenticate requests to API. Requires scope of [`manage_products`, `manage_customers`, `manage_types`]. More info on how to get the access token [here](https://docs.commercetools.com/http-api-authorization.html#authorization-flows)
+  - `resource` (Object): [resource] that need to be deleted.
+  - `predicate` (String): Query string specifying (where) predicate. More info on predicates [here](https://docs.commercetools.com/http-api.html#predicates) (Optional)
+- An optional object containing the following values:
+  - deleteHours (Number): Amount of time to delete a resource (lastModified)
```
daern91: I see this is both documented and also an option inside the CLI but there is no functionality for it?

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,106 @@
...
+import { getCredentials } from '@commercetools/get-credentials'
+import { exec } from 'mz/child_process'
+import { version } from '@commercetools/resource-deleter/package.json'
+import {
+  cart,
+  categories,
+  customer,
+  sampleProductType,
+  anotherSampleProductType,
+} from './helpers/resource-delete.data'
```
daern91: why not also import, create and test the other sample data? I see you created `review` and `payment` for example.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,106 @@
...
+      expect(stderr).toBeFalsy()
+      expect(stdout).toBe(`${version}\n`)
+    })
+  })
+
+  describe('resource-deleter function', () => {
+    let filePath
+    let stdout
+    let stderr
+    const resources = ['carts', 'categories', 'customers', 'productTypes']
```
daern91: We should test all the resources which are supported according to docs: https://github.com/commercetools/nodejs/pull/985/files#diff-7aee84337ab0ab2deb7a11a30db78801R6

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,106 @@
...
+        resources.map(async resource => {
+          ;[stdout, stderr] = await exec(
+            `${bin} -o ${filePath} -p ${projectKey} -r ${resource}`
+          )
+          results.push({ stdout, stderr })
+        })
+      )
+    })
+
+    it('should log success messages', async () => {
```
daern91: I think this is really cool, however, maybe its a better idea to spread them out so its easier to debug? I mean run one test per resource so we also see a beautiful tree. E.g.
```
describe('when deleting carts')
  it('should log success message')
describe('when deleting categories')
  it('should log success message')
describe('when deleting customers')
  it('should log success message')
...
...
...
```

What do you think?

docs/cli/resource-deleter.md
adnasa: more documentation, the better.

integration-tests/cli/resource-deleter.it.js
adnasa: this lib already reached jest core
https://github.com/mattphillips/jest-chain

use it

```js
it.each(
  [assertionName, expectedValue, ...args]
)(
  'should do something',
  (expectedValue, ...args) => {
  }
)

```

adnasa:
Is it ready for review again or?
I'm getting notifications however no mention on whether the concerns raised by @daern91 have been addressed

Babanila:
> Is it ready for review again or?
> I'm getting notifications however no mention on whether the concerns raised by @daern91 have been addressed

Hi @adnasa, the package is ready for review again now. Please do not mind the broken int test. It got introduced with `jest-chain` and I will fix it asap. The error happens in an unrelated package :)

integration-tests/cli/csv-parser-price.it.js
```diff
@@ -103,17 +103,17 @@ describe('CSV and CLI Tests', () => {
       const tmpFile = tmp.fileSync()
       const expectedError = 'Row length does not match headers'
       const csvFilePath = path.join(samplesFolder, 'faulty-sample.csv')
-
-      await expect(
-        exec(
+      try {
```
Babanila: Try/catch was used here to catch the error because async/await could not catch the error due to the introduction of jest-chain dependency.

integration-tests/jest.test.config.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
   testRegex: '(.*)\\.it\\.js$',
+  setupTestFrameworkScriptFile: 'jest-chain',
```
Babanila: Introducing `jest-chain` dependency for multiple assertions on the same variable.

adnasa: do we really need this? 
`jest-chain` should have come right out of the box

integration-tests/package.json
```diff
@@ -46,6 +46,7 @@
     "node-fetch": "2.3.0",
     "streamtest": "1.2.4",
     "tmp": "0.0.33",
-    "unzip": "0.1.11"
+    "unzip": "0.1.11",
+    "jest-chain": "1.0.5"
```
adnasa: This should not be needed. As already mentioned, it is part of core by now

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,118 @@
...
+      request,
+      (response: ClientResponse): Promise<any> => {
+        if (response.statusCode !== 200) {
+          return Promise.reject(
+            new Error(`Request returned status code ${response.statusCode}`)
+          )
+        }
+        const results = response.body?.results
+        if (!results || !results.length)
+          return Promise.reject(new Error(`Nothing to delete`))
```
adnasa: I think we can provide a more useful message than this.
WDYT? cc @daern91

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,118 @@
...
+          )
+      },
+      { accumulate: false }
+    )
+  }
+
+  buildUri(projectKey: string, predicate: ?string): string {
+    const service = createRequestBuilder({
+      projectKey: this.apiConfig.projectKey,
+    })[this.resource]
```
adnasa: This smells danger to me.
Do we at least prompt user that they are about to delete an entity (of any entity type)?

integration-tests/jest.test.config.js
daern91: Ahh, interesting. I checked the docs real quick and guess this is what you refer to, @adnasa ?

https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout

If so I definitely agree on using the core part instead of adding an unnecessary dependency

packages/resource-deleter/src/main.js
daern91: Good point, maybe we could add a prompt inside the CLI? Just to make sure they are aware its final and not reversible.

packages/resource-deleter/src/cli.js
```diff
@@ -0,0 +1,147 @@
...
+      predicate: args.where,
+      logger: {
+        error: logger.error.bind(logger),
+        warn: logger.warn.bind(logger),
+        info: logger.info.bind(logger),
+        debug: logger.debug.bind(logger),
+      },
+    }
+    const resourceDeleter = new ResourceDeleter(deleterOptions)
+    const response = await prompts({
```
Babanila: Add prompt message to the user for confirmation

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,176 @@
...
+  let resourceDeleter
+  let logger: {
+    error: () => {},
+    warn: () => {},
+    info: () => {},
+    debug: () => {},
+  }
+  const bin = './integration-tests/node_modules/.bin/resource-deleter'
+
+  function getResource(resource) {
```
Babanila: The function is used to check the live data if the data is delete.

docs/cli/resource-deleter.md
```diff
@@ -0,0 +1,115 @@
...
+      host: 'https://auth.sphere.com'
+      project_key: 'PROJECT_KEY',
+      credentials: {
+        clientId: '*********',
+        clientSecret: '*********'
+      }
+    },
+    accessToken: '123456yuhgfdwegh675412wefb4rgb',
+    resource: 'my-resource'
+    predicate: 'key="my-desired-key"'
```
daern91: missing commas

~resource: 'my-resource'~
~predicate: 'key="my-desired-key"'~
resource: 'my-resource'**,**
predicate: 'key="my-desired-key"'**,**

packages/resource-deleter/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "json"
+    ],
+    "author": {
+        "name": "Babajide Williams",
+        "email": "babajide.williams@commercetools.de"
+    },
+    "license": "MIT",
+    "dependencies": {
+        "@commercetools/api-request-builder": "4.0.2",
+        "@commercetools/get-credentials": "^2.1.10",
```
daern91: plz pin this one 👍

docs/cli/resource-deleter.md
```diff
@@ -0,0 +1,115 @@
...
+For more direct usage, it is possible to use this module directly:
+
+```js
+import resourceDeleter from '@commercetools/resource-deleter'
+import fs from 'fs'
+
+const options = {
+    apiConfig: {
+      apiUrl: 'https://api.sphere.io'
+      host: 'https://auth.sphere.com'
```
daern91: missing comma

docs/cli/resource-deleter.md
```diff
@@ -0,0 +1,115 @@
...
+
+For more direct usage, it is possible to use this module directly:
+
+```js
+import resourceDeleter from '@commercetools/resource-deleter'
+import fs from 'fs'
+
+const options = {
+    apiConfig: {
+      apiUrl: 'https://api.sphere.io'
```
daern91: missing comma

integration-tests/cli/resource-deleter.it.js
```diff
@@ -67,16 +98,18 @@ describe('Resource Deleter', () => {
   // clear resources on API
   afterAll(async () => {
     await Promise.all([
+      await Promise.all([
```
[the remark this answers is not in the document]
adnasa: Looks redundant to me

integration-tests/cli/resource-deleter.it.js
```diff
@@ -100,467 +133,44 @@ describe('Resource Deleter', () => {
...
+    'zones',
+    'productTypes',
+    'categories',
+    'channels',
+    'customerGroups',
+    'customers',
+    'customObjects',
+    'inventory',
+    'payments',
+  ])('should delete resource', resource => {
```
adnasa: boo yah

integration-tests/cli/resource-deleter.it.js
```diff
@@ -100,467 +133,44 @@ describe('Resource Deleter', () => {
...
-    })
-
-    it('should delete zones', async () => {
-      const data = await fs.readFile(filePath, { encoding: 'utf8' })
-      expect(data)
-        .toBeDefined()
-        .toEqual('')
-        .toHaveLength(0)
+  describe.each([
+    'carts',
```
adnasa: sorted alphabetically would be appreciated

integration-tests/cli/resource-deleter.it.js
```diff
@@ -100,467 +133,44 @@ describe('Resource Deleter', () => {
...
-        .toMatch(/Starting to delete fetched zones/)
-        .toMatch(/All deleted/)
-    })
-
-    it('should delete zones', async () => {
-      const data = await fs.readFile(filePath, { encoding: 'utf8' })
-      expect(data)
-        .toBeDefined()
-        .toEqual('')
-        .toHaveLength(0)
```
adnasa: burn it.. all of it ☝️ 🔥

packages/resource-deleter/src/cli.js
```diff
@@ -126,7 +128,20 @@ resolveCredentials(args)
...
     }
-    return new ResourceDeleter(deleterOptions)
-  })
-  .then(resourceDeleter => resourceDeleter.run())
-  .catch(errorHandler)
+    const resourceDeleter = new ResourceDeleter(deleterOptions)
+    const response = await prompts({
+      type: 'confirm',
+      name: 'value',
+      message: `You are about to delete all ${
```
adnasa: nitpick: _... all of..._

integration-tests/cli/resource-deleter.it.js
```diff
@@ -66,41 +105,523 @@ describe('Resource Deleter', () => {
...
+    const resource = 'customObjects'
+
+    beforeAll(async () => {
+      filePath = tmp.fileSync().name
+      ;[stdout, stderr] = await exec(
+        `${bin} -o ${filePath} -p ${projectKey} -r ${resource}`
+      )
+    })
+
+    it('should log success messages', async () => {
```
adnasa: do you really need to `async` this?

integration-tests/cli/resource-deleter.it.js
```diff
@@ -66,41 +105,523 @@ describe('Resource Deleter', () => {
...
+    const resource = 'discountCodes'
+
+    beforeAll(async () => {
+      filePath = tmp.fileSync().name
+      ;[stdout, stderr] = await exec(
+        `${bin} -o ${filePath} -p ${projectKey} -r ${resource}`
+      )
+    })
+
+    it('should log success messages', async () => {
```
adnasa: same question here

integration-tests/cli/resource-deleter.it.js
```diff
@@ -66,41 +105,523 @@ describe('Resource Deleter', () => {
...
+    const resource = 'inventory'
+
+    beforeAll(async () => {
+      filePath = tmp.fileSync().name
+      ;[stdout, stderr] = await exec(
+        `${bin} -o ${filePath} -p ${projectKey} -r ${resource}`
+      )
+    })
+
+    it('should log success messages', async () => {
```
adnasa: I'm confused about async on normal assertions

integration-tests/jest.test.config.js
adnasa: yes it was. sorry for the late reply

integration-tests/cli/resource-deleter.it.js
Babanila: It is okay but I think the deleter package should be used for deleting only.

integration-tests/cli/resource-deleter.it.js
Babanila: I removed it because it does not work without the `jest-chain` dependency

Babanila:
> Please provide a more informative commit message in the future. what did your commit do? what did you refactor? was it a change on a feature? was it to enable a feature development?

Noted.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,176 @@
...
+  reviews,
+  shippingMethods,
+  taxCategories,
+  types,
+  zones,
+} from './helpers/resource-delete.data'
+import { createData, clearData } from './helpers/utils'
+
+let projectKey
+if (process.env.CI === 'true') projectKey = 'resource-deleter-int-test'
```
junajan: one liner?
```js
const projectKey = process.env.CI === 'true' ? 'resource-deleter-int-test' : process.env.npm_config_projectkey
```

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    return client.execute(request)
+  }
+
+  beforeAll(async () => {
+    // Get test credentials
+    const credentials = await getCredentials(projectKey)
+    apiConfig = {
+      host: 'https://auth.sphere.io',
```
junajan: maybe this can be moved to some shared place so when we finally decide to leave sphere.io we can do it just once in constants file.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,176 @@
...
+    await Promise.all([
+      createData(apiConfig, 'carts', carts),
+      createData(apiConfig, 'products', products),
+      createData(apiConfig, 'shippingMethods', shippingMethods),
+    ])
+  }, 30000)
+
+  // clear resources on API
+  afterAll(async () => {
+    await Promise.all([
```
junajan: as described above - this can be simplified by iterating over a list of resources or by using `Promise.map`:
```js
await Promise.map(['carts', 'products', 'shippingMethods'], (resource) => clearData(apiConfig, resource))
```

integration-tests/cli/resource-deleter.it.js
junajan: Some resources are dependant on others (eg: you can't remove shipping info when there are any orders using it) so when sorted alphabetically it may throw some errors.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,176 @@
...
+    })
+
+    it('should print the module version given the version flag', async () => {
+      const [stdout, stderr] = await exec(`${bin} --version`)
+      expect(stderr).toBeFalsy()
+      expect(stdout).toBe(`${version}\n`)
+    })
+  })
+
+  describe.each([
```
junajan: Could you please add also a test also for using `where` predicate?

packages/resource-deleter/src/cli.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+    const resourceDeleter = new ResourceDeleter(deleterOptions)
+    const response = await prompts({
+      type: 'confirm',
+      name: 'value',
+      message: `You are about to delete all ${args.resource} from this project.
+      WARNING: This operation is final and is not reversible. 
+      Are you sure about this?`,
+      initial: false,
+    })
```
junajan: I love this idea. This will save us from a lot of potential problems. One thing I would also add is a `--yes` or `-y` parameter which will skip this confirmation - eg if you want to use it by some other script it will be easier to add one more parameter instead of pushing something to stdin.

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,124 @@
...
+    if (predicate) service.where(predicate)
```
junajan: Add also a `perPage`. By default API returns batch of 20 elements - we can increase it to the max of 500 items per request.

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,124 @@
...
+        const results = response.body?.results
+        if (!results || !results.length)
+          return Promise.reject(
+            new Error(
+              `No ${this.resource} is found in the project ${
+                this.apiConfig.projectKey
+              }, therefore nothing to delete.`
+            )
+          )
+        return Promise.all(
```
junajan: What about some concurrency? If we fetch for example 500 products, we don't want to send 500 delete requests at once..

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,124 @@
...
+          results.map(
+            (result: Object): Promise<any> =>
+              this.client.execute({
```
junajan: Here the things will, unfortunately, get a little bit complicated. There are some edge cases which have to be treated differently.
First one is when deleting published products - you have to unpublish them first or API will return an error.
The second case is when deleting categories - you should delete first children before parents.

Also, the question is what should happen when one deletion throws an error. I think we should log an error and continue with other elements?

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,124 @@
...
+                  .withVersion(result.version)
+                  .build(),
+                method: 'DELETE',
+              })
+          )
+        )
+          .then((): void => this.logger.info('All deleted'))
+          .catch(
+            (error: Error): Promise<Error> => {
+              this.logger.info(error)
```
junajan: Shouldn't an error be logged as an error? `this.logger.error(error)`

packages/resource-deleter/test/main.spec.js
```diff
@@ -0,0 +1,166 @@
...
+              { id: 'fooboo3', key: 'foboKey', version: 3 },
+            ],
+          },
+        }
+
+        resourceDeleter.client.execute = jest.fn(() => Promise.resolve(payload))
+      })
+
+      test('should delete fetched resource', async () => {
+        const data = await resourceDeleter.run('sample-test-project')
```
junajan: `resourceDeleter.run()` fn does not take any arguments.. The same thing occurs also bellow.

packages/resource-deleter/test/main.spec.js
```diff
@@ -0,0 +1,166 @@
...
+          body: {
+            results: [],
+          },
+        }
+
+        resourceDeleter.client.execute = jest.fn(() => Promise.resolve(payload))
+      })
+
+      test('should throw error for empty resource', () => {
+        expect(resourceDeleter.run('sample-test-project')).rejects.toThrow(
```
junajan: Imho this should not be an error. The expected outcome of the delete operation is, that selected elements won't be on API anymore. If they do not exist and we run delete operation, the outcome is the same which is a valid scenario and not an error.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,289 @@
...
+    it(`The ${resource} offspring deleted before its ancestors`, async () => {
+      const payload = await getResource(resource)
+      expect(payload.body.results).toHaveLength(3)
+      await resourceDeleter.run()
+      const newPayload = await getResource(resource)
+      expect(newPayload.body.results).toHaveLength(0)
+    })
+  })
+
+  describe('should delete a published product', () => {
```
junajan: 👍  good tests

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,212 @@
...
+          this.logger.info(
+            `No ${this.resource} is found in the project ${
+              this.apiConfig.projectKey
+            }, therefore nothing to delete.`
+          )
+          return Promise.resolve('nothing to delete')
+        }
+
+        // Check if the resource is categories
+        if (this.resource === 'categories') {
```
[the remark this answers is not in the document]
Babanila: - The problem is that the `parent categories` does not have the `parent attributes`

- The `parent categories` & `categories without child` have the `ancestors attribute` and the value for both is `[]` .

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,185 @@
...
+        (response: ClientResponse): Promise<any> => {
+          if (response.statusCode !== 200) {
+            return Promise.reject(
+              new Error(`Request returned status code ${response.statusCode}`)
+            )
+          }
+          let results = response.body?.results
+
+          // Check if the resource is empty
+          if (!results || !results.length) {
```
junajan: Hello @Babanila  .. the `client.process()` method calls the processing function for every batch of items loaded from API. If there are no results loaded, it won't call processing function which means that this code should not be needed at all.

packages/resource-deleter/src/main.js
```diff
@@ -0,0 +1,185 @@
...
+                deletedItems += 1
+                return this.deleteResource({
+                  ...result,
+                  version: newVersion || result.version,
+                })
+              }
+            )
+          ).then(
+            (): void =>
+              this.logger.info(`${deletedItems} ${this.resource} deleted`)
```
junajan: Thank you for adding the counter but here it is pointless - because `deletedItems` will be every time equal to `results.length` (you always delete all loaded results). I thinking more about creating a global counter which would at the very end (line 122) log the total count of all items deleted (and not only count of deleted items in every batch).

packages/resource-deleter/src/main.js
junajan: so based on the internal parts of process method it calls the callback function every time at least once (even if there are no results)

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,335 @@
...
+
+const projectKey =
+  process.env.CI === 'true'
+    ? 'resource-deleter-int-test'
+    : process.env.npm_config_projectkey
+
+describe('Resource Deleter', () => {
+  let apiConfig
+  let resourceDeleter
+  let logger: {
```
daern91: why not import this one from here? https://github.com/commercetools/nodejs/blob/9e598aaaed4ec29d232538cef1ed7bd189ee7136/packages/resource-deleter/src/utils/silent-logger.js

types/resourceDeleter.js
```diff
@@ -0,0 +1,78 @@
...
+    clientId: string,
+    clientSecret: string,
+  },
+  scopes: Array<string>,
+  apiUrl?: string,
+}
+
+/* Client */
+// eslint-disable-next-line max-len
+export type MethodType =
```
daern91: This type is already declared here: https://github.com/commercetools/nodejs/blob/0ded6ee95bc19a667f001f870f14f2d1c4aa30e9/types/sdk.js#L4

types/resourceDeleter.js
```diff
@@ -0,0 +1,78 @@
...
+  | 'GET'
+  | 'POST'
+  | 'DELETE'
+  | 'HEAD'
+  | 'OPTIONS'
+  | 'PUT'
+  | 'PATCH'
+  | 'TRACE'
+
+export type ClientRequest = {
```
daern91: and this one here: https://github.com/commercetools/nodejs/blob/0ded6ee95bc19a667f001f870f14f2d1c4aa30e9/types/sdk.js#L15

types/resourceDeleter.js
```diff
@@ -0,0 +1,78 @@
...
+  uri: string,
+  method: MethodType,
+  body?: string | Object,
+  headers?: {
+    [key: string]: string,
+  },
+}
+
+export type HttpErrorType = {
```
daern91: https://github.com/commercetools/nodejs/blob/0ded6ee95bc19a667f001f870f14f2d1c4aa30e9/types/sdk.js#L30

?

types/resourceDeleter.js
```diff
@@ -0,0 +1,78 @@
...
+
+export type SuccessResult = {
+  body: Object,
+  statusCode: number,
+  headers?: Object,
+}
+
+export type ClientResult = SuccessResult | HttpErrorType
+
+export type CustomClientResult = ClientResult & {
```
daern91: Why not just import `ClientResult` instead of copying everything above?

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,345 @@
...
+  describe('should delete categories with its grandchildren', () => {
+    const resource = 'categories'
+    beforeAll(async () => {
+      const options = {
+        apiConfig,
+        resource,
+        logger,
+      }
+
+      const parentCategories = [
```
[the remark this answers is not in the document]
Babanila: Moving it to the file will be a problem when trying to create categories like `parentCategories`, `childCategories` etc because bluebird promise is used to import from the file.
Importing only some part of the data is not possible.
```
await Promise.each(Object.keys(resources), name => {
      return createData(apiConfig, name, resources[name])
    })
```

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,345 @@
...
+      ]
+
+      resourceDeleter = new ResourceDeleter(options)
+      await createData(apiConfig, resource, parentCategories)
+      await createData(apiConfig, resource, childCategories)
+      await createData(apiConfig, resource, grandChildCategories)
+      await createData(apiConfig, resource, greatGrandChildCategories)
+      await createData(apiConfig, resource, greatGreatGrandChildCategories)
+
+      const categories = [
```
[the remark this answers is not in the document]
Babanila: Removed.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,345 @@
...
+        parentCategories,
+        childCategories,
+        grandChildCategories,
+        greatGrandChildCategories,
+        greatGreatGrandChildCategories,
+      ]
+
+      return categories
```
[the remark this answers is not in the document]
Babanila: Removed.

integration-tests/cli/resource-deleter.it.js
```diff
@@ -0,0 +1,345 @@
...
+  describe('should delete a published product', () => {
+    const productType = [
+      {
+        name: 'sampleProductType',
+        key: 'sample-product-123PTKey',
+        description: 'Sample Product Type',
+        version: 1,
+      },
+    ]
+    const products = [
```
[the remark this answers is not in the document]
Babanila: Same as for parentCategories above.

integration-tests/cli/resource-deleter.it.js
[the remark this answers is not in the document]
daern91: Ahh, makes sense, thanks!

PR #1007, opened 2018-12-06
fix(product-json-to-csv): fix prefixing of conflicting attribute names

### LINKED ISSUE #970: No prefix for Lenum attributes (name same as keywords) while exporting
### Support reference number

SUPPORT-4036

### Description

The attributes which has the same name as the keywords such "productType" are exported with a prefix "attribute". If there exists an attribute "productType" then it is exported as attribute.productType. If this attribute is an lenum we have an additional column in export file such as "productType.en" without prefix.

### Expected Behavior

The attribute with the language should also have the prefix.

### Current Behavior

No prefix on the exported file

### Context

The customer receives an error, if the exported file is imported because of the prefix missing for the attribute.

enumattribute

junajan:
#### Summary
Fixes #970 

#### Description
When exporting a product with no product description field but an attribute named as description, we save incorrectly the attribute to `description` column instead of prefixing it with an `attribute.` string.

This PR adds a list of conflicting names which should be prefixed.
#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/product-json-to-csv/src/map-product-data.js
+++ b/packages/product-json-to-csv/src/map-product-data.js
@@ -32,6 +32,7 @@ export default class ProductMapping {
   categoryBy: 'name' | 'key' | 'externalId' | 'namedPath'
   mainLanguage: string
   languages: Array<string>
+  productLevelProperties: Array<string>
   multiValDel: string
 
   constructor({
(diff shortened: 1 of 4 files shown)


packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -50,6 +51,31 @@ export default class ProductMapping {
     this.languages = languages || [language]
     this.multiValDel = multiValueDelimiter
     this.createShortcuts = createShortcuts
+
+    this.productLevelProperties = [
+      'id',
+      'key',
+      'version',
+      'createdAt',
+      'lastModifiedAt',
```
daern91: do we really need `createdAt` and `lastModifiedAt` here? Just thinking since we dont update them no?

maybe we could add a comment above with the link to where to fetch this list?

junajan: all attributes which have the same name as any item from this list are prefixed with `attribute.` string. So I would say we should also prefix `createdAt` because then it can make problems when importing.

junajan: updated here 4af192e

PR #1021, opened 2018-12-11
feat(sdk-auth): add invalidateTokenInfo fn, lock fetching

### LINKED ISSUE #1006: [sdk-auth] Invalidate token to use new flow
### Description

I'm using the sdk-auth module with the TokenProvider and I have initialized it with the following:
```
const tokenProvider = new TokenProvider({
  sdkAuth: new SdkAuth({ ... }),
  fetchTokenInfo: sdkAuth => sdkAuth.clientCredentialsFlow(),
  ...
});
```
When a customer logs in, I want to change the way the token is fetched with:
```
  tokenProvider.fetchTokenInfo = sdkAuth => sdkAuth.customerPasswordFlow({ username, password });
```
But when I do this, the tokenProvider still returns me the old token instead of giving me a new token with the new flow.

### Expected Behavior

When I change the fetchTokenInfo function because of logging in, I would like the new token provided (by calling `tokenProvider.getTokenInfo()`) to be fetched with the new flow.

### Current Behavior

When I call `tokenProvider.getTokenInfo()`, it still gives me the token from the old flow.

### Context

I'm trying to use the sdk-auth to authenticate to commercetools using Apollo as the GraphQL client.

### Possible Solution

Providing a method to invalidate the token, so that I ensure that the next call of `tokenProvider.getTokenInfo()` will invoke the new flow.

junajan:
#### Summary
Resolves #1006 

#### Description
This PR brings following changes:
 - add `invalidateTokenInfo()` method
 - locks fetching so we send only one request to api
 - fix issue with missing `refresh_token` property

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [x] Documentation

### DIFF
--- a/packages/sdk-auth/src/tokenProvider.js
+++ b/packages/sdk-auth/src/tokenProvider.js
@@ -8,6 +8,8 @@ export default class TokenProvider {
   tokenInfo: ?TokenInfo
   sdkAuth: SdkAuth
   onTokenInfoChanged: (tokenInfo: TokenInfo) => void
+  fetchTokenInfoPromise: ?Promise<TokenInfo>
+  refreshTokenFlowPromise: ?Promise<TokenInfo>
   fetchTokenInfo: (sdkAuth: SdkAuth) => Promise<TokenInfo>
   onTokenInfoRefreshed: (
     newTokenInfo: TokenInfo,
@@ -40,6 +42,9 @@ export default class TokenProvider {
 
     this.sdkAuth = sdkAuth
     this.tokenInfo = tokenInfo
+
+    this.fetchTokenInfoPromise = null
+    this.refreshTokenFlowPromise = null
   }
 
   static _validateTokenInfo(tokenInfo: TokenInfo) {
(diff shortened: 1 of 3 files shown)


packages/sdk-auth/test/token-provider.spec.js
```diff
@@ -266,5 +283,34 @@ describe('Token Provider', () => {
...
+      expect(_tokenProvider.sdkAuth.refreshTokenFlow).toHaveBeenCalledWith(
+        'refreshToken'
+      )
+
+      expect(refreshedInfo).toEqual(['refreshedInfo', 'refreshedInfo'])
+    })
+  })
+
+  describe('invalidateTokenInfo', () => {
+    test('should invalidate stored tokenInfo', async () => {
```
daern91: no need for `async` here

packages/sdk-auth/src/tokenProvider.js
```diff
@@ -90,14 +113,22 @@ export default class TokenProvider {
       .then(
         (tokenInfo: TokenInfo): void => {
           newTokenInfo = tokenInfo
-          newTokenInfo.refresh_token = oldTokenInfo?.['refresh_token']
+          if (oldTokenInfo?.['refresh_token'])
```
daern91: why not `oldTokenInfo?.refresh_token` like in the assignment below? Looks a bit strange with two different ways of calling the same.

packages/sdk-auth/src/tokenProvider.js
```diff
@@ -64,11 +69,29 @@ export default class TokenProvider {
         new Error('Method "fetchTokenInfo" was not provided')
       )
 
-    return Promise.resolve(this.fetchTokenInfo(this.sdkAuth))
+    // do not run fetchTokenInfo more than once at any given time and cache the result
+    if (this.fetchTokenInfoPromise) return this.fetchTokenInfoPromise
+
+    this.fetchTokenInfoPromise = Promise.resolve(
+      this.fetchTokenInfo(this.sdkAuth)
```
daern91: woah, what is happening here, is it a nested promise?

junajan: Hi @daern91  - the idea behind this code is, that `this.fetchTokenInfo()` can return a non-promise value so that's why is it wrapped into a promise.

Also, we use `this.fetchTokenInfoPromise` variable to lock this function so it won't fetch multiple tokens from API. Instead, it will start fetching token only on the first call and for all next calls, it will return the same promise so one result will be replicated to all function calls

packages/sdk-auth/src/tokenProvider.js
junajan: Flow was complaining that the property should not be in underscore notation. I am not sure why it complained only in this case.

packages/sdk-auth/test/token-provider.spec.js
junajan: Updated here ee9a2c1

packages/sdk-auth/src/tokenProvider.js
daern91: ahhh, ok, then I guess we can just leave it there.

PR #1083, opened 2019-01-14
chore: drop support for node 6

daern91:
#### Summary

As Node 6 is getting closer to reaching its end date (april 2019 - [ref](https://github.com/nodejs/Release#release-schedule)), I suggest we drop support for it now and start running CI on node 8/10 instead of 6/8.

PR updates this and makes sure we are targeting node 8 in future build outputs through Babel.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration


daern91:
Pinging @junajan @tdeekens @adnasa, are you OK with this?

adnasa:
I don't mind

.circleci/config.yml
```diff
@@ -46,24 +46,24 @@ jobs:
...
+    <<: *node8Environment
     steps:
       - checkout
       - restore-cache: *restore_yarn_cache
       - run: *yarn_install
       - save_cache: *save_yarn_cache
       - run: *unit_test
-  test_unit_node_8:
-    <<: *node8Environment
+  test_unit_node_10:
```
tdeekens: You can add node_11 if you want.

daern91: instead or in addition to 10?

daern91: Interesting idea, what is the thought behind it? My first reaction to it was to just wait for Node 12 instead since 11 will never be lts.

PR #1134, opened 2019-02-10
fix(): unused global variables error

### LINKED ISSUE #1133: [sdk] fix wrong eslint fail report of `request` and `headers`
### Description
Wrong flow & eslint fail report of `request` and `headers` not used in 'sdk-auth` and `middlewares`.

Flow & eslint test failed locally and sometimes on CI due to the comment statement of  `/* global fetch Request Headers */`.

### Expected Behavior

The tests should be pass.

### Current Behavior

Test failed.

Babanila:
#### Summary
Unused global variables error.

#### Description

Global variables`request` and `headers` were declared but not used  in `sdks`.
#### Todo

- Tests
  - [x] Unit
  - [x] Integration

Closes #1133


packages/sdk-auth/src/auth.js
```diff
@@ -1,6 +1,5 @@
 /* @flow */
-/* global fetch Request Headers */
-
+// global fetch Request Headers
```
daern91: Is this some new rule to eslint regarding comments or why are we making this change?

Babanila: When `yarn flow` is run, `request` and `headers` are both reported as an error  and sometimes it does happen too on CI as well.

daern91: I would suggest digging into this one a bit deeper. What did the error say and why are we having them there inside a comment? It looks like a configuration setting to me.

Babanila: It is eslint change of rules usage [see here](https://eslint.org/docs/rules/no-unused-vars).

daern91: Sorry @Babanila but I am not following here. What is the change of rule?

Babanila: The global variables `Request` and `Headers` was defined but were not used in the file,  setting them in the eslint configuration file to avoid the error report [docs link](https://eslint.org/docs/user-guide/configuring#specifying-globals).

PR #1141, opened 2019-02-12
fix(): only pin devDependencies

sheepsteak:
#### Summary

- [x] Update `.prettierignore` to allow more files to be parsed
  * Allows `package.json` to be parsed now that Prettier handles correctly
  * Fixes spacing in some of the `package.json` files
- [x] Add `^` to all `dependencies` to promote semver range
- [x] Tell Renovate to only pin `devDependencies`
  * This is the [recommended setting for JavaScript libraries](https://renovatebot.com/docs/presets-config/#configjs-lib)

#### Description

At the moment all `dependencies` and `devDependencies` are pinned to a particular version. This is fine for `devDependencies`, but for `dependencies` it means that users of `@commercetools/*` are stuck with the exact version you specify and can't get reuse of package versions across all their dependencies.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration


.prettierignore
```diff
@@ -1,4 +1 @@
-package.json
-package-lock.json
```
daern91: Could you please put `package-lock.json` back into the ignore? We don't use these files but in case someone installs with npm instead of yarn I think it would make sense to ignore formatting them.

sheepsteak: Done 👍

PR #1142, opened 2019-02-15
feat(api-request-builder): add My Shopping Lists and My Payments services

sheepsteak:
#### Summary

- [x] Add [My Shopping Lists](https://docs.commercetools.com/http-api-projects-me-shoppingLists) and [My Payments](https://docs.commercetools.com/http-api-projects-me-payments) services
#### Description

These services are currently missing from the default services so to use them we have to add them as custom services. Since [My Cart](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-services.js#L158) already exists in there I thought I would add these two as well to save us needing to do a custom service for each.
#### Todo

- Tests
  - [x] Unit
  - [x] Integration

### DIFF
--- a/packages/api-request-builder/src/default-services.js
+++ b/packages/api-request-builder/src/default-services.js
@@ -179,6 +179,18 @@ export default {
       features.queryExpand,
     ],
   },
+  myShoppingLists: {
+    type: 'my-shopping-lists',
+    endpoint: '/me/shopping-lists',
+    features: [
+      features.create,
+      features.update,
+      features.del,
+      features.query,
+      features.queryOne,
+      features.queryExpand,
+    ],
+  },
   orders: {
     type: 'orders',
     endpoint: '/orders',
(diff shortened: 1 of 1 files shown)


sheepsteak:
Before this gets merged can I add [MyCustomerProfile](https://docs.commercetools.com/http-api-projects-me-profile) too?

Babanila:
> Before this gets merged can I add [MyCustomerProfile](https://docs.commercetools.com/http-api-projects-me-profile) too?

Thanks, but no need to add it now because `MyCustomerProfile` is still in the beta stage. [See docs](https://docs.commercetools.com/http-api-projects-me-profile)

sheepsteak:
OK, but MyShoppingLists are in beta too. Also MyCarts is too according to the docs and that's already in. Do the docs just need updating.

sheepsteak:
OK, I'll close this PR for now but keep the branch on my fork 😄

PR #1167, opened 2019-02-28
feat(): add support for head method on sdk middleware http

### LINKED ISSUE #1022: add support for HEAD methods on `sdk-middleware-http`
### Support reference number

- [PCM-1078](https://jira.commercetools.com/browse/PCM-1078)

### Description

On `sdk-middleware-http`, there is a lack of support for clients to send `HEAD` requests.

The request is sent, however the promise from the request couldn't resolve, since the fetcher itself expects that there is a body to be parsed

https://github.com/commercetools/nodejs/blob/4d5391c295c7a1b83ddf94564a8a45b3290a16ca/packages/sdk-middleware-http/src/http.js#L112

However a HEAD response doesn't contain any body

### Expected Behavior

Resolve the promise of the request even when the method is `HEAD`

### Current Behavior

At the moment, the request is sent to the API.
When the API responds, the promise for a request couldn't be resolved.

### Context

We want to use HEAD request in a use case such as checking if a server is up and running or an index of a specific project is still active

### Possible Solution

Update `http.js` to call `next` and return

Babanila:
#### Summary

Add support for HEAD requests.

#### Description

This PR  support HEAD requests to be sent by the clients and receive a response without throwing an error.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration

  Closes #1022


packages/sdk-middleware-http/src/http.js
```diff
@@ -112,9 +112,10 @@ export default function createHttpMiddleware({
             res.json().then((result: Object) => {
```
emmenko: Shouldn't we simply check the method if it's `HEAD` and return without parsing the response body? 🤔

adnasa:
I don't know if this is the right approach and it's been a while that I looked into this.. IIRC, you just resolve it.

packages/sdk-middleware-http/src/http.js
```diff
@@ -112,9 +112,11 @@ export default function createHttpMiddleware({
...
               const parsedResponse: Object = {
                 ...response,
-                body: result,
                 statusCode: res.status,
               }
+
+              if (fetchOptions.method !== 'HEAD') parsedResponse.body = result
```
emmenko: As mentioned before, you need to check for the method **before** parsing the body, as `HEAD` responses do not send a body.

I'm not sure how I can make it more clear. What is confusing you? Can I help you somehow?

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -59,6 +59,37 @@ describe('Http', () => {
...
+        includeResponseHeaders: true,
+        fetch,
+      }
+      const httpMiddleware = createHttpMiddleware(httpOptions)
+      nock(testHost)
+        .defaultReplyHeaders({
+          'Content-Type': 'application/json',
+        })
+        .head('/foo/foo')
+        .reply(200, {})
```
daern91: Maybe this might help clarify things.

You reply with an empty body `{}` here, which is not the same as when the body does not exist at all. If you change it to `.reply(200)` you should see the real issue here.

packages/sdk-middleware-http/src/http.js
```diff
@@ -109,12 +109,21 @@ export default function createHttpMiddleware({
       fetcher(url, fetchOptions).then(
         (res: Response) => {
           if (res.ok) {
+            if (fetchOptions.method === 'HEAD') {
```
adnasa: This is what I had in mind when I created the issue.
I think this would help.

PR #1168, opened 2019-02-28
feat(product-json-to-csv): add customerGroup to product prices

### LINKED ISSUE #1166: Expand customerGroup in product export
### Support reference number

https://jira.commercetools.com/browse/SUPPORT-4933

### Description

Currently, we only expand on `channels` inside the product exporter. We should also expand and resolve `customerGroups`

daern91:
#### Summary

PR adds missing functionality to include the customerGroup when exporting product prices

#### Description

For some reason, this was never implemented. PR should be straight-forward as it's only extending parts of the current code.

#### Todo

- Tests
  - [x] Unit
  - [x] Integration

closes #1166

### DIFF
--- a/integration-tests/cli/product-json-to-csv.it.js
+++ b/integration-tests/cli/product-json-to-csv.it.js
@@ -13,6 +13,7 @@ import {
   sampleParentCategory,
   sampleCategory,
   samplePriceChannel,
+  samplePriceCustomerGroup,
   createProducts,
 } from './helpers/product-json2csv.data'
 import { createData, clearData } from './helpers/utils'
@@ -31,6 +32,7 @@ async function cleanup(apiConfig) {
     clearData(apiConfig, 'states'),
     clearData(apiConfig, 'taxCategories'),
     clearData(apiConfig, 'channels'),
+    clearData(apiConfig, 'customerGroups'),
   ])
 }
 
@@ -61,6 +63,9 @@ describe('CSV and CLI Tests', () => {
     const priceChannel = await createData(apiConfig, 'channels', [
       samplePriceChannel,
     ])
+    const priceCustomerGroup = await createData(apiConfig, 'customerGroups', [
+      samplePriceCustomerGroup,
+    ])
     const state = await createData(apiConfig, 'states', [sampleState])
     const taxCategory = await createData(apiConfig, 'taxCategories', [
       sampleTaxCategory,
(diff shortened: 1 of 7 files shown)


packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -263,13 +263,17 @@ export default class ProductMapping {
 
   static _mapPriceToString(price: Price): string {
     // Full price:
-    // 'country-currencyCode centAmount|discounted.centAmount customerGroup.name#channel.key$validFrom~validUntil'
+    // 'country-currencyCode centAmount|discounted.centAmount@customerGroup.name#channel.key$validFrom~validUntil'
```
daern91: The only thing that might be controversial is this little change. `oneLineTrim` would remove the empty space between these two as it's written on a new line. 

However, intuitively I think `@` is the right choice here.

packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -263,13 +263,17 @@ export default class ProductMapping {
...
 
     return oneLineTrim`
       ${price.country ? `${price.country}-` : ''}
       ${price.value.currencyCode} ${price.value.centAmount}
       ${price.discounted ? `|${price.discounted.value.centAmount}` : ''}
-      ${`${price.customerGroup?.name ?? ''}`}
```
daern91: Funny that this part was even here since we never expanded on it in the code before 🤔

integration-tests/cli/product-json-to-csv.it.js
```diff
@@ -80,6 +85,10 @@ describe('CSV and CLI Tests', () => {
       typeId: 'channel',
       id: priceChannel[0].body.id,
     }
+    sampleProducts[0].masterVariant.prices[0].customerGroup = {
```
adnasa: why wouldn't you move this logic into `createProducts` and avoid mutation (during tests)?
mutation during test could lead to harmful false positive

daern91: That's a great question. It looks like the initial code that I followed was a bit sloppy. Moved both the mutation of `channel` and `customerGroup` into this function instead
https://github.com/commercetools/nodejs/blob/b67baed9170c60674007723b3d06e6a9570e0277/integration-tests/cli/product-json-to-csv.it.js#L83

in this commit: https://github.com/commercetools/nodejs/pull/1168/commits/d18dc1ea926ebc183f1977eb8b65ddb7577d69b2

PR #1269, opened 2019-04-21
feat(sync-actions): Add a support for changeAttributeOrderByName action

### LINKED ISSUE #1252: [Sync-actions] ProductTypes changeAttributeOrderByName
### Description

In the HTTP API it is possible to update the order of attributes on ProductTypes through the _changeAttributeOrderByName_ update action.  This order is important for customers using the Merchant Center.

Currently the Sync-Actions library does not support this update.

### Expected Behavior

```js
before = {
  id: "00000000-0000-0000-0000-000000000001",
  attributes: [
    {
      name: "color"
    },
    {
      name: "material"
    }
  ]
}
now = {
  id: "00000000-0000-0000-0000-000000000001",
  "attributes": [
    {
      name: "material"
    },
    {
      name: "color"
    }
  ]
}
updateActions = productTypesSync.buildActions(now, before)

expect(updateActions).toEqual([
  {
    action: 'changeAttributeOrderByName',
    attributeNames : ['material', 'color'],
  },
])
```

### Current Behavior

No update actions are provided.

### Context

This is a direct request from a client.  They are storing ProductType configuration within source control and have deployment scripts for changes.  They are leveraging the Sync-Actions library for determining updates and have asked if we could implement this feature.

junajan:
#### Summary

This PR adds support for `changeAttributeOrderByName` update action on productType sync.
Fixes #1252

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation


packages/sync-actions/src/product-types-actions.js
```diff
@@ -256,11 +256,56 @@ const generateUpdateActionsForAttributeEnumValues = (
...
+    .map(action => action.attribute.name)
+
+  // changeAttributeOrder action will be sent to CTP API as the last action so we have to
+  // calculate how the productType will look like after adding/removing attributes
+  const patchedOrderOld = attrsOld
+    .map(attribute => attribute.name)
+    .filter(name => !removedAttributeNames.includes(name))
+    .concat(addedAttributeNames)
+
+  if (orderNew.join(',') !== patchedOrderOld.join(','))
```
adnasa: sorry, but isn't this the same as `orderNew.join() !=== attrsOld.map(a => a.name).join()`?

packages/sync-actions/src/product-types-actions.js
```diff
@@ -256,11 +256,56 @@ const generateUpdateActionsForAttributeEnumValues = (
...
-  ),
-]
+const generateChangeAttributeOrderAction = (
+  attrsOld = [],
+  attrsNew = [],
+  updateActions = []
+) => {
+  if (!attrsOld.length || !attrsNew.length) return null
+
+  const orderNew = attrsNew.map(attribute => attribute.name)
```
adnasa: Let's try to find a better naming.. One could easily get confuse this with `Order` resource type

packages/sync-actions/src/product-types-actions.js
junajan: Hi @adnasa .. thx for the comment. Let's assume, that user provided new attributes:
```
attr1, attr2, attr4
```
And the productType saved on API has following attributes in this order:
```
attr1, attr2, attr3
```
The productType sync would generate actions `remove attr3, add attr4` so the order which we should compare is not `attrsOld.map(a => a.name).join()` (alias: `attr1, attr2, attr3`) but patched productType with the new order `attr1, attr2, attr4` (because API will also perform actions for adding and removing attributes and then it will compare order).

We have to do this check because if we send `changeOrder` actions with the same order attributes currently have, API will throw an error.

packages/sync-actions/src/product-types-actions.js
junajan: Updated here 04415f14

packages/sync-actions/src/product-types-actions.js
adnasa: Thanks for clarifying. This sounds like a good candidate for a code comment ☝🏽
because there is no way I will remember this in the future

PR #1345, opened 2019-06-12
feat(sync-actions): add sync actions for category assets

katmatt:
#### Summary

This PR adds support for synching category assets to our `synch-actions` module.

#### Description

This implementation is the simplest solution for synching category assets and follows the approach we use for product variants:
- we only use `addAsset` and `removeAsset` update action instead of using the more fine grained asset update actions (e.g. `changeAssetName`)

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/sync-actions/src/assets-actions.js
+++ b/packages/sync-actions/src/assets-actions.js
@@ -26,32 +26,20 @@ export default function actionsMapAssets(diff) {
 
   if (assets && assets._t === 'a') {
     const keys = Object.keys(assets)
-    const changeDeltas = keys
-      .filter(key => REGEX_NUMBER.test(key))
-      .map(key => assets[key])
-    const addAssetActions = changeDeltas.map(createAddAssetAction)
-    assetsActions.push(addAssetActions)
 
     const removeDeltas = keys
       .filter(key => REGEX_UNDERSCORE_NUMBER.test(key))
       .map(key => assets[key])
     const removeAssetActions = removeDeltas.map(createRemoveAssetAction)
     assetsActions.push(removeAssetActions)
-  } else if (Array.isArray(assets)) {
-    const assetActions = diff.assets.map(assetDiff => {
-      let assetAction
-      if (Array.isArray(assetDiff)) {
-        const asset = assetDiff[0]
 
-        if (assetDiff.length === 1) {
-          assetAction = {
-            action: 'addAsset',
-            asset,
-          }
-        }
-      }
-      return assetAction
-    })
+    const changeDeltas = keys
+      .filter(key => REGEX_NUMBER.test(key))
+      .map(key => assets[key])
+    const addAssetActions = changeDeltas.map(createAddAssetAction)
+    assetsActions.push(addAssetActions)
+  } else if (Array.isArray(assets)) {
+    const assetActions = diff.assets.map(createAddAssetAction)
     assetsActions.push(assetActions)
   }
   return flatten(assetsActions)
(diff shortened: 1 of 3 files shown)


daern91:
Hi @adnasa and @emmenko, since I'm on vacation, could you please give this one a look? 🙏 

Thanks a lot

emmenko:
Sure, I'll take a look at it tomorrow

katmatt:
@daern91 It's not very urgent, that's why I assigned it to you 😉  But I would of course appreciate getting a code review from @emmenko too 😄

packages/sync-actions/src/assets-actions.js
```diff
@@ -0,0 +1,46 @@
...
+  return action
+}
+
+function createAddAssetAction(delta) {
+  const asset = delta[0]
+  const action = { action: 'addAsset', asset }
+  return action
+}
+
+export default function actionsMapAssets(diff) {
```
tdeekens: Maybe I misread something but is there a reason we don't do it as usually? https://github.com/commercetools/nodejs/blob/master/packages/sync-actions/src/zones-actions.js#L26-L59

packages/sync-actions/src/assets-actions.js
```diff
@@ -0,0 +1,46 @@
+import flatten from 'lodash.flatten'
+
+const REGEX_NUMBER = new RegExp(/^\d+$/)
+const REGEX_UNDERSCORE_NUMBER = new RegExp(/^_\d+$/)
+
+function createRemoveAssetAction(delta) {
+  const assetIdentifier = asset =>
+    asset.id ? { assetId: asset.id } : { assetKey: asset.key }
+  const action = Object.assign(
```
tdeekens: Nit: could just spread I suppose.

katmatt:
@tdeekens TBH: I didn't understand how the "normal" way works. Thx for pointing me in the correct direction and I will change my code accordingly.

katmatt:
@tdeekens Thx for your help! I adapted the code to your comments and am looking forward to your review :-)

packages/sync-actions/src/assets-actions.js
```diff
@@ -0,0 +1,41 @@
+import createBuildArrayActions, {
+  ADD_ACTIONS,
+  REMOVE_ACTIONS,
+  CHANGE_ACTIONS,
+} from './utils/create-build-array-actions'
+
+function toAssetIdentifier(asset) {
```
tdeekens: Extracting this like this is nice.

packages/sync-actions/src/assets-actions.js
```diff
@@ -0,0 +1,41 @@
...
+      action: 'addAsset',
+      asset: newAsset,
+    }),
+    [REMOVE_ACTIONS]: oldAsset => ({
+      action: 'removeAsset',
+      ...toAssetIdentifier(oldAsset),
+    }),
+    [CHANGE_ACTIONS]: (oldAsset, newAsset) => {
+      const result = []
+      result.push({
```
tdeekens: Maybe adding a comment why we don't use `changeAsset` etc could help the next person to not wonder why?

packages/sync-actions/src/assets-actions.js
```diff
@@ -0,0 +1,41 @@
...
+        action: 'removeAsset',
+        ...toAssetIdentifier(oldAsset),
+      })
+
+      result.push({
```
tdeekens: I don't really know why we `push` here twice though. Couldn't we just `return [..., ...]`?

tdeekens:
:shipit:

PR #1350, opened 2019-06-24
fix: add isEmailVerified action to customer sync actions

rycu:
#### Summary

Simply adds isEmailVerified action to customer-actions.js & customer-sync.spec.js

#### Description

Just as https://github.com/commercetools/nodejs/pull/1242 added locale to customer sync actions; this pr adds isEmailVerified for the sync-actions for customers.

- Tests
  - [x] Unit


rycu:
okay, that makes sense.
In our scenario we can set emailVerified as true when we create a customer
👍

PR #1371, opened 2019-07-15
fix(req-builder): enable setting fuzzyLevel to 0

### LINKED ISSUE #1348: [api-request-builder]: Allow user to set fuzzyLevel to  0
### Description

fuzzyLevel checks only against falsy values, that why it throws error on 0 as
argument

### Expected Behavior

```js
fuzzyLevel(0)
```

should reduce fuzziness to 1 or 2 characters
[0 - for texts of one or two characters](https://docs.commercetools.com/http-api-projects-products-search#fuzzy-search
### Current Behavior)

but throws error: 'Required argument for `fuzzyLevel` is missing'

### Possible Solution

- include `0` in boolean expression of  [argument validation](https://github.com/commercetools/nodejs/blob/b308655e923f2576f3fab34703318537ed0eac38/packages/api-request-builder/src/query-search.js#L38)

daern91:
#### Summary

PR enables setting `fuzzyLevel` to 0.

#### Description

We were checking for falsey values in the conditional. And since `0` evaluates to falsey it was not possible to use it. More info in the referenced issue below.

#### Todo

- Tests
  - [x] Unit

resolves #1348

### DIFF
--- a/packages/api-request-builder/src/query-search.js
+++ b/packages/api-request-builder/src/query-search.js
@@ -35,7 +35,7 @@ export function fuzzy(): Object {
  * @return {Object} The instance of the service, can be chained.
  */
 export function fuzzyLevel(value: number): Object {
-  if (!value && value !== 0)
+  if (value === undefined || value === null)
     throw new Error('Required argument for `fuzzyLevel` is missing')
   this.params.search.fuzzyLevel = value
   return this
--- a/packages/api-request-builder/src/query-search.js
+++ b/packages/api-request-builder/src/query-search.js
@@ -35,7 +35,8 @@ export function fuzzy(): Object {
  * @return {Object} The instance of the service, can be chained.
  */
 export function fuzzyLevel(value: number): Object {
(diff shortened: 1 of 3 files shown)


packages/api-request-builder/src/query-search.js
```diff
@@ -35,7 +35,8 @@ export function fuzzy(): Object {
  * @return {Object} The instance of the service, can be chained.
  */
 export function fuzzyLevel(value: number): Object {
-  if (!value) throw new Error('Required argument for `fuzzyLevel` is missing')
+  if (!value && value !== 0)
```
acbeni: since we're evaluating two cases anyways, i would suggest to be more explicit by using 
```suggestion
  if (variable === undefined || variable === null)
```

daern91: Sounds good to me 👍 Fixed here: https://github.com/commercetools/nodejs/pull/1371/commits/087f49d7636bac359e6c45d99584d6a72ce661d4

PR #1373, opened 2019-07-16
fix(sync-actions): custom field with long text (#1357)

### LINKED ISSUE #1357: Patching a channel custom field diff is causing an error
### Support reference number

https://jira.commercetools.com/browse/PCM-1310

### Description

Patching a channel text custom field is causing an error.

Field old value:

```
word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word word 1234
```

Field new value:

```
word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word wordword word word word word word word word word word word word word word word word word word word word 1234 1
```

The old value is updated by adding `1` at the end. This will invoke the following code

https://github.com/commercetools/nodejs/blob/b533e7caeb66eeed1dc2e5bd41a91d31eaf083a7/packages/sync-actions/src/utils/diffpatcher.js#L39-L57

with `arr` value as:

```js
[
  `@@ -991,8 +991,10 @@
   ord 1234
  + 1
  `,
  0,
  2
];
```
Executed example: https://codesandbox.io/embed/commercetoolsnodejs1357-r1wp8

### Expected Behavior

The patching should be done without error.

Not sure though if this is a bug or the expected behavior as `getDeltaValue` is invoked without the `originalObject` parameter.

https://github.com/commercetools/nodejs/blob/fd66b0e709836e9c460f49c6f644feb3245dd17c/packages/sync-actions/src/utils/action-map-custom.js#L27-L30

### Current Behavior

```
Error: Cannot apply patch to long text diff. Missing original object.
```

### Context

This is causing an [issue](https://jira.commercetools.com/browse/PCM-1310) in the mc.

Kapture 2019-07-03 at 14 05 18

### Possible Solution

Passing `originalObject` parameter?

katmatt:
#### Summary

This PR fixes #1357 

#### Description

This PR fixes the problem by ignoring the diffed value. This is possible since our code has access to the new object and then we can get the new value simply from this object. But I'm not 100% sure that my assumptions are correct and I hope that we have enough tests 😉 

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -1,8 +1,12 @@
 ### TODOs
-| Filename | line # | TODO
-|:------|:------:|:------
-| packages/api-request-builder/src/create-service.js | 134 | this can lead to invalid URIs as getIdOrKey can return
-| packages/custom-objects-importer/src/main.js | 127 | remove `FlowFixMe` when [this](https://github.com/facebook/flow/issues/5294) issue is fixed
-| packages/product-json-to-csv/test/writer.spec.js | 263 | the "unzip" package fires finish event before entry events
-| packages/product-json-to-csv/test/writer.spec.js | 308 | the "unzip" package fires finish event before entry events
-| packages/sync-actions/src/utils/diffpatcher.js | 3 | create an issue here https://github.com/benjamine/jsondiffpatch/issues/new
+
+| Filename                                                                                     | line # | TODO                                                                                        |
+| :------------------------------------------------------------------------------------------- | :----: | :------------------------------------------------------------------------------------------ |
+| packages/api-request-builder/src/create-service.js                                           |  134   | this can lead to invalid URIs as getIdOrKey can return                                      |
+| packages/custom-objects-importer/src/main.js                                                 |  127   | remove `FlowFixMe` when [this](https://github.com/facebook/flow/issues/5294) issue is fixed |
+| packages/product-json-to-csv/test/writer.spec.js                                             |  263   | the "unzip" package fires finish event before entry events                                  |
+| packages/product-json-to-csv/test/writer.spec.js                                             |  308   | the "unzip" package fires finish event before entry events                                  |
+| packages/sync-actions/src/utils/diffpatcher.js                                               |   3    | create an issue here https://github.com/benjamine/jsondiffpatch/issues/new                  |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/custom-objects-importer/src/main.js     |  127   | remove `FlowFixMe` when [this](https://github.com/facebook/flow/issues/5294) issue is fixed |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/product-json-to-csv/test/writer.spec.js |  263   | the "unzip" package fires finish event before entry events                                  |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/product-json-to-csv/test/writer.spec.js |  308   | the "unzip" package fires finish event before entry events                                  |
--- a/TODOS.md
+++ b/TODOS.md
@@ -6,6 +6,3 @@
 | packages/product-json-to-csv/test/writer.spec.js | 263 | the "unzip" package fires finish event before entry events
 | packages/product-json-to-csv/test/writer.spec.js | 308 | the "unzip" package fires finish event before entry events
 | packages/sync-actions/src/utils/diffpatcher.js | 3 | create an issue here https://github.com/benjamine/jsondiffpatch/issues/new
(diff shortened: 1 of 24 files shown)


packages/sync-actions/src/utils/action-map-custom.js
```diff
@@ -24,50 +24,43 @@ const extractTypeFields = (diffedFields, nextFields) =>
   Array.isArray(diffedFields)
     ? diffpatcher.getDeltaValue(diffedFields)
     : nextFields
-const extractFieldValue = (diffedFields, nextFields, fieldName) =>
-  Array.isArray(diffedFields[fieldName])
-    ? diffpatcher.getDeltaValue(diffedFields[fieldName])
-    : nextFields[fieldName]
+const extractFieldValue = (newFields, fieldName) => newFields[fieldName]
```
tdeekens: So this in essence is the change. The rest is a bit of renaming?

daern91: Yeah

tdeekens: So what does the diffpatcher return if it is an array? Is is something else than the `newFields[fieldName]`? Did we check that?

katmatt: The main change is that we don't call `diffpatcher.getDeltaValue()` anymore, because we already now that we changed a text/string field. And `diffpatcher.getDeltaValue()` tries to patch the old field with the delta, which for me doesn't make a lot of sense if you already have the `newFields[fieldName]. Why should I then complicate things by applying a patch to the old value?

katmatt: @tdeekens I will revert the changes to our `yarn.lock` and `TODOS.md`, looks like I didn't check my changes properly before opening this PR 😞

tdeekens: > And diffpatcher.getDeltaValue() tries to patch the old field with the delta, which for me doesn't make a lot of sense if you already have the `newFields[fieldName].

To me neither. I was just wondering if there is an expectation towards that somewhere. But I also can't recall. I guess we have to try :)

TODOS.md
```diff
@@ -1,11 +1,12 @@
...
+| packages/product-json-to-csv/test/writer.spec.js                                             |  308   | the "unzip" package fires finish event before entry events                                  |
+| packages/sync-actions/src/utils/diffpatcher.js                                               |   3    | create an issue here https://github.com/benjamine/jsondiffpatch/issues/new                  |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/custom-objects-importer/src/main.js     |  127   | remove `FlowFixMe` when [this](https://github.com/facebook/flow/issues/5294) issue is fixed |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/product-json-to-csv/test/writer.spec.js |  263   | the "unzip" package fires finish event before entry events                                  |
+| /Users/deriksson/Sites/commercetools/nodejs/packages/product-json-to-csv/test/writer.spec.js |  308   | the "unzip" package fires finish event before entry events                                  |
```
adnasa: should we keep `Users/<name>` info in this list? 😆

packages/sync-actions/src/utils/action-map-custom.js
adnasa: a code comment outlining that `fields` of `custom` is not an array would be helpful and that they are of a different shape to `fieldDefinitions` (which is an array)

TODOS.md
daern91: hehe... nvm this one. Looks like my integration of `leasot` some time ago doesn't work properly 😇

katmatt: Good catch and I'm fine with fixing @daern91 issues 😄

katmatt:
@adnasa I will add a comment👍

katmatt:
@adnasa I see: 👎 means you approve it 😄

PR #1374, opened 2019-07-16
chore: automate changelog building

daern91:
#### Summary

This is insane. Please do not merge before discussing.

#### Description

The changelog had not been updated properly since 2017, so we added some stuff manually but now need to make sure it does not happen again by automating it.

In this PR I added `lerna-changelog` but did not figure out a proper way to set it up. So there are some insane shell hacks included.

In addition, `lerna-changelog` builds it a bit ugly whenever there are interdependencies within the monorepo. Please see the perfect example in the newly built one in the PR. This is of course something not to be fixed on our end... but still. Would love to get your comments on it.

Adding a couple of you that I know have experience with this for discussion.

Thanks!


package.json
```diff
@@ -13,6 +13,8 @@
     "precommit": "lint-staged",
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "commit": "git-cz",
+    "build:changelog": "lerna-changelog --from=$(git log --grep='chore(release): releasing component' --format='%H' | sed -n 2p) > tmp && cat CHANGELOG.md >> tmp && cat tmp > CHANGELOG.md && rm tmp",
```
daern91: Injecting crazium.

`lerna-changelog` includes the `--from` flag which also works with commit ID's. So, in order to fetch all PR's between the last release and current one we do the following...

1. Fetch all git log ID's that include the msg `chore(release): releasing component` which we use when releasing through `lerna`.
2. Select the 2nd line, as the first one is obviously the current release.
3. Appending this directly to the changelog would make it show up at the bottom of the file. So instead, we put it into a `tmp file` and then append the `changelog` to that file instead. Then finally overwrite the `changelog` with the result and deleting the `tmp file`

75% crazium injected

package.json
```diff
@@ -13,6 +13,8 @@
...
+    "push:changelog": "git add CHANGELOG.md && git commit -m 'docs: update changelog' && git push",
```
daern91: then we git add the changelog, commit it and push it.

package.json
```diff
@@ -26,7 +28,7 @@
     "test:coverage": "npm run build && npm test -- --no-watchman --coverage",
     "test:ci": "npm run test -- --ci --no-watchman --maxWorkers=10",
     "test:coverage:ci": "npm run test:ci -- --coverage && codecov",
-    "release": "cross-env NODE_ENV=cli npm run build && lerna publish --registry=https://registry.npmjs.org/ && npm run docs:publish",
+    "release": "cross-env NODE_ENV=cli npm run build && lerna publish --registry=https://registry.npmjs.org/ && build:changelog && push:changelog && npm run docs:publish",
```
daern91: finally, we put it all into the release pipeline.

100% crazium injected.

CHANGELOG.md
```diff
@@ -1,3 +1,190 @@
...
+
+#### Committers: 1
+
+- Daniel Eriksson ([@daern91](https://github.com/daern91))
+
+## @commercetools/api-request-builder@5.3.2 (2019-07-16)
+
+#### :bug: Bug Fix
+
+- `api-request-builder`
```
daern91: Almost 200 lines of docs just because of one update to the `api-req-builder` 🐼

package.json
tdeekens: Could also just be `post:build:changelog`?

package.json
tdeekens: I don't follow why we can just only run `lerna-changelog`?

CHANGELOG.md
tdeekens: 1. I would always strip committers. We do that manually actually.
2. I am not sure why it creates so many entries even though the package is used among all other packages as a dep.

package.json
daern91: Good question. I'll try to explain below.

> By default lerna-changelog will show all pull requests that have been merged since the latest tagged commit in the repository
([ref](https://github.com/lerna/lerna-changelog#usage))

So, running `lerna-changelog` without any args **after** a release will not print out anything since no PRs has been made. 

On the other hand, running it **before** releasing would include the PRs, but put them under `## Unreleased` and thus, not include the versioning.

Exampe attached from docs below:
```
## Unreleased (2018-05-24)

#### :bug: Bug Fix
* [#198](https://github.com/my-org/my-repo/pull/198) Avoid an infinite loop ([@helpful-hacker](https://github.com/helpful-hacker))

#### :house: Internal
* [#183](https://github.com/my-org/my-repo/pull/183) Standardize error messages ([@careful-coder](https://github.com/careful-coder))

#### Commiters: 2
- Helpful Hacker ([@helpful-hacker](https://github.com/helpful-hacker))
- [@careful-coder](https://github.com/careful-coder)
```

CHANGELOG.md
daern91: 1. _Manually_ - we are clearly not thinking lazily enough 👯‍♂ 
2. Yaa, although, it makes sense as all those packages get bumped when they depend on it. But structurally it would make so much more sense to just list all the packages that are affected by the same change IMO 😕

package.json
daern91: Putting them all into one script?

package.json
tdeekens: We run it _before_ a release. So the process is

1. Generate changelog
2. Commit to release branch
3. Merge PR
4. Release

This ensures that the `CHANGELOG.md` is actually part of the npm package. More [here](https://github.com/commercetools/merchant-center-application-kit/blob/master/CONTRIBUTING.md#cutting-a-release).

daern91: Hmm very interesting. And lerna then updates the `.md` upon the actual release? Or do you do that manually after?

tdeekens: Manually. We copy and paste it in. For us it works just fine. We don‘t want to release three times a day. So it‘s a 80/20 for how much automation we need.

daern91: Ahh got it.. thanks a lot for sharing about your process on it.

Maybe having it fully automated is just a pipe dream of mine. Will sit down and rethink...

cc @katmatt

daern91:
Will do!

Will also add a follow-up issue to use tech time to discuss with `lerna-changelog` if we can improve this. Then potentially open a PR there. Design is the only thing that keeps bugging me...

daern91:
@tdeekens, added docs and variable check here: https://github.com/commercetools/nodejs/pull/1374/commits/104112840ec2b3903fb13bf71a1f339cd381ff90

Also added an issue for discussing with `lerna-changelog` here: #1375 
Will bring it up in our next planning and try to carve out some time.

PR #1379, opened 2019-07-22
fix(product-json-to-csv): split requests with too large predicate

### LINKED ISSUE #1377: product-json-to-csv async _getChannelsById creates greater than 15k request and is blocked by google with a 413 error
### Support reference number
Support-6235

### Description
product export calling product-json-to-csv with more than 15k worth of active price channel ids fails to reach the api due to the sizeof the request.

### Expected Behavior
No export size limits and  No error

### Current Behavior
User receives  >Error 413 (Request Entity Too Large)

### Context
large projects are unable to export product in one export. Workaround to spiit export with where predicate is not practical.

### Possible Solution
price channel cache  need to be changed to build in multiple calls to channel api.

daern91:
#### Summary

PR adds batching to `get requests` with `channel` and `custumerGroup` predicates.

Previously, we were just stacking them all up which could lead to requests with hundreds of IDs. These requests would then be denied by our platform load balancer which is not configurable. 

resolves #1377

### DIFF
--- a/packages/product-json-to-csv/src/main.js
+++ b/packages/product-json-to-csv/src/main.js
@@ -26,7 +26,7 @@ import highland from 'highland'
 import fetch from 'node-fetch'
 import Promise from 'bluebird'
 import JSONStream from 'JSONStream'
-import { memoize, map, get, pick } from 'lodash'
+import { memoize, map, get, pick, chunk } from 'lodash'
 import ProductMapping from './map-product-data'
 import { writeToSingleCsvFile, writeToZipFile } from './writer'
 import pkg from '../package.json'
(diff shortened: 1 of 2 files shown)


packages/product-json-to-csv/src/main.js
```diff
@@ -217,12 +217,34 @@ export default class ProductJsonToCsv {
 
     if (!channelIds.length && !customerGroupIds.length) return prices
 
+    const CHUNK_SIZE = 150
     // resolve channels by their ids
-    const channelsById: Object = await this._getChannelsById(channelIds)
+    let channelsById: Object = {}
+    if (channelIds.length > CHUNK_SIZE) {
+      channelsById = await chunk(channelIds, CHUNK_SIZE).reduce(
```
adnasa: I would just move this into the `_get*` so that the logic is a bit "hidden" for the `resolvePriceReferencesThingy`

PR #1393, opened 2019-09-09
refactor: migrate user-agent packages to TS

emmenko:
Setup PR:
- #1394 
- #1395 
- #1396 
- #1397 (_current base branch_)

For now I migrated 2 small packages:
- `http-user-agent`
- `sdk-middleware-user-agent`

I also added a new package `sdk-types`, which contains shared TS types (e.g. middleware).


emmenko:
FYI: I will rebase this branch once #1395 has been merged, so that it only contains the packages migration.

packages/api-request-builder/src/query-id.js
```diff
@@ -6,7 +6,7 @@
  * @throws If `id` is missing.
  * @return {Object} The instance of the service, can be chained.
  */
-export function byId(id: string): Object {
+export function byId(id: string): { [key: string]: any } {
```
acbeni: here i would assume the return types is a request, i think if that's the case it would make sense to point out Request as a return type.

emmenko: Not really, `Object` is actually just an alias for `any` in Flow.

Don't worry much about this, when we migrate the packages to TS, we will fix all the current "broken" types.

packages/sdk-types/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/* Middlewares */
+export type MiddlewareRequest = ClientRequest
+export type MiddlewareResponse = {
+  resolve(response: JsonObject): void
+  reject(error: JsonObject): void
+  body?: JsonObject
+  error?: HttpErrorType
+  statusCode: number
+  headers?: JsonObject<string>
```
acbeni: headers should be a map `{(key:string): string}`

acbeni:
the incremental approach sounds good for now just one additional remark for the header and i think we re good to go 🎖

.prettierrc
```diff
@@ -1,5 +1,14 @@
...
   "trailingComma": "es5",
-  "singleQuote": true
-}
+  "singleQuote": true,
+  "parser": "babel-flow",
+  "overrides": [
+    {
+      "files": "*.ts",
+      "options": {
+        "parser": "typescript"
```
emmenko: Here we use the `override` feature to tell prettier to use `typescript` parser for the ts files, otherwise the `babel-flow` parser.

packages/http-user-agent/.eslintrc.js
```diff
@@ -0,0 +1,41 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
```
emmenko: As mentioned before, for now TS packages should have their own eslint config, which is merged with the base one.

Once we've migrated all packages, we can merge this back with the base one.

emmenko:
Alright, this is now ready for review.

/cc @tdeekens for awareness

packages/csv-parser-state/src/main.js
```diff
@@ -125,17 +125,19 @@ export default class CsvParserState {
...
               typeId: 'state',
               id: response.body.id,
             })
           )
           resolve({ ...remainingState, transitions: stateReferences })
-        }
-      )
+        })
+        .catch((error: Error) => {
+          reject(error)
```
emmenko: This was missing. Somehow the tests caught this now

packages/csv-parser-state/test/main.spec.js
```diff
@@ -125,12 +125,14 @@ describe('CsvParserState', () => {
...
   describe('::_transformTransitions', () => {
     describe('without state transitions', () => {
-      test('should not mutate state', async () => {
+      test('should not mutate state', () => {
         const actual = {
           foo: 'bar',
           key: 'my-state',
         }
-        expect(csvParser._transformTransitions(actual)).resolves.toEqual(actual)
+        return expect(csvParser._transformTransitions(actual)).resolves.toEqual(
```
emmenko: There were a bunch of tests that did not return the expectation with the promise.

emmenko:
Alright, CI is green now. I'll leave this open for the rest of the day and merge it tomorrow. Then we can continue migrating packages step-by-step.

packages/typescript-sdk/src/gen/client/cart-discounts/ByProjectKeyCartDiscountsRequestBuilder.ts
```diff
@@ -1,10 +1,10 @@
 import { ByProjectKeyCartDiscountsByIDRequestBuilder } from './ByProjectKeyCartDiscountsByIDRequestBuilder'
 import { ByProjectKeyCartDiscountsKeyByKeyRequestBuilder } from './ByProjectKeyCartDiscountsKeyByKeyRequestBuilder'
-import { CartDiscountDraft } from './../../models/CartDiscount'
-import { CartDiscountPagedQueryResponse } from './../../models/CartDiscount'
-import { CartDiscount } from './../../models/CartDiscount'
-import { Middleware } from './../../base/common-types'
-import { ApiRequest } from './../../base/requests-utils'
```
daern91: @acbeni fixing these looks like something we should do in the generator :-)

PR #1394, opened 2019-09-09
chore: fix flow setup

emmenko:
Extracted changes/fixes noticed in #1393 

Also, updated flow to latest version `0.107`


emmenko:
Merging, as I need those changes for another PR.

PR #1395, opened 2019-09-09
chore: ts setup, fix ts-sdk issues

emmenko:
Extract some of the setup changes from #1393 

Now the ts config is for the all project and not in the ts-sdk anymore.

Additionally run prettier in the ts-sdk (hence most of the changes in this PR).


packages/typescript-sdk/.eslintrc.js
```diff
@@ -0,0 +1,46 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
```
emmenko: As long as we have both flow and ts, we need to have "custom" eslint configs in the TS packages. Eventually, we will have only one config once flow has been removed.

packages/typescript-sdk/babel.config.js
```diff
@@ -1,41 +0,0 @@
-const getPresets = () => {
```
emmenko: Not needed anymore, it's now included in the main babel config as `overrides`

packages/typescript-sdk/src/gen/base/common-types.ts
```diff
@@ -1,33 +1,35 @@
 export type MethodType =
-  | "GET"
-  | "HEAD"
-  | "POST"
-  | "PUT"
-  | "DELETE"
-  | "CONNECT"
-  | "OPTIONS"
-  | "TRACE";
+  | 'GET'
```
emmenko: From here on is basically all prettier changes 👇

PR #1397, opened 2019-09-09
chore: configure rollup to build multiple outputs

emmenko:
Instead of having one build command for each output format, we can include them in rollup.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,6 @@
+// Do this as the first thing so that any code reading it knows the right env.
+process.env.BUILD_ROLLUP = true
+
 /* eslint-disable */
 const resolve = require('rollup-plugin-node-resolve')
 const commonjs = require('rollup-plugin-commonjs')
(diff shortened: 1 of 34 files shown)


emmenko:
Sure, sorry for the poor description.

The main reason is mostly to keep the setup simple and consistent throughout the packages. Also, since we run `yarn build` each time we install/update the deps, I wanted to make the process faster and I feel like this approach does improve things.

> I am not sure I should be reviewing this.

Yeah, I just added you since you are also familiar with the rollup setup. I'll also consider if we should use the same in appkit...

emmenko:
**Before**

```
lerna success run Ran npm script 'build' in 32 packages in 43.2s:
```

**After**

```
lerna success run Ran npm script 'build' in 32 packages in 30.6s:
```

rollup.config.js
```diff
@@ -12,37 +15,66 @@ const babelConfig = require('./babel.config')
...
 
-const config = {
-  output: {
-    sourcemap: true,
+const createConfig = cliArgs => [
+  {
+    input: cliArgs.input,
+    output: {
+      format: 'umd',
+      file: `dist/${packageName}.umd.js`,
```
tdeekens: Are we only building one UMD bundle now?

emmenko: Look below 😊
The second one is minified

tdeekens: 🤦‍♂

PR #1412, opened 2019-09-24
Add sync actions to stores

amine-benselim:
#### Summary

this PR adds sync actions to stores, I followed the steps of previous PRs adding sync actions, I also added the relevant documentation for it.

#### Todo

- Tests
  - [X] Unit
- [X] Documentation


packages/sync-actions/test/stores-sync.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+describe('Actions', () => {
+  let storesSync
+  beforeEach(() => {
+    storesSync = storesSyncFn()
+  })
+
+  test('should build `setName` action', () => {
+    const before = {
+      name: { locale: 'en', value: 'Algeria' },
```
katmatt: A `LocalizedString` has the format (see https://docs.commercetools.com/http-api-types.html#localizedstring):
```js
name: {
  'en': 'Algeria'
...
}
```

packages/sync-actions/test/stores-sync.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+    const now = {
+      name: [
```
adnasa: The API accepts both `name` as an array and as a single value of `{ locale, value }`.
We would want to make sure in the MC that we pass in always `array`.

Because the end user might have the intention to update only the `en` value and retain `de`. If we pass in only `{ locale, value }` (not an array), we would end up overriding the entire `nameAllLocales`

katmatt: Really? Why isn't it documented then?

packages/sync-actions/src/stores-actions.js
```diff
@@ -0,0 +1,15 @@
+import { buildBaseAttributesActions } from './utils/common-actions'
+
+export const baseActionsList = [
+  { action: 'setLanguages', key: 'languages' },
```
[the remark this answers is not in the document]
amine-benselim: yes it is an upcoming feature, it is available in the graphql schema if you are using the graphql IDE in Impex.

packages/sync-actions/test/stores-sync.spec.js
amine-benselim: one assumption is that maybe the  REST api is different from the Graphql api? 🤔 at least in graphql that is the case.

packages/sync-actions/test/stores-sync.spec.js
adnasa: This is true and `sync-actions` only deals with REST shapes so far.
We should change this, @amine-benselim

packages/sync-actions/src/stores-actions.js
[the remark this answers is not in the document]
amine-benselim: since it is not documented yet I decided to remove it. I will add it whenever  the documentation is released.

amine-benselim:
@katmatt what is the process of releasing for the SDK? should I take care of that at all? 😄

packages/sync-actions/test/stores-sync.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+describe('Actions', () => {
+  let storesSync
+  beforeEach(() => {
+    storesSync = storesSyncFn()
+  })
+
+  test('should build `setName` action', () => {
+    const before = {
+      name: [{ en: 'Algeria' }],
```
adnasa: a `LocalizedString` is never an array

amine-benselim: correct 👍 fixed!

katmatt:
@amine-benselim @adnasa May be - in the future and not as part of this PR - we should add some integration tests? Or switch to our generated typescript-sdk? Otherwise we always have to check the code against the official docs....

adnasa:
this is out of scope. but yeah maybe in the future it makes sense

amine-benselim:
@katmatt any guidelines on how to do that (maybe a previous PR to get "inspiration" from? 😅). I agree with you and I will gladly do it.

katmatt:
@amine-benselim We currently - as far as i know - don't have any integration tests in this repo. But all of our other SDKs contain a ton of integration test. You basically have to send the update actions created by the `sync-actions`module against the ct api, then fetch the data (in your case the store) from the ct api and then compare it with what you expected. But this of course requires some changes on CI server too. And that means it needs a bit more of coordination and communication.

adnasa:
I can do the release

PR #1413, opened 2019-09-25
chore(package.json): deps installed

adnasa:
#### Summary

I received these changes locally after running `yarn release` 🤷‍♂ 
I'm not entirely sure if we should keep this.

feel free to just close the PR if it is irrelevant, but I'd like to hear what is going on here

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -1,8 +1,9 @@
 ### TODOs
-| Filename | line # | TODO
-|:------|:------:|:------
-| packages/api-request-builder/src/create-service.js | 136 | this can lead to invalid URIs as getIdOrKey can return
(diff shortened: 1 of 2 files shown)


daern91:
I think they get installed by the `gitbook cli` when publishing the docs. No need to have them saved/installed for everyone so we could close it

adnasa:
@daern91  oops... I was so eager :D 
remove them?

daern91:
haha no problems. I dont mind keeping them tbh, we could just move them into the devdependencies. Probably better for onboarding anyway so its less confusing :-)

daern91:
I also see that the linter is failing now :trollface:

PR #1418, opened 2019-10-10
feat(http-middleware): custom timeout and abortcontroller options

### LINKED ISSUE #190: Allow to set optional timeout for requests to API
### Description
Allow to set custom timeout for long lasting requests

### Expected Behavior
Should accept custom request timeout

### Current Behavior
Custom timeout is not supported

daern91:
#### Summary

PR adds `timeout` and `AbortController` options to the http middleware. When using the timeout option, we opted for using the controller over the `timeout` option which is available on some fetch implementations. This decision was based on the fact that `timeout` is getting deprecated in most common fetchers, including [node-fetch](https://www.npmjs.com/package/node-fetch#options). Instead, they are following the guidelines from the [Fetch Standard](https://fetch.spec.whatwg.org/) where `timeout` does not exist.

More reading and info here https://github.com/bitinn/node-fetch/issues/523

In addition, this allows users to implement their own abort mechanisms by simply passing in the controller without using the timeout.

#### Todo

- Tests
  - [x] Unit
- [x] Documentation

resolves #190

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -3,6 +3,7 @@ const getPresets = () => {
     case 'development':
       return {
         targets: { node: 'current' },
+        useBuiltIns: 'usage',
         modules: 'commonjs',
       }
     case 'rollup':
@@ -11,6 +12,7 @@ const getPresets = () => {
           browsers: ['last 2 versions'],
           node: 'current',
         },
+        useBuiltIns: 'usage',
         modules: false,
       }
     case 'production':
@@ -19,11 +21,13 @@ const getPresets = () => {
           browsers: ['last 2 versions'],
           node: '8',
         },
+        useBuiltIns: 'usage',
         modules: false,
       }
     case 'cli':
       return {
         targets: { node: '8' },
+        useBuiltIns: 'usage',
         modules: 'commonjs',
       }
     default:
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -57,6 +57,7 @@ export default function createHttpMiddleware({
   includeOriginalRequest,
   maskSensitiveHeaderData = true,
   enableRetry,
+  signal,
   retryConfig: {
     // encourage exponential backoff to prevent spamming the server if down
     maxRetries = 10,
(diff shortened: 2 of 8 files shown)


babel.config.js
```diff
@@ -19,11 +21,13 @@ const getPresets = () => {
           browsers: ['last 2 versions'],
           node: '8',
         },
+        useBuiltIns: 'usage',
         modules: false,
       }
     case 'cli':
       return {
         targets: { node: '8' },
+        useBuiltIns: 'usage',
```
daern91: These are only needed as long as we support `Node 8` since it does not support Promise#finally.

Can be removed as soon as we move to Node 10/12 pairings.

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,115 +117,120 @@ export default function createHttpMiddleware({
...
     function executeFetch() {
       // $FlowFixMe
-      fetcher(url, fetchOptions).then(
-        (res: Response) => {
-          if (res.ok) {
-            if (fetchOptions.method === 'HEAD') {
-              next(request, {
-                ...response,
-                statusCode: res.status,
+      fetcher(url, fetchOptions)
```
daern91: only prettier noise from here to the `finally` clause

types/sdk.js
```diff
@@ -220,13 +220,24 @@ export type PasswordAuthMiddlewareOptions = {
   fetch?: (url: string, args?: Object) => Promise<any>,
 }
 
+// translation of https://dom.spec.whatwg.org/#abortcontroller
```
katmatt: This link is very helpful👍

butenkor:
@daern91 There might be a misunderstanding and this change might not be necessary. In the issue #190 I meant to set custom timeout on HTTP client as it was the case in old SDK. In current SDK one has to pass own `fetch` instance to `createHttpMiddleware` https://github.com/commercetools/nodejs/blob/master/integration-tests/cli/helpers/utils.js#L11 anyway where both timeout and AbortController can be defined already. 

Also personally I am not sure how AbortController should be more useful than timeout (besides being deprecated) as by setting custom timeout you rather set it higher to avoid to early request cancelations and not really cut the connection after timeout yourself but I might be missing some other useful cases.

PR #1419, opened 2019-10-10
feat(http-middleware): add isCommercetools error property

### LINKED ISSUE #1391: SUPPORT-6705 - Improve error logging, req/resp introspection
Hello,
when a requests fail e.g with a 500 status code I got this as error message:

```
InternalServerError
Oops. This shouldn't happen.
```
this isn't helpful.

### Expected Behavior
The error message should be much more descriptive e.g including the `correlation-id`, `request-url`, `status-code`.

Beside this fact I would like know how can I access the request/response `headers` of the failed request. The `commerceClient.execute` only returns the `body`.

```
{ body:
   { limit: 20,
     offset: 0,
     count: 5,
     total: 5,
     results: [ [Object], [Object], [Object], [Object], [Object] ] },
  statusCode: 200 }
```

### Proposal

Attach all useful informations to the error object in order to provide a descriptive log message.

daern91:
#### Summary

In order to make it easier to recognize commercetools errors, PR adds this property.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration

resolves #1391

### DIFF
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -186,6 +186,7 @@ export default function createHttpMiddleware({
               const error: HttpErrorType = createError({
                 statusCode: res.status,
                 originalRequest: request,
+                isCommercetools: true,
                 retryCount,
                 headers: parseHeaders(res.headers),
                 ...(typeof parsed === 'object'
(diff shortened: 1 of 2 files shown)


daern91:
Let's discuss this one @acbeni and @katmatt, if we go this route we are removing the generic part of the SDK. Meaning if anyone uses the client for requests outside of the CTP they would still have the `isCommercetools` property attached to eventual errors. 

Need to decide if we want to keep it generic or not.

PR #1441, opened 2019-12-05
feat(sync-actions): add setLanguages update action to stores

amine-benselim:
#### Summary

this PR adds support for `setLanguages` as an update action for stores. 

the action is available in graphql (found it in impex graphiql playground) but is not yet in the documentation so it should probably be added there too.

https://docs.commercetools.com/http-api-projects-stores#update-actions

### DIFF
--- a/packages/sync-actions/src/stores-actions.js
+++ b/packages/sync-actions/src/stores-actions.js
@@ -1,6 +1,9 @@
 import { buildBaseAttributesActions } from './utils/common-actions'
 
-export const baseActionsList = [{ action: 'setName', key: 'name' }]
+export const baseActionsList = [
+  { action: 'setName', key: 'name' },
+  { action: 'setLanguages', key: 'languages' },
+]
 
 export function actionsMapBase(diff, oldObj, newObj) {
   return buildBaseAttributesActions({
--- a/packages/sync-actions/test/stores-sync.spec.js
+++ b/packages/sync-actions/test/stores-sync.spec.js
@@ -7,7 +7,10 @@ describe('Exports', () => {
   })
 
   test('correctly define base actions list', () => {
-    expect(baseActionsList).toEqual([{ action: 'setName', key: 'name' }])
+    expect(baseActionsList).toEqual([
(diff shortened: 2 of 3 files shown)


katmatt:
@amine-benselim Don'get my wrong, but we shouldn't add undocumented update actions to our SDKs/libs. May be there is a reason that this update action isn't yet documented? I would suggest to first discuss it with the context team and if they then documented it, we should add it here. Does this sound like a plan to you too?

amine-benselim:
> @amine-benselim Don'get my wrong, but we shouldn't add undocumented update actions to our SDKs/libs. May be there is a reason that this update action isn't yet documented? I would suggest to first discuss it with the context team and if they then documented it, we should add it here. Does this sound like a plan to you too?

@katmatt that is a fair concern. I created a task to add documentation for it and will ask about it in the Context daily. 

I will update this PR with the information I get there, until then please block this PR :)

katmatt:
@amine-benselim Thanks and I'm looking forward to your update 😄

amine-benselim:
@katmatt I have raised the topic in our daily, and Yann explained that this is kind of a deadlock situation where one thing waits for another. 

**;tldr:** Yann thinks it's fine to release this before documentation.
 
Usually the SDK part comes after the documentation, but in this situation the documentation is waiting for other work on the backend (filtering out projections based on locales) to be done and then the documentation for it is going to be released. 

However, the MC part related to stores is ready and for that we need this SDK action. (I can also manually create the action inside the MC codebase as a workaround but it seems unnecessary for me).

katmatt:
@amine-benselim Thanks for the update. And you're unblocked now 😄

amine-benselim:
I don't have the rights to merge this so you can do so @katmatt or @daern91.

Also, I assume we have to create a release for the package

daern91:
I can do that. Thanks for bringing up that point, @katmatt, it's so important to go through the correct flow to avoid any eventual unnecessary breaking changes. And thanks @amine-benselim for quickly checking with the team.

I'll DM you as soon as it's released.

PR #1456, opened 2020-01-10
chore: align renovate config with other repos

emmenko:
This aligns the renovate config with what we use in other repos. Basically we get every Monday up  to 3 PRs:
- lockfile maintenance
- all deps (non major)
- all deps (major)

Example from appkit:
image

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -2,12 +2,27 @@
   "extends": [
     "config:base",
     ":pinOnlyDevDependencies",
-    ":prHourlyLimitNone",
-    ":prConcurrentLimitNone",
-    ":automergeMinor",
-    "group:allNonMajor",
-    ":automergeBranchMergeCommit",
     ":enableVulnerabilityAlerts",
+    ":automergeMinor",
     "schedule:weekly"
-  ]
+  ],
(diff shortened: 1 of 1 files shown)


.renovaterc.json
```diff
@@ -2,12 +2,24 @@
   "extends": [
     "config:base",
     ":pinOnlyDevDependencies",
-    ":prHourlyLimitNone",
-    ":prConcurrentLimitNone",
-    ":automergeMinor",
```
daern91: I think we should keep this one, as we use the [renovate-approve bot](https://github.com/renovatebot/renovate-approve-bot) to automatically approve then merge all non-major deps, and it should then also work with lock file maintenance. 

However, you guys have also experimented with another bot for auto-mergers, correct?

emmenko: > you guys have also experimented with another bot for auto-mergers

Not really, we have `kodiakhq` which we triggers an auto merge only when we assign a specific label.

> I think we should keep this one

Sure, I can add that back

daern91: I see, then everything looks great methinks. Thanks for thinking of us when aligning all the repos!

PR #1459, opened 2020-01-14
fix(api-request-builder): reset key

### LINKED ISSUE #1446: api-requestbuilder retains key param
### Description

Requestbuilder retains key param.

```js
const { createRequestBuilder } = require('@commercetools/api-request-builder');

this.ctRequestBuilder = createRequestBuilder({ projectKey: 'anykey' });

const request = this.ctRequestBuilder.categories.byKey('Test').build();

console.log(request);
// /anykey/categories/key=Test

const nextRequest = this.ctRequestBuilder.categories
  .parse({
    where: [`parent(id="test")`],
    sort: [{ by: 'orderHint', direction: 'asc' }]
  })
  .build();

console.log(nextRequest); 
// /**anykey**/categories/key=Test?where=parent(id%3D%22test%22)&sort=orderHint%20asc

```

### Expected Behavior

build() should reset all params.

### Current Behavior

Requestbuilder retains key param after build() is called

katmatt:
#### Summary

Closes #1446

#### Description

The key param is not a query parameter and it thus wasn't properly handled. I added reseting of the key param to our `setDefaultParams` function and added a test to make sure that it works.

#### Todo

- Tests
  - [ ] Unit
  - [x] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/api-request-builder/src/default-params.js
+++ b/packages/api-request-builder/src/default-params.js
@@ -68,6 +68,8 @@ export function getDefaultSearchParams(): ServiceBuilderDefaultParams {
  */
 export function setDefaultParams() {
   this.params.expand = getDefaultQueryParams().expand
+  
+  this.params.key = null
 
   if (this.features.includes(features.queryOne))
     this.params.id = getDefaultQueryParams().id
--- a/packages/api-request-builder/test/create-request-builder.spec.js
+++ b/packages/api-request-builder/test/create-request-builder.spec.js
@@ -65,8 +65,7 @@ describe('createRequestBuilder', () => {
     )
   })
 
-
-  test('calling build resets all paramas', () => {
+  test('calling build resets all params', () => {
     const requestBuilder = createRequestBuilder({
       projectKey: 'foo'
     })
--- a/packages/api-request-builder/test/create-request-builder.spec.js
+++ b/packages/api-request-builder/test/create-request-builder.spec.js
@@ -64,4 +64,17 @@ describe('createRequestBuilder', () => {
       expectedServiceKeys.concat('foo').sort()
     )
   })
+
+
+  test('calling build resets all paramas', () => {
+    const requestBuilder = createRequestBuilder({
+      projectKey: 'foo'
+    })
+    requestBuilder.categories.byKey('Test').build();
+    const nextRequest = requestBuilder.categories.parse({ where: ['bar']}).build();
+    expect(nextRequest).toEqual(
+      "/foo/categories?where=bar"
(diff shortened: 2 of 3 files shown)


packages/api-request-builder/test/create-request-builder.spec.js
```diff
@@ -64,4 +64,17 @@ describe('createRequestBuilder', () => {
       expectedServiceKeys.concat('foo').sort()
     )
   })
+
+
+  test('calling build resets all paramas', () => {
```
daern91: nitpick: typo 

~paramas~
params

packages/api-request-builder/src/default-params.js
```diff
@@ -68,6 +68,8 @@ export function getDefaultSearchParams(): ServiceBuilderDefaultParams {
  */
 export function setDefaultParams() {
   this.params.expand = getDefaultQueryParams().expand
+  
+  this.params.key = null
```
daern91: Nice 👍 

Looking through this file, I'm also wondering if we are ever resetting all other params? For example, version can be set here https://github.com/commercetools/nodejs/blob/68c34e7a126d7a82cc867e99fab553ab77827ea7/packages/api-request-builder/src/default-params.js#L215

but it's not reset inside this function, `setDefaultParams()`

daern91:
Oops, to quick to approve. Seems like some tests are failing.. :- )

katmatt:
@daern91 Test are green now 👍 Do we need another approval then?

packages/api-request-builder/test/default-params.spec.js
```diff
@@ -8,6 +8,7 @@ describe('defaultParams', () => {
     setDefaultParams.call({ features: serviceFeatures, params })
     expect(params).toEqual({
       id: null,
+      key: null,
```
daern91: Functionality wise I think this works great. I'm just wondering if it makes sense to add it here since `key` is not included inside the below function? Should it be added to both, perhaps?
https://github.com/commercetools/nodejs/blob/a1560ec8819cbb84af4896929b35dddbf70a64c0/packages/api-request-builder/src/default-params.js#L13-L34

katmatt: @daern91 I thought about that too, but technically speaking the `key` param is not a query parameter: It's part of the uri and that's why I didn't want to add it to this function 😄

daern91: Ahh, good point. Then we're ready to go :D

PR #1461, opened 2020-01-14
feat(product): add support for dimensions in image url

### LINKED ISSUE #1460: Add support for Dimensions in image url when export the products
### Support reference number
https://jira.commercetools.com/browse/SUPPORT-5263

### Description
When an export of products is performed and now images are using labels and dimensions

https://jeanscentre-static.joggroup.net/sys-master/images/ha9/hf0/9027232661534/871821258791|ProductImage|100x100

url|label|WxH;url2|label|WxH and so on...

VineetKumarKushwaha:
#### Summary
Add support for Dimensions in image url when export the products

below is example
https://jeanscentre-static.joggroup.net/sys-master/images/ha9/hf0/9027232661534/871821258791|ProductImage|100x100

closes #1460

### DIFF
--- a/packages/product-json-to-csv/src/map-product-data.js
+++ b/packages/product-json-to-csv/src/map-product-data.js
@@ -402,8 +402,8 @@ export default class ProductMapping {
   _mapImagesToString(images: Array<Image>): string {
     return images
       .map((image: Image): string => {
-        const { url, label } = image
-        return label ? `${url}|${label}` : url
+        const { url, label, dimensions = { w: 0, h: 0 } } = image
+        return `${url}|${label || ''}|${dimensions.w}x${dimensions.h}`
       })
       .join(this.multiValDel)
   }
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-^0.112.0
+^0.115.0
 
 [ignore]
 <PROJECT_ROOT>/_book/.*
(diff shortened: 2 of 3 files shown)


packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -402,8 +402,8 @@ export default class ProductMapping {
   _mapImagesToString(images: Array<Image>): string {
     return images
       .map((image: Image): string => {
-        const { url, label } = image
-        return label ? `${url}|${label}` : url
+        const { url, label, dimensions = { w: 0, h: 0 } } = image
+        return `${url}|${label || ''}|${dimensions.w}x${dimensions.h}`
```
daern91: Does the platform set `w` and `h` to `0` by default? 

Otherwise, wouldn't it make sense to have the dimensions to also default to empty strings, like the label?

VineetKumarKushwaha: When I create product from api playground  it is not allowing me to skip dimensions and even not w and h. So I thought both the things are mandatory so I have put it like this. It would be helpfull for the user to just enter number width and height not the separator which might they enter wrong. 

and yes in import they have default w and h reference here https://github.com/sphereio/sphere-node-product-csv-sync/blob/c5d2797131972189a05a0aa1764427ef56d90116/src/coffee/mapping.coffee#L468

daern91: Great answer, thanks! 👍

PR #1463, opened 2020-01-16
fix(ts-sdk): serialization of array query parameters

### LINKED ISSUE #1462: [typescript-sdk] Array query params serializes to 'undefined'
### Description

When using array query parameters, they get serialized to 'undefined'. For example:

```
    const response = await api
        .productProjections()
        .get({
            queryArgs: {
                sort: ["name.en desc", "foo"]
            }
        })
        .execute();
```

Issues a request to `https://api.sphere.io/{project}/product-projections?sort=undefined`.

### Expected Behavior

URL should be `https://api.sphere.io/{project}/product-projections?sort=name.en%20desc&sort=foo`

### Possible Solution

As suggested in https://github.com/commercetools/nodejs/issues/1454, I suggest to just use the built in `querystring` module to format this: [link](https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options).

I will open a PR with my proposed change.

lenkan:
#### Summary

This PR fixes the serialization of Array query parameters and simplifies it by using the build in `querystring.stringify` function.

#### Description

Fixes #1462 

#### Todo

I tried to find a place where tests for this is written, but couldn't find any. Do you have a plan for unit testing the typescript-sdk? It would be helpful to prevent these sort of bugs. I know this bit is generated, but that doesn't mean it cannot contain bugs.

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/packages/typescript-sdk/src/generated/base/requests-utils.ts
+++ b/packages/typescript-sdk/src/generated/base/requests-utils.ts
@@ -65,8 +65,8 @@ function reduceMiddleware(op1: Middleware, op2: Middleware): Middleware {
   }
 }
 
-function isDefined<T>(value: T | undefined): value is T {
-  return typeof value !== "undefined"
+function isDefined<T>(value: T | undefined | null): value is T {
+  return typeof value !== "undefined" && value !== null
 }
 
 function cleanObject<T extends VariableMap>(obj: T): T {
--- a/packages/typescript-sdk/src/generated/base/requests-utils.ts
+++ b/packages/typescript-sdk/src/generated/base/requests-utils.ts
@@ -1,5 +1,6 @@
-import { Middleware, MiddlewareArg, ClientResponse } from './common-types'
+import { Middleware, MiddlewareArg, ClientResponse, VariableMap } from './common-types'
 import { CommonRequest } from './local-common-types'
+import { stringify } from "querystring"
 
 export class ApiRequestExecutor {
   private middleware: Middleware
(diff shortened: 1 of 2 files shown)


acbeni:
I guess that's what happens when i push stuff without proper testing 😢 
Can you add a test case for your changes, in this case 🙏

packages/typescript-sdk/src/generated/base/requests-utils.ts
```diff
@@ -64,36 +65,39 @@ function reduceMiddleware(op1: Middleware, op2: Middleware): Middleware {
   }
 }
 
+function isDefined<T>(value: T | undefined): value is T {
+  return typeof value !== "undefined"
```
acbeni: i would add the null check here
```suggestion 
  return typeof value !== "undefined" && value !== null
```

lenkan: Fixed, thanks. I wasn't sure about this since null would in almost all cases be explicitly set.

lenkan:
> Only one small change + test case would do!

I will add some test cases.

lenkan:
@acbeni I've added some test cases covering the query params functionality under `packages/typescript-sdk/test`.

acbeni:
You are awesome @lenkan  🎉 🎉 🎉 
To test your changes i created a [branch](https://github.com/commercetools/nodejs/pull/1464) from yours just to make sure the tests passes, once they are green i will merge yours

lenkan:
@acbeni, perfect! Can you also ping me in some way when it has been published to npm?

acbeni:
@lenkan it's already done!, you can find it here  https://www.npmjs.com/package/@commercetools/typescript-sdk/v/0.0.10-alpha.0

PR #1468, opened 2020-01-20
feat(sync/customers): add `setStores`

adnasa:
#### Summary

- adds `setStores`. ref https://docs.commercetools.com/http-api-projects-customers#set-stores-beta
This is to prepare for a feature in the MC

### DIFF
--- a/packages/sync-actions/test/customer-sync.spec.js
+++ b/packages/sync-actions/test/customer-sync.spec.js
@@ -30,6 +30,10 @@ describe('Exports', () => {
         key: 'defaultShippingAddressId',
         actionKey: 'addressId',
       },
+      {
+        action: 'setStores',
+        key: 'stores',
+      },
     ])
   })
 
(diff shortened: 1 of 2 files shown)


packages/sync-actions/test/customer-sync.spec.js
```diff
@@ -417,4 +421,20 @@ describe('Actions', () => {
...
+      stores: ['canada'],
+    }
+    const now = {
+      stores: ['canada', 'usa'],
+    }
+    const actual = customerSync.buildActions(now, before)
+    expect(actual).toEqual([
+      {
+        action: 'setStores',
+        stores: ['canada', 'usa'],
```
katmatt: Are you really sure this works against our api? according to our docs (https://docs.commercetools.com/http-api-projects-customers#set-stores-beta) you have to provide a `ResourceIdentifier` (https://docs.commercetools.com/http-api-types.html#resourceidentifier) and not just a `string`...

adnasa: 2403522

katmatt:
Looks very good to me 👍

adnasa:
the lint passed. Do I need to wait for the other checks too? it doesn't look like they are running

katmatt:
@adnasa Which other checks do you mean?

adnasa:
oh well never mind... here we go 🎉

Issue #1442, opened 2019-12-09
Bug in csv-parser-price


skyriakou:
### Support reference number

https://jira.commercetools.com/browse/SUPPORT-7363

VineetKumarKushwaha:
@skyriakou  Can I start on this ?

skyriakou:
@VineetKumarKushwaha  sure.. go for it :)

PR #1473, opened 2020-01-20
fix(csv-parser-price): handle error when variant-sku key is missing/mis-spelled

VineetKumarKushwaha:
### handle error when variant-sku header key is missing/mis-spelled

closes: #1442

### DIFF
--- a/packages/csv-parser-price/src/cli.js
+++ b/packages/csv-parser-price/src/cli.js
@@ -78,8 +78,9 @@ const logError = error => {
   const errorFormatter = new PrettyError()
 
   if (npmlog.level === 'verbose')
-    process.stderr.write(`ERR: ${errorFormatter.render(error)}`)
-  else process.stderr.write(`ERR: ${error.message || error}`)
+    process.stderr.write(`ERR: ${errorFormatter.render(error)} \n`)
+  // Add new line to make it intuitive
+  else process.stderr.write(`ERR: ${error.message || error} \n`) // Add new line to make it intuitive
 }
 
 const errorHandler = errors => {
--- a/packages/csv-parser-price/src/main.js
+++ b/packages/csv-parser-price/src/main.js
@@ -61,8 +61,25 @@ export default class CsvParserPrice {
           strict: true,
         })
       )
+      .map(a => {
+        if (!a[CONSTANTS.header.sku]) {
+          throw new Error(
+            `${CONSTANTS.header.sku} key is missing/mis-spelled in the csv header.`
+          )
+        }
+        return a
+      })
+      .errors((err, push) => {
+        push(err.message)
+      })
+      .stopOnError(err => {
+        this.logger.error(err)
+        output.emit('error', err)
+      })
       // Sort by SKU so later when reducing prices we can easily group by SKU
-      .sortBy((a, b) => a['variant-sku'].localeCompare(b['variant-sku']))
+      .sortBy((a, b) =>
+        a[CONSTANTS.header.sku].localeCompare(b[CONSTANTS.header.sku])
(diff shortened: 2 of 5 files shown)


packages/csv-parser-price/src/cli.js
```diff
@@ -78,8 +78,9 @@ const logError = error => {
   const errorFormatter = new PrettyError()
 
   if (npmlog.level === 'verbose')
-    process.stderr.write(`ERR: ${errorFormatter.render(error)}`)
-  else process.stderr.write(`ERR: ${error.message || error}`)
+    process.stderr.write(`ERR: ${errorFormatter.render(error)} \n`)
+  // Add new line to make it intuitive
+  else process.stderr.write(`ERR: ${error.message || error} \n`) // Add new line to make it intuitive
```
daern91: 👍

packages/csv-parser-price/src/main.js
```diff
@@ -61,8 +61,25 @@ export default class CsvParserPrice {
           strict: true,
         })
       )
+      .map(a => {
+        if (!a[CONSTANTS.header.sku]) {
```
daern91: Cool, I didn't know that SKU is mandatory when working with these tools.

katmatt: @daern91 How can we be sure about it? Was this fix requested by the only customer using this importer? I think that this change is at least a breaking change and have doubts that it works for every existing customer. What do you think?

packages/csv-parser-price/test/main.spec.js
```diff
@@ -240,6 +240,29 @@ describe('CsvParserPrice::parse', () => {
...
+    return new Promise(done => {
+      const csvParserPrice = new CsvParserPrice({ apiConfig, logger })
+      const inputStream = fs.createReadStream(
+        path.join(__dirname, 'helpers/invalid-sku-sample.csv')
+      )
+
+      const spy = sinon.spy(csvParserPrice.logger, 'error')
+
+      const outputStream = streamtest.v2.toText(() => {
+        console.log(spy.args)
```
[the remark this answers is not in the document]
VineetKumarKushwaha: yes its by mistake. Will remove.

packages/csv-parser-price/src/main.js
VineetKumarKushwaha: @danie1sullivan  @katmatt What I understood from the code is we push price corresponding to the sku provided. And even there are code below dependent on sku-id. So my assumption was if there is not sku provided then we won't be able to update anything.
This my understanding, I may be wrong please correct me if I am.

daern91: Hi again @katmatt and @VineetKumarKushwaha,

Yes below these lines I also found that we use [highland.reduce](https://highlandjs.org/#reduce) which takes an initial source value and an iterator. 

https://github.com/commercetools/nodejs/blob/1292a8c3f248a0815cbb9e8e3e614aef4a408213/packages/csv-parser-price/src/main.js#L83-L111

The iterator sent in is `this.mergeBySku`
https://github.com/commercetools/nodejs/blob/1292a8c3f248a0815cbb9e8e3e614aef4a408213/packages/csv-parser-price/src/main.js#L182-L201
It seems like we would throw above if an SKU was not present. So I guess this tool only works with SKUs and couldnt add prices with variant ID numbers.

The [example in the docs](https://commercetools.github.io/nodejs/cli/csv-parser-price.html#configuration) also use SKU

katmatt: @daern91 Thanks for your explanation and then I'm fine with the changes 👍

Issue #1485, opened 2020-02-08
Remove @ in front of customer group name when export the products.


VineetKumarKushwaha:
### Support reference number

https://jira.commercetools.com/browse/SUPPORT-7090

### Description

We have a customer attempting to import a csv file after exporting the same file and he's hitting an issue where if a price is associated with a customer group, the file adds an @ to the price (i.e. USD 4999@). When he performs the import he gets the error "Can not parse price". Is this expected behavior that the @ be added to the price?

Remove @ in front of customer group name when export the products.

PR #1487, opened 2020-02-11
fix(product-json-to-csv): replace the @ with space before from the customer group

VineetKumarKushwaha:
#### Replace the @ with space before from the customer group while exporting the products.

Fix: #1485


packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -262,16 +262,13 @@ export default class ProductMapping {
   static _mapPriceToString(price: Price): string {
     // Full price:
     // 'country-currencyCode centAmount|discounted.centAmount@customerGroup.name#channel.key$validFrom~validUntil'
+    const customerGroupName = price.customerGroup && price.customerGroup.name
```
daern91: we can use the newer syntax here :) 

`const customerGroupName = price.customerGroup?.name`

VineetKumarKushwaha: nice I will change this.

PR #1488, opened 2020-02-12
chore(typescript): move typescript check at package level from root

VineetKumarKushwaha:
#### Move typescript check at packages level from root

Lerna provide a way to call script from root project.
```
lerna run some-command
```
Lerna will look for  some-command inside all the package and if found will run.

### DIFF
--- a/packages/typescript-sdk/package.json
+++ b/packages/typescript-sdk/package.json
@@ -28,6 +28,7 @@
     "dist"
   ],
   "scripts": {
+    "typecheck:ts": "tsc --noEmit",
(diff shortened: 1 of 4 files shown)


packages/typescript-sdk/package.json
```diff
@@ -28,6 +28,7 @@
     "dist"
   ],
   "scripts": {
+    "typecheck:ts": "yarn tsc --noEmit",
```
daern91: nitpick: I think we could omit `yarn` here.

`"typecheck:ts": "tsc --noEmit"`

Issue #1480, opened 2020-01-29
[product-exporter] Default value for exported productType should be key if available.


svaj:
### Description

The product exporter's default behavior will produce a CSV with the product's product type exported - with the *name* of the product type used as the value for "productType".  There are no unique constraints around product type names, so an export of two products with two different product types that share the same name will have the same value for "productType" in the generated CSV.

This can be reproduced by creating a two product types with the same name but different keys.  Then create some products for each and exporting those via the product exporter's defaults values.

### Expected Behavior

The productType column in the CSV should contain keys or ids of the productType. Keys should be preferred over ids.

### Current Behavior

The CSV is generated with the productType column containing the name of the product types - there is no distinction between the two product types that share a name.

### Context

A customer complained about troubles of using impex, particularly around importing products exported via impex.  After some research, it was found that part of their problems seem to come from having 2 product types that share the same name.

### Possible Solution

Change the default value exported for product type to be key (if available) and then possibly id. As names are not meant to be identifiers, they should not be used to identify the referenced product type.

### Workaround

The customer *could* export via json, or specifying to supply a csv template specifying the productType's id or key be exported.

VineetKumarKushwaha:
Here to solve this we are adding one more column while exporting the products which is productTypeKey where users can see the key.

PR #1496, opened 2020-03-06
feat(product-exporter): add productTypeKey column while exporting

VineetKumarKushwaha:
#### Summary

Adding productTypeKey column while exporting the products.

#### Description

Fixes: #1480


packages/product-json-to-csv/src/map-product-data.js
```diff
@@ -268,7 +268,7 @@ export default class ProductMapping {
       ${price.country ? `${price.country}-` : ''}
       ${price.value.currencyCode} ${price.value.centAmount}${
       price.discounted ? `|${price.discounted.value.centAmount}` : ''
-    }${customerGroupName ? ` ${price.customerGroup.name}` : ''}
+    }${customerGroupName ? ` ${customerGroupName}` : ''}
```
daern91: Nice find 👍

Issue #1497, opened 2020-03-06
Configurable Token Cache


kamaz:
### Description

As part **sdk-middleware-auth** you cache `auth` and reuse when it is present but that only supports a single client. For details: https://github.com/commercetools/nodejs/blob/master/packages/sdk-middleware-auth/src/base-auth-flow.js#L68

As part of our implementation, we have to support multiple clients and we would like to extend **sdk-middleware-auth** to add that support. 

The proposal for the change: 
1. Extend `AuthMiddlewareOptions` to allow passing custom `tokenCache` implementation
2. Change interface for `tokenCache` (https://github.com/commercetools/nodejs/blob/master/types/sdk.js#L185) to: 

```js
type TokenCache {
  get: (client_id?: string) => TokenStore
  set: (cache: TokenStore, client_id?: string) => TokenStore
}
```

For backward compatibility keep the order of parameters and make the`client_id` parameter optional.

3. Change the code occurrences to pass the `client_id` to all gets and sets 

Please let me know if you would be interested with above I would be happy to contribute.

emmenko:
That would be nice yes. We should also do the same for the `TokenProvider` (https://github.com/commercetools/nodejs/blob/master/packages/sdk-auth/src/tokenProvider.js) I guess.

If you feel like it, we are happy to welcome a PR from you 🤗

kamaz:
I've created #1498 

I had a look at `TokenProvider` and that looks like a wrapper around `SdkAuth`. IMHO making it `client` specific may not be a good idea. 

Please have a look at the PR and let me know what you think? 

Also in the context of the change it would be good to hear how would you see that working with `TokenProvider`.

daern91:
Fixed in #1498, thanks again @kamaz! 💯🏅

studentIvan:
Configurable Token Cache is good, but the big part of the data providers, even files are async and it does not support the async token management. For example cloudflare workers have the KV Storage and I would like to have something like this

```ts
function cfTokenStore() {
  return {
    get: () => {
      return JSON.parse(await BFF_KV.get('tokendata'));
    },
    set: (val: { token: string; expirationTime: number }) => {
      /** make async set on set, with the value work as default */
      await BFF_KV.put('tokendata', JSON.stringify(val));
      return val;
    }
  };
}
```

PR #1498, opened 2020-03-06
feat: shared token cached between different client

kamaz:
#### Summary

Following a quick discussion #1497 , I thought I will draft a quick PR to get your view on the implementation and what would like to see as part of it. 

#### Description

We have to create credentials for different clients in the context of a request. Ideally, we don't want to create a custom middleware but would prefer to use yours as it has been for a while and it is well tested and widely used.  

At the moment our code looks following:
```ts
export const createCommercetoolsClient = (ctParams: CommercetoolsClientParams): ByProjectKeyRequestBuilder => {
  const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
    host: ctParams.authUrl,
    projectKey: ctParams.projectKey,
    credentials: {
      clientId: ctParams.clientId,
      clientSecret: ctParams.secret
    },
    fetch
  })
  const httpMiddleware = createHttpMiddleware({
    host: ctParams.host,
    fetch
  })

  const client = createClient({
    middlewares: [authMiddleware, httpMiddleware]
  })

  const apiRoot = createApiBuilderFromCtpClient(client)

  return apiRoot.withProjectKey({
    projectKey: params.projectKey
  })
}
```
but that is not optimal because each time we would need to create an authentication token. To avoid that I would like to propose to have a shared token cache to allow sharing token between clients. 

```ts
// very naive implementation
const tokenCache: TokenCache = {
  cache: {},
  get(client) {
    return this.cache[client]
  },
  set(value, client) {
    this.cache[client] = value
  },
}

export const createCommercetoolsClient = (ctParams: CommercetoolsClientParams): ByProjectKeyRequestBuilder => {
  const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
    host: ctParams.authUrl,
    projectKey: ctParams.projectKey,
    credentials: {
      clientId: ctParams.clientId,
      clientSecret: ctParams.secret
    },
    fetch, 
    tokenCache
  })
  const httpMiddleware = createHttpMiddleware({
    host: ctParams.host,
    fetch
  })

  const client = createClient({
    middlewares: [authMiddleware, httpMiddleware]
  })

  const apiRoot = createApiBuilderFromCtpClient(client)

  return apiRoot.withProjectKey({
    projectKey: ctParams.projectKey
  })
}
```

Please let me know your opinion and what would you like to extend/add as part of this PR.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation


types/sdk.js
```diff
@@ -169,6 +171,10 @@ export type TokenStore = {
   expirationTime: number,
   refreshToken?: string,
 }
+export type TokenCache = {
+  get: (clientId?: string) => TokenStore,
+  set: (cache: TokenStore, clientId?: string) => TokenStore,
```
emmenko: Nit: setters should return `void`

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -65,7 +66,7 @@ function executeRequest({
               const expirationTime = calculateExpirationTime(expiresIn)
 
               // Cache new token
-              tokenCache.set({ token, expirationTime, refreshToken })
+              tokenCache.set({ token, expirationTime, refreshToken }, clientId)
```
emmenko: I would like that we pass different options that might be useful for scoping the cached tokens and not just the `clientId`. The user can then freely decide how to construct the cache key.

For instance, possible options that we can pass are `clientId`, `projectKey`, `host`.

```js
tokenCache.set(
  { token, expirationTime, refreshToken },
  tokenCacheOptions
)
```

The `tokenCacheOptions` contains those values mentioned above.

kamaz: yes that sounds like more generic solution (I've tailored that to our implementation) 

I will also pass them to `get` otherwise the won't be a way to retrieve the token 😄

kamaz:
Hi @emmenko, updated the PR which changes. Please let me know if this is what you think there are other changes required.

packages/sdk-middleware-auth/test/base-auth-flow.spec.js
```diff
@@ -408,4 +408,184 @@ describe('Base Auth Flow', () => {
       )
       createBaseMiddleware({ pendingTasks, tokenCache, requestState }, call2)
     }))
+
+  describe('client id token cache', () => {
+    let tokenCache
+    const createCacheKey = cacheKey =>
+      `${cacheKey.clientId}-${cacheKey.projectKey}-${cacheKey.host}`
```
emmenko:
```suggestion
    const createCacheKey = options =>
      `${options.clientId}-${options.projectKey}-${options.host}`
```

packages/sdk-middleware-auth/test/base-auth-flow.spec.js
```diff
@@ -408,4 +408,184 @@ describe('Base Auth Flow', () => {
...
+        },
+        set(value, client) {
+          this.cache[client] = value
+        },
+      }
+    })
+
+    test('it stores token in token cache in context of token cache key', () =>
+      new Promise(resolve => {
+        const tokenCacheKey = {
```
emmenko:
```suggestion
        const tokenCacheKeyOptions = {
```

kamaz:
@emmenko thanks and I applied the suggestion

emmenko:
@daern91 can you take over and merge?

daern91:
Hi @kamaz , thanks a lot for the PR and this is now released as `@commercetools/sdk-middleware-auth@v6.1.0`

Issue #1482, opened 2020-02-05
Remove "*.sphere.io" defaults in favor of no defaults or the new Host name convention


nkuehn:
### Description

To improve the developer experience and easen our support and training burden all existing references to  `*.sphere.io` and `.commercetools.co`  host names should be either removed in favor of not defaulting to a specific region (a common complaint of US customers is that EU is defaulted everywhere) or, if needed for backwards compatibility, be replaced with the new `*.europe-west1.gcp.commercetools.com` notation. 

The dominant place seem to be the generated typescript SDK files, so this could be a relatively easy fix. 

### Expected Behavior

full text search over the repository for ".sphere.io"  and ".commercetools.co" should yield zero results

### Context

https://docs.commercetools.com/release-notes#releases-2020-02-03-aws-and-new-hostnames

daern91:
fixed in: #1507

PR #1507, opened 2020-03-23
improvement(packages): change .sphere.io .commercetools.co to .europe-west1.gcp.commercetools.com

MarkusAzer:
#### Summary

Following Issue [1482](https://github.com/commercetools/nodejs/issues/1482) , I replaced sphere.io, .commercetools.co To .europe-west1.gcp.commercetools.com

### DIFF
--- a/docs/cli/category-exporter.md
+++ b/docs/cli/category-exporter.md
@@ -29,7 +29,7 @@ Options:
   --apiUrl                   The host URL of the HTTP API service.
                                               [default: "https://api.europe-west1.gcp.commercetools.com"]
   --authUrl                  The host URL of the OAuth API service.
-                                             [default: "https://https://docs.commercetools.com/http-api-authorization#http-api---authorization"]
+                                             [default: "https://auth.europe-west1.gcp.commercetools.com"]
   --accessToken              CTP client access token
                              Required scopes: ['view_products', 'manage_products'][string]
   --projectKey, -p           API project key.                         [required]
--- a/docs/cli/category-exporter.md
+++ b/docs/cli/category-exporter.md
@@ -27,9 +27,9 @@ Options:
   --version                  Show version number                       [boolean]
   --output, -o               Path to output file.            [default: "stdout"]
   --apiUrl                   The host URL of the HTTP API service.
-                                              [default: "https://api.sphere.io"]
+                                              [default: "https://api.europe-west1.gcp.commercetools.com"]
   --authUrl                  The host URL of the OAuth API service.
-                                             [default: "https://auth.sphere.io"]
+                                             [default: "https://https://docs.commercetools.com/http-api-authorization#http-api---authorization"]
   --accessToken              CTP client access token
                              Required scopes: ['view_products', 'manage_products'][string]
   --projectKey, -p           API project key.                         [required]
@@ -59,7 +59,7 @@ import fs from 'fs'
 
 const options = {
     apiConfig: {
-      apiUrl: 'https://api.sphere.io'
+      apiUrl: 'https://api.europe-west1.gcp.commercetools.com'
       host: 'https://auth.sphere.com'
       project_key: 'PROJECT_KEY',
       credentials: {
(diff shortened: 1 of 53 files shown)


docs/cli/category-exporter.md
```diff
@@ -59,7 +59,7 @@ import fs from 'fs'
 
 const options = {
     apiConfig: {
-      apiUrl: 'https://api.sphere.io'
+      apiUrl: 'https://api.europe-west1.gcp.commercetools.com'
       host: 'https://auth.sphere.com'
```
emmenko: What about this one?

docs/cli/resource-deleter.md
```diff
@@ -50,9 +50,9 @@ Options:
   --version                  Show version number.                       [boolean]
   --output, -o               Path to output file.            [default: "stdout"]
   --apiUrl                   The host URL of the HTTP API service.
-                                              [default: "https://api.sphere.io"]
+                                              [default: "https://api.europe-west1.gcp.commercetools.com"]
   --authUrl                  The host URL of the OAuth API service.
-                                             [default: "https://auth.sphere.io"]
+                                             [default: "https://https://docs.commercetools.com/http-api-authorization#http-api---authorization"]
```
emmenko: Those urls are wrong

docs/cli/category-exporter.md
MarkusAzer: prefect and thanks :), i really didnt notice it

emmenko:
What about my point?

> From my point of view, we should also remove the default URL values and just require them.

/cc @daern91 @katmatt

katmatt:
@emmenko You mean that we require that you have to pass in the urls? That would be a breaking change right? Which is fine with me, but should then be reflected in our version update.

emmenko:
Yes of course. It would be a breaking change. There should be no problem with that from my side.

katmatt:
@emmenko I agree with you that with our multi cloud setup it doesn't make a lot of sense to provide defaults. I personally like the idea to have an enum that lists all of our environments and you already suggested it for [labd/terraform-provider-commercetools](https://github.com/labd/terraform-provider-commercetools) . The enums would then provide the auth and api url and that would improve our developer experience. 
That's why I would suggest to keep the default urls and instead discuss how to implement the enum solution. Why should we waste time by fixing a feature that we want to improve?

emmenko:
You mean that we only require `region` and `cloudRegion` and the SDK will then build the proper URLs internally?

Yeah we could also do that but we should be consistent will all other SDKs.
For now I would suggest to just drop the default, and start a discussion with the other SDK teams to look into the more generic approach.

WDYT?

katmatt:
@emmenko all of our other SDKs kept the default urls :wink:  And in our generated Javs SDKs we stopped providing default urls.

/CC @jenschude

jenschude:
I would go a two step approach. First replace the default values and release a minor version and afterwards remove the defaults and require the URLs by releasing a new major version. So it would still be possible to upgrade to the new URLs without touching the code base.

About the enums I had the discussion also with Nikolaus once. The issue is that we have to parameters in the Platform API (importer too?) and for example only one for the ML services and the two parameters can't be used in all combinations.

This would have led to introduce an enum with the Full-URI and a meaning full name. Btw in RAML it's not possible to define parameters in the AuthURIs 🤦‍♂

daern91:
I was thinking of a two-step approach too and think it makes the most sense.

katmatt:
This PR is already quite big and dropping support for default urls seems to require some more discussions:

- Do we drop support for default urls in all of our cli applications too?
- Do we want to implement an enum based better solution or do we still want to have separate auth and api urls?

That's why I would suggest to merge this PR, which already fixed what the associated issue #1482 describes. And we then should discuss the proper solution separately.

emmenko:
Yep, we can of course tackle the other points in other PRs.

packages/sdk-middleware-auth/test/anonymous-session-flow.spec.js
```diff
@@ -16,7 +16,7 @@ function createTestRequest(options) {
 
 function createTestMiddlewareOptions(options) {
   return {
-    host: 'https://auth.commercetools.co',
+    host: 'https://auth.europe-west1.gcp.commercetools.com',
```
daern91: just as a heads up, technically these pointed towards US env but shouldn't matter for tests.

PR #1508, opened 2020-03-26
fix(packages-tests): change .commercetools.co to .europe-west1.gcp.commercetools.com

MarkusAzer:
#### Summary

fix packages tests

#### Description

Some links werent changed yet, so this PR is update for [1507](https://github.com/commercetools/nodejs/pull/1507)


jhr:
.co should be US, not EU, right?

PR #1511, opened 2020-03-30
feat(sync-actions): adding set key action for cart-discounts

NickDevG:
#### Summary
Add `setKey` to CartDiscounts list of actions

#### Description
Adding coverage to the cart discounts sync actions by detecting the `setKey` update action.


NickDevG:
> I just have one question, does this also work for removing the key?
Nice catch. I've only tested for null -> value not the other way around

NickDevG:
> I just have one question, does this also work for removing the key?

I just tested in case it wasn't but it works as expected

Issue #1279, opened 2019-04-26
[sdk-client, sdk-middleware-auth] client.execute() does not reject with invalid auth credentials


MBrendel:
### Support reference number

--

### Description

When providing incorrect auth credentials to the _sdk-middleware-auth_, the promise returned by `client.execute(…)` only rejects on the first call.

```js
import { createRequestBuilder } from '@commercetools/api-request-builder';
import { createClient } from '@commercetools/sdk-client';
import { createAuthMiddlewareForClientCredentialsFlow } from '@commercetools/sdk-middleware-auth';
import { createHttpMiddleware } from '@commercetools/sdk-middleware-http';
import { createQueueMiddleware } from '@commercetools/sdk-middleware-queue';
import fetch from 'node-fetch';

const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
    host: process.env.CT_AUTH_HOST,
    projectKey: process.env.CT_PROJECT_KEY,
    scopes: [process.env.CT_CLIENT_SCOPES],
    credentials: {
        clientId: process.env.CT_CLIENT_ID,
        clientSecret: process.env.CT_CLIENT_SECRET,
    },
    fetch,
});

const httpMiddleware = createHttpMiddleware({
    host: process.env.CT_HOST,
    fetch,
});

const queueMiddleware = createQueueMiddleware({
    concurrency: 5,
});

const client = createClient({
    middlewares: [authMiddleware, httpMiddleware, queueMiddleware],
});

const requestBuilder = createRequestBuilder({
    projectKey: process.env.CT_PROJECT_KEY,
});

const request = {
    uri: requestBuilder.categories.build(),
    method: 'GET',
    headers: {
        Accept: 'application/json',
        'Content-Type': 'application/json',
    },
};

const runRequest = () =>
    client
        .execute(request)
        .then(() => console.info('success'))
        .catch(err => {
            console.error('error');
            throw err;
        });

runRequest()
    .then(runRequest)
    .catch(runRequest);

```

### Expected Behavior

When providing **incorrect** auth credentials (_clientSecret_) the code above should log 'error' **twice**.

### Current Behavior

With **correct** credentials the code above logs 'success' **two** times.
With **incorrect** credentials it logs 'error' only **once**.

Only the first call of `client.execute(…)` returns a promise that gets rejected. 
Promises returned by subsequent calls never get rejected.

### Context

We ran into this issue after deploying code to a new stage where the ENV variables were not set up properly.

PR #1512, opened 2020-03-30
fix(packages): set requestStates in case of wrong auth

MarkusAzer:
#### Summary
fix set requestStates in case of wrong auth

#### Description

Package : sdk-middleware-auth
fix set requestStates in case of wrong auth

This PR is update for [1279](https://github.com/commercetools/nodejs/issues/1279)


MarkusAzer:
Thanks @daern91  :)

You right and i will add test.

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -99,10 +99,17 @@ function executeRequest({
...
         if (parsed) error.body = parsed
+
+        // Dispatch all pending requests
+        requestState.set(false)
+
         response.reject(error)
       })
     })
     .catch((error: Error) => {
+      // Dispatch all pending requests
```
daern91: same question here

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -99,10 +99,17 @@ function executeRequest({
         }
         const error: Object = new Error(parsed ? parsed.message : text)
         if (parsed) error.body = parsed
+
+        // Dispatch all pending requests
```
daern91: I'm not understanding this comment, it looks like we are just rejecting below and not actually dispatching any requests?

MarkusAzer: There is queue of requests,
For ex, once 1st request is failed, then requestState set to false, and dispatch all other queued requests
so 2nd request starts and so on

daern91: I can see that we do that in the condition above: https://github.com/commercetools/nodejs/blob/192f8d02a657bc500ffb0b8845f80bcb1aeadad2/packages/sdk-middleware-auth/src/base-auth-flow.js#L82-L88

but are you sure we also keep dispatching here? I cant see it in the code so maybe I'm just missing a part of the logic somewhere.

MarkusAzer: You right, it should be then sth like reset request status!

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -71,7 +71,7 @@ function executeRequest({
                 tokenCacheKey
               )
 
-              // Dispatch all pending requests
+              // Reset requestState to false
```
daern91, line 74: we shouldn't change this comment

packages/sdk-middleware-auth/src/base-auth-flow.js
```diff
@@ -99,10 +99,17 @@ function executeRequest({
         }
         const error: Object = new Error(parsed ? parsed.message : text)
         if (parsed) error.body = parsed
+
+        // Reset requestState to false
```
daern91, line 103: Could we please make this comment a bit more descriptive? I.e., why are we resetting the request state? 

That we are setting it to false is already described well by the code already so currently it's a superfluous comment IMO.

packages/sdk-middleware-auth/test/base-auth-flow.spec.js
```diff
@@ -371,6 +371,53 @@ describe('Base Auth Flow', () => {
...
+    const requestState = store(false)
+    const pendingTasks = []
+
+    const startCreateBaseMiddleware = () =>
+      new Promise((resolve, reject) => {
+        expect(requestState.get()).toBe(false)
+
+        // fire off promise returning function and change requestState to true
+        // we do not await the response before the assertion since we set
+        // reqeustState to false in the catch block
```
daern91, line 400: nitpick: I checked the code now and technically we don't set it in the catch block but here: https://github.com/commercetools/nodejs/blob/acf7cd4a4b208736dee67a6b946a8e7ccbd07eb8/packages/sdk-middleware-auth/src/base-auth-flow.js#L92-L107

packages/sdk-middleware-auth/test/base-auth-flow.spec.js
```diff
@@ -371,6 +371,53 @@ describe('Base Auth Flow', () => {
...
+        })
+
+        expect(requestState.get()).toBe(true)
+        expect(requestCount).toBe(1) // Make sure that nock runned the mocked request
+
+        return start
+      })
+
+    try {
+      // await promise to fail and set requestState to false
```
daern91, line 414:
```suggestion
      // await promise failure which should set `requestState` to `false`
```

Issue #1492, opened 2020-02-14
[sync-actions] changeAttributeOrderByName doesn't work when adding attribute def


katmatt:
### Description

If I add a new attribute definition to a product type, then build the update actions for the product type and send the updates to the backend, then I get the following error:
```json
[{"code":"InvalidOperation","message":"The given attributes must be equal to the type attributes.","action":{"type":"changeAttributeOrderByName","attributeNames":["color","size","weight"]},"actionIndex":1}]
```
Our documentation for the [changeAttributeOrderByName](https://docs.commercetools.com/http-api-projects-productTypes#change-the-order-of-attributedefinitions) update action talk about _current_ attribute defintions and my understanding is that this doesn't include attribute definitions added during the same update request.

### Expected Behavior

The `changeAttributeOrderByName` update action should only be added if the attribute defintions were only reordered, but no defintion was added or removed.

### Current Behavior

The `changeAttributeOrderByName` update action is be added if the order of attribute defintions was changed, even when an attribute definition was added or removed.

### Context

One of our users reported this problem when he tried to update an existing product type.

### Possible Solution

We should only add the `changeAttributeOrderByName` update action when the order of attribute definitions was changed. And not if defintions were added or removed. This of course then requires that an user that wants to update the order has to perform to update requests:

1. the first one to add or remove attribute definitions
2. the second one to correctly order the attribute defintions

danrleyt:
Issue not reproducible anymore.

PR #1513, opened 2020-04-02
fix(sync-actions): removing changeattributeorder remove or add is per…

danrleyt:
…formed

#### Summary

Removing the changeAttributeOrderByName action when remove or add action is previously performed. 

#### Description

This is related to the issue #1492 .

#### Todo

- Tests
- [ ] Unit
- [ ] Integration
- [ ] Acceptance
- [x] Documentation

### DIFF
--- a/packages/sync-actions/src/product-types-actions.js
+++ b/packages/sync-actions/src/product-types-actions.js
@@ -263,7 +263,6 @@ const generateChangeAttributeOrderAction = (
   updateActions = []
 ) => {
   if (!attrsOld.length || !attrsNew.length) return null
-
   const newAttributesOrder = attrsNew.map((attribute) => attribute.name)
 
   const removedAttributeNames = updateActions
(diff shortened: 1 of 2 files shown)


katmatt:
@danrleyt I created this issue while working on our importer. But I'm not 100% sure if my assumption is correct that you can't change the attribute order when you added or removed an attribute. It would be great if you could check my assumption.

/CC @adnasa @tdeekens @emmenko I guess we added support for the `changeAttributeOrderAction` update action because we needed it for reordering of attributes in our merchant center. Would this fix break this feature?

adnasa:
> I guess we added support for the changeAttributeOrderAction update action because we needed it for reordering of attributes in our merchant center. Would this fix break this feature?

This feature does not exist in MC as far as I am concerned. Can you check the team about this?

katmatt:
@adnasa I'll check with the team 👍

danrleyt:
Issue is not reproducible.

PR #1515, opened 2020-04-03
feat(request-builder): add orderNumber support on order service

jherey:
#### Summary

Add order number support on order service

#### Description

- added function `byOrderNumber` 
- added orderNumber key to uri

### DIFF
--- a/packages/api-request-builder/test/query-id.spec.js
+++ b/packages/api-request-builder/test/query-id.spec.js
@@ -53,7 +53,7 @@ describe('queryId', () => {
 
   test('should throw if orderNumber is invalid', () => {
     expect(() => service.byOrderNumber('hi')).toThrow(
-      /Required argument for `byCustomerId` is missing/
+      /Required argument for `byOrderNumber` is missing/
     )
   })
 
--- a/packages/api-request-builder/test/query-id.spec.js
+++ b/packages/api-request-builder/test/query-id.spec.js
@@ -34,8 +34,25 @@ describe('queryId', () => {
     expect(service.params.customerId).toBe('myCustomer')
   })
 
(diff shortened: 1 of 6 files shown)


packages/api-request-builder/test/query-id.spec.js
```diff
@@ -34,12 +34,29 @@ describe('queryId', () => {
     expect(service.params.customerId).toBe('myCustomer')
   })
 
+  test('should set the orderNumber param', () => {
```
daern91, line 37: nitpick: It would be nice to move this block down below the test for `customerId` so they are all organized together.

packages/api-request-builder/test/query-id.spec.js
```diff
@@ -34,12 +34,29 @@ describe('queryId', () => {
...
+
   test('should throw if customerId is missing', () => {
     expect(() => service.byCustomerId()).toThrow(
       /Required argument for `byCustomerId` is missing/
     )
   })
 
+  test('should throw if orderNumber is missing', () => {
+    expect(() => service.byOrderNumber()).toThrow(
+      /Required argument for `byOrderNumber` is missing or invalid/
```
daern91, line 50:
```suggestion
      /Required argument for `byOrderNumber` is missing/
```

another ocd nitpick from me 😆

packages/api-request-builder/test/query-id.spec.js
jherey: I've moved it, take another look.

packages/api-request-builder/test/query-id.spec.js
jherey: lol, this is fixed now

PR #1524, opened 2020-04-06
refactor(api-request-builder): rename createService argument

jherey:
#### Summary

Renamed `options` to `projectKey`

#### Description

Changed the 2nd argument to `createService` function to `projectKey`. This makes it easy to read the code and understand, as against `options` which means an object.


daern91 (approved):
Another nice little find, thanks @jherey!

danrleyt (approved):
Good catch!

katmatt (approved):
You have eyes sharp like an 🦅  Looks very good to me 👍

PR #1525, opened 2020-04-09
fix(sync-actions): copy empty array props for products and variants

markus-azer:
#### Summary
Fix sync actions (products and variants)

#### Description

Make sure to copy empty array props for products and variants between old object and new object while updating to avoid this error `TypeError: Cannot read property '0' of undefined`

### DIFF
--- a/packages/sync-actions/src/products.js
+++ b/packages/sync-actions/src/products.js
@@ -125,8 +125,7 @@ function moveMasterVariantsIntoVariants(
   before: Object,
   now: Object
 ): Array<Object> {
-  /* eslint-disable no-param-reassign */
-  ;[before, now] = copyEmptyArrayProps(before, now)
+  const [beforeCopy, nowCopy] = copyEmptyArrayProps(before, now)
   const move = (obj: Object): Object => ({
     ...obj,
     masterVariant: undefined,
@@ -135,8 +134,8 @@ function moveMasterVariantsIntoVariants(
   const hasMasterVariant = (obj: Object): Object => obj && obj.masterVariant
 
   return [
-    hasMasterVariant(before) ? move(before) : before,
-    hasMasterVariant(now) ? move(now) : now,
+    hasMasterVariant(beforeCopy) ? move(beforeCopy) : beforeCopy,
+    hasMasterVariant(nowCopy) ? move(nowCopy) : nowCopy,
   ]
 }
 
--- a/packages/sync-actions/src/products.js
+++ b/packages/sync-actions/src/products.js
@@ -122,10 +122,11 @@ function createProductMapActions(
 }
 
 function moveMasterVariantsIntoVariants(
-  objectBefore: Object,
-  objectNow: Object
+  before: Object,
+  now: Object
 ): Array<Object> {
-  const [before, now] = copyEmptyArrayProps(objectBefore, objectNow)
+  /* eslint-disable no-param-reassign */
+  ;[before, now] = copyEmptyArrayProps(before, now)
   const move = (obj: Object): Object => ({
     ...obj,
     masterVariant: undefined,
--- a/packages/sync-actions/src/products.js
+++ b/packages/sync-actions/src/products.js
@@ -11,6 +11,7 @@ import createMapActionGroup from './utils/create-map-action-group'
 import * as productActions from './product-actions'
 import * as diffpatcher from './utils/diffpatcher'
 import findMatchingPairs from './utils/find-matching-pairs'
+import copyEmptyArrayProps from './utils/copy-empty-array-props'
 
 const actionGroups = [
   'base',
@@ -121,9 +122,10 @@ function createProductMapActions(
 }
 
 function moveMasterVariantsIntoVariants(
-  before: Object,
-  now: Object
+  objectBefore: Object,
+  objectNow: Object
 ): Array<Object> {
+  const [before, now] = copyEmptyArrayProps(objectBefore, objectNow)
   const move = (obj: Object): Object => ({
     ...obj,
     masterVariant: undefined,
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -9,18 +9,15 @@ export default function copyEmptyArrayProps(oldObj, newObj) {
   const newObjWithFixedEmptyArray = Object.entries(oldObj).reduce(
     (acc, [key, value]) => {
       const isArray = Array.isArray(value)
+      const newObjectValueIsUndefined = newObj[key] === undefined
 
-      // check if the value is array on `oldObj` and the key not found in `acc` (newObjWithFixedEmptyArray)
-      if (isArray && acc[key] === undefined) {
-        // init `key` with empty array
+      if (isArray && newObjectValueIsUndefined) {
         return { ...acc, [key]: [] }
       }
 
-      // note that `typeof value === 'object'` returns true also in case of array
-      // check if its not array and its object
-      if (!isArray && typeof value === 'object' && acc[key]) {
+      if (!isArray && typeof value === 'object' && newObj[key]) {
         // recursion, so we could check for nested objects
-        const returnedNewObjCopy = copyEmptyArrayProps(value, acc[key])[1]
+        const returnedNewObjCopy = copyEmptyArrayProps(value, newObj[key])[1]
         return { ...acc, [key]: returnedNewObjCopy }
       }
 
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -17,6 +17,7 @@ export default function copyEmptyArrayProps(oldObj, newObj) {
         newObj[key] = [] // eslint-disable-line no-param-reassign
       }
     } else if (typeof value === 'object' && oldObj[key] && newObj[key]) {
+      // recursion, so we could check for nested objects
       copyEmptyArrayProps(oldObj[key], newObj[key])
     }
   })
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -1,35 +1,3 @@
-const deepCopyObj = (obj) => JSON.parse(JSON.stringify(obj))
-
-const copyEmptyArrayPropsRecursion = (oldObjCopy, newObjCopy) => {
-  const newObjCopyOne = Object.entries(oldObjCopy).reduce(
-    (acc, [key, value]) => {
-      // check if the value is array on `oldObjCopy`
-      if (Array.isArray(value)) {
-        const foundKey = Object.keys(acc).findIndex(
-          (accKey) => JSON.stringify(accKey) === JSON.stringify(key)
-        )
-
-        // if the key not found in `acc` (newObjectCopy)
-        if (foundKey === -1) {
-          return { ...acc, [key]: [] }
-        }
-      } else if (typeof value === 'object' && acc[key]) {
-        // recursion, so we could check for nested objects
-        const returnedNewObjCopy = copyEmptyArrayPropsRecursion(
-          value,
-          acc[key]
-        )[1]
-        return { ...acc, [key]: returnedNewObjCopy }
-      }
-
-      return acc
-    },
-    { ...newObjCopy }
-  )
-
-  return [oldObjCopy, newObjCopyOne]
-}
-
 /**
  * @function copyEmptyArrayProps
  * @description Takes two objects and if there are Array props in oldObj which doesn't exist in newObj, then copy it with an empty value.
@@ -39,8 +7,28 @@ const copyEmptyArrayPropsRecursion = (oldObjCopy, newObjCopy) => {
  */
 
 export default function copyEmptyArrayProps(oldObj, newObj) {
-  const oldObjCopy = deepCopyObj(oldObj)
-  const newObjCopy = deepCopyObj(newObj)
+  const newObjCopy = Object.entries(oldObj).reduce(
+    (acc, [key, value]) => {
+      // check if the value is array on `oldObj`
+      if (Array.isArray(value)) {
+        const foundKey = Object.keys(acc).findIndex(
+          (accKey) => JSON.stringify(accKey) === JSON.stringify(key)
+        )
 
-  return copyEmptyArrayPropsRecursion(oldObjCopy, newObjCopy)
+        // if the key not found in `acc` (newObjectCopy), init `key` with empty array
+        if (foundKey === -1) {
+          return { ...acc, [key]: [] }
+        }
+      } else if (typeof value === 'object' && acc[key]) {
+        // recursion, so we could check for nested objects
+        const returnedNewObjCopy = copyEmptyArrayProps(value, acc[key])[1]
+        return { ...acc, [key]: returnedNewObjCopy }
+      }
+
+      return acc
+    },
+    { ...newObj }
+  )
+
+  return [oldObj, newObjCopy]
 }
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -16,6 +16,8 @@ export default function copyEmptyArrayProps(oldObj, newObj) {
       if (foundKey === -1) {
         newObj[key] = [] // eslint-disable-line no-param-reassign
       }
+    } else if (typeof value === 'object' && oldObj[key] && newObj[key]) {
+      copyEmptyArrayProps(oldObj[key], newObj[key])
     }
   })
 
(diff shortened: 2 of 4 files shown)


packages/sync-actions/src/products.js
```diff
@@ -121,9 +122,10 @@ function createProductMapActions(
 }
 
 function moveMasterVariantsIntoVariants(
-  before: Object,
-  now: Object
+  objectBefore: Object,
+  objectNow: Object
```
katmatt, line 126: I think we should keep the existing names because we use them everywhere else. That's why I suggest to revert these changes.

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -982,6 +982,99 @@ describe('Actions', () => {
     ])
   })
 
+  test('update product-variant', async () => {
```
katmatt, line 985: I don't really understand this test. I thought the problem is that some array type properties aren't defined? Where can I see this in your test?

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -16,6 +16,8 @@ export default function copyEmptyArrayProps(oldObj, newObj) {
       if (foundKey === -1) {
         newObj[key] = [] // eslint-disable-line no-param-reassign
       }
+    } else if (typeof value === 'object' && oldObj[key] && newObj[key]) {
```
katmatt, line 19: I don't understand this condition. If  `oldObj[key] && newObj[key]` is true, what needs to be fixed then? In that case both objects already have a defined value and there is no need to fix them. Or am I wrong here?

katmatt (changes requested):
With your description of the problem - which I like and can follow - and your test, I still don't get how you fixed it. I would suggest to improve your test and then see if your fix works. Sounds like a plan to you too? 😄

packages/sync-actions/src/utils/copy-empty-array-props.js
MarkusAzer: You right, i should add more descriptive comment

Its recursion, so if we have obj like this one 
`{ 
  prop1: []
  obj: {
    prop2: []
  }
}`

with the above condition, then we copy both prop1 and prop2, without this condition then we will just copy prop1 which will cause the issue `TypeError: Cannot read property '0' of undefined`

packages/sync-actions/src/products.js
MarkusAzer: i just changed it so that who reads the code could follow up easily 
plus i kept same names with this line
`[before, now] = copyEmptyArrayProps(objectBefore, objectNow)`

if i reverted it, it will be like below, the same object changed
`function moveMasterVariantsIntoVariants(
  before: Object,
  now: Object
): Array<Object> {
  [before, now] = copyEmptyArrayProps(before, now)`

what do you think?!

katmatt: @MarkusAzer Where can I see this change? How about pushing your changes, then it's easier to review them? 😄

MarkusAzer: @katmatt sorry forgot to push, now its up

katmatt:
@MarkusAzer Cool, looks fine to me. And you now just have to improve your test 😄

packages/sync-actions/test/product-sync-variants.spec.js
MarkusAzer: i should be testing this func `moveMasterVariantsIntoVariants` but its not exported,
so i have to start with this func `SyncAction`.

On before object, `masterVariant` which is nested object from product, which contain these props 
`
       prices: [],
        images: [],
        attributes: [],
        assets: [],
`
while now object doesnt contain these props which causing the issue  
`TypeError: Cannot read property '0' of undefined`

katmatt (commented):
@daern91 What do you think about the recursion? Should we try to reduce the maximum recursion that we allow? And do you know if we already have a similar issue somewhere else?

packages/sync-actions/src/products.js
```diff
@@ -124,6 +125,8 @@ function moveMasterVariantsIntoVariants(
   before: Object,
   now: Object
 ): Array<Object> {
+  /* eslint-disable no-param-reassign */
+  ;[before, now] = copyEmptyArrayProps(before, now)
```
daern91, line 129: I'm not a fan of bending the linting rules here to mutate the passed args. Couldn't we just name them something else? Like 

`const [beforeCopy, nowCopy] = copyEmptyArrayProps(before, now)`

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -982,6 +982,33 @@ describe('Actions', () => {
     ])
   })
 
+  test('shouldnt thow `TypeError: Cannot read property 0 of undefined` if array props exist on before and doesnt exist on now object', async () => {
```
daern91, line 985:
```suggestion
  test('should build <insert-the-change-action-here> when value is an array', async () => {
```

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -982,6 +982,33 @@ describe('Actions', () => {
...
+      masterVariant: {
+        id: 1,
+        key: 'default-masterVariant-key-de07e9bc-e352-422f-abe6-910d9879b995',
+        isMasterVariant: true,
+      },
+    }
+
+    productsSync.buildActions(now, before)
+
+    expect(before).toMatchObject(now)
```
daern91, line 1009: Please have a look at the other tests in this file, we should do the assertion on the returning value from the `buildActions` method. Not on the variables we assigned within this test.

daern91 (changes requested):
Great question @katmatt. Generally, I think we should avoid any recursion in javascript since it doesn't support tail-call optimization. But maybe, in this case, it's OK since I don't expect the stack to grow too large? Do you know any cases where it could grow out of hand? I'm not 100% sure about the domain, to be honest.

katmatt:
@MarkusAzer @daern91 Just another question that came to my mind: Why do we have to fix this code at all? Why did it work with the previous versions? Would be interesting to investigate this a bit further, just to avoid that we stumble over this issue again with a different resource type...

katmatt:
@daern91 You could pass in an object with a reference to itself or a child. Then the code will throw a stack overflow error. What does node do if the stack overflows?

packages/sync-actions/src/products.js
MarkusAzer: @katmatt

packages/sync-actions/test/product-sync-variants.spec.js
MarkusAzer: Yep i did it like this beginning https://github.com/commercetools/nodejs/pull/1525/commits/625b1525e16fe9ba749351f5b66bea8d542128db

But the main point of the test is to compare the two objects to make sure that array props which exist on before will be coppied to now object to avoid the error `TypeError: Cannot read property 0 of undefined`

MarkusAzer:
> @MarkusAzer @daern91 Just another question that came to my mind: Why do we have to fix this code at all? Why did it work with the previous versions? Would be interesting to investigate this a bit further, just to avoid that we stumble over this issue again with a different resource type...

this pull request should fix issue https://github.com/commercetools/commercetools-importer/issues/1163 ,  import api tests were failing cause of this bug

packages/sync-actions/src/products.js
daern91: I saw the discussion above and agree that we should keep the passed argument names. But I think it would make sense to change them here instead.

katmatt: @MarkusAzer We use PRs to discuss code and to learn as a team together. We decided to use eslint to check our coding guidelines. And this means we should avoid bending the lint rules if we can. And it seems to be easy to fix it. That's why I suggest to fix it as @daern91 suggested it. What do you think about it?

packages/sync-actions/test/product-sync-variants.spec.js
daern91: I see. So we are mutating the original `now`?

daern91:
> @daern91 You could pass in an object with a reference to itself or a child. Then the code will throw a stack overflow error. What does node do if the stack overflows?

I see. Then we can't use recursion here I'm afraid :/

packages/sync-actions/test/product-sync-variants.spec.js
daern91: https://github.com/commercetools/nodejs/commit/625b1525e16fe9ba749351f5b66bea8d542128db looks like a good test to me 👍

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -7,17 +32,8 @@
...
-      if (foundKey === -1) {
-        newObj[key] = [] // eslint-disable-line no-param-reassign
-      }
-    }
-  })
+  const oldObjCopy = deepCopyObj(oldObj)
+  const newObjCopy = deepCopyObj(newObj)
 
-  return [oldObj, newObj]
+  return copyEmptyArrayPropsRecursion(oldObjCopy, newObjCopy)
```
katmatt, line 38: You now solve the issue by copying the input parameters, which obviously works. But if I have to fix another bug in this module in the future, I will still have to understand that the `copyEmptyArrayPropsRecursion` has the side effect that it modifies its input parameter - which I consider as a bad practice. How difficult would it to be to fix this function to not mutate the parameters?

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,3 +1,28 @@
+const deepCopyObj = (obj) => JSON.parse(JSON.stringify(obj))
```
katmatt, line 1: BTW: This will fail with an error if the `obj` has a cyclic reference too 😉

MarkusAzer: True, cyclic reference Not my best param 😆

packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -982,6 +982,46 @@ describe('Actions', () => {
...
+  test('should build `setAttribute` action', async () => {
+    const before = {
+      categories: [],
+      masterVariant: {
+        id: 1,
+        key: 'default-masterVariant-key-de07e9bc-e352-422f-abe6-910d9879b995',
+        isMasterVariant: true,
+        prices: [],
+        images: [],
+        attributes: [],
```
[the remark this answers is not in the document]
MarkusAzer: `before` object is mostly pulled from CT platform with init values so for array values, it init with empty array, so its supposed to exist on `before` obj.
while `now` obj is missing these props which causing the error

if you check the now obj, you will see its missing these props

packages/sync-actions/src/utils/copy-empty-array-props.js
katmatt: @MarkusAzer See PR #1526 😄 😉

MarkusAzer: @katmatt i really love the test :), but the coping nested object didnt added!
could you take alook on the new changes?!

packages/sync-actions/src/utils/copy-empty-array-props.js
MarkusAzer: i totally removed this function

packages/sync-actions/src/utils/copy-empty-array-props.js
katmatt: @MarkusAzer I would prefer if you would review my PR together with @daern91 i think it makes more sense to use `Object.fromEntries` in this case, because it's the inverse function of `Object.entries` and easier to understand then `reduce`. 
If we merged that PR to master you can merge master to this branch. And then you can add the recursion and update the test to have the recursion covered too. What do you think, @MarkusAzer ? Sounds like a plan?

MarkusAzer: Yep 😎 plan

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -8,12 +8,25 @@
...
+
+      // check if the value is array on `oldObj` and the key not found in `acc` (newObjWithFixedEmptyArray)
+      if (isArray && acc[key] === undefined) {
+        // init `key` with empty array
+        return { ...acc, [key]: [] }
+      }
+
+      // note that `typeof value === 'object'` returns true also in case of array
+      // check if its not array and its object
+      if (!isArray && typeof value === 'object' && acc[key]) {
```
daern91, line 21: Just a question that came to my mind, do we ever have an array of objects here?

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -8,12 +8,25 @@
 export default function copyEmptyArrayProps(oldObj, newObj) {
   const newObjWithFixedEmptyArray = Object.entries(oldObj).reduce(
     (acc, [key, value]) => {
-      const replaceWithEmptyArray =
-        Array.isArray(value) && newObj[key] === undefined
-      acc[key] = replaceWithEmptyArray ? [] : newObj[key]
+      const isArray = Array.isArray(value)
+
+      // check if the value is array on `oldObj` and the key not found in `acc` (newObjWithFixedEmptyArray)
+      if (isArray && acc[key] === undefined) {
```
daern91, line 14: I found this comment a bit confusing to read and follow

suggestion: we could rename `isArray` variable to `oldObjectValueIsArray`. Then make a second variable 
`const newObjectValueIsUndefined = acc[key] === undefined`

then we won't need the comment and just have the condition
``if (oldObjectValueIsArray && newObjectValueIsUndefined) {...}

wdyt?

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -8,12 +8,25 @@
...
+      if (!isArray && typeof value === 'object' && acc[key]) {
+        // recursion, so we could check for nested objects
+        const returnedNewObjCopy = copyEmptyArrayProps(value, acc[key])[1]
+        return { ...acc, [key]: returnedNewObjCopy }
+      }
+
       return acc
     },
-    {}
+    { ...newObj }
```
daern91, line 29: very nice 👍

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -8,12 +8,25 @@
...
     (acc, [key, value]) => {
-      const replaceWithEmptyArray =
-        Array.isArray(value) && newObj[key] === undefined
-      acc[key] = replaceWithEmptyArray ? [] : newObj[key]
+      const isArray = Array.isArray(value)
+
+      // check if the value is array on `oldObj` and the key not found in `acc` (newObjWithFixedEmptyArray)
+      if (isArray && acc[key] === undefined) {
+        // init `key` with empty array
+        return { ...acc, [key]: [] }
```
daern91, line 16: I like this a lot 👍

daern91 (commented):
This is starting to look really good! Thanks a lot for your work and patience here, @MarkusAzer. I just left a few suggestions/questions. And of course, feel free to argue back at any time 👍

katmatt:
@daern91 Looks like @MarkusAzer and I already agreed on how to proceed 😄 And he wanted to contact you. For now I would suggest to wait for @MarkusAzer

packages/sync-actions/src/utils/copy-empty-array-props.js
MarkusAzer: Yep thats prefect and more readable  👍

packages/sync-actions/src/utils/copy-empty-array-props.js
MarkusAzer: nope, but i shall create new pull request with this feature, cause i wanted this PR to be simple.

packages/sync-actions/src/utils/copy-empty-array-props.js
katmatt: I also relearned this dynamic property trick from @MarkusAzer 😎

packages/sync-actions/test/utils/copy-empty-array-props.spec.js
```diff
@@ -7,9 +7,114 @@ test('should add empty array for undefined prop', () => {
...
+  const newObj = {
+    anotherProp: 2,
+  }
+
+  const [old, fixedNewObj] = copyEmptyArrayProps(oldObj, newObj)
+
+  expect(old).toEqual(oldObj)
+  expect(fixedNewObj).toEqual({ ...newObj, emptyArray: [] })
+  expect(newObj).toEqual({ anotherProp: 2 })
+})
```
katmatt, line 120: Wow, I like these tests a lot 👍 Great job 😎

katmatt (approved):
Great job @MarkusAzer 🥇 💯 Thanks a lot for adding the nice unit tests and having so much patience 👍

daern91 (approved):
Looks great to me now, well done @MarkusAzer and thanks for the dedication 🥇 🚀

MarkusAzer:
> Looks great to me now, well done @MarkusAzer and thanks for the dedication 🥇 🚀

Thanks to @katmatt and @daern91 for your help and getting this code to look pretty nice and easy to read 💯 😃

jherey (approved):

packages/sync-actions/src/utils/copy-empty-array-props.js
adnasa: this looks to me like a breaking change

PR #1527, opened 2020-04-15
fix: use reduce instead of fromPairs

daern91:
I saw that the tests failed since node 10 didn't support `fromEntries`. PR stops relying on this function and instead builds the object with help of reduce.

### DIFF
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -6,7 +6,7 @@
  * @returns {Array} Ordered Array [oldObj, newObj]
  */
 export default function copyEmptyArrayProps(oldObj, newObj) {
-  const entriesWithEmptyArrays = Object.entries(oldObj).reduce(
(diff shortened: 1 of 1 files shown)


packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -6,11 +6,15 @@
  * @returns {Array} Ordered Array [oldObj, newObj]
  */
 export default function copyEmptyArrayProps(oldObj, newObj) {
-  const entriesWithEmptyArrays = Object.entries(oldObj).map(([key, value]) => {
-    const replaceWithEmptyArray =
-      Array.isArray(value) && newObj[key] === undefined
-    return [key, replaceWithEmptyArray ? [] : newObj[key]]
-  })
+  const entriesWithEmptyArrays = Object.entries(oldObj).reduce(
```
katmatt, line 9: But we should rename this, may be to `valueWithFixedEmptyArray`?

PR #1530, opened 2020-04-22
feat(sync-actions/stores): add `setDistributionChannels`

adnasa:
#### Summary

- adds support `setDistributionChannels` 

#### Description

Context added this update-action recently to the Stores API.
It may look like it is not documented. It is documented, however we are waiting for the MC feature to be released before we release the documentation.

please view.

### DIFF
--- a/packages/sync-actions/test/stores-sync.spec.js
+++ b/packages/sync-actions/test/stores-sync.spec.js
@@ -10,6 +10,7 @@ describe('Exports', () => {
     expect(baseActionsList).toEqual([
       { action: 'setName', key: 'name' },
       { action: 'setLanguages', key: 'languages' },
+      { action: 'setDistributionChannels', key: 'distributionChannels' },
     ])
   })
 })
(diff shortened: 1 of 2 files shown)


packages/sync-actions/test/stores-sync.spec.js
```diff
@@ -45,4 +46,35 @@ describe('Actions', () => {
     const expected = [{ action: 'setLanguages', languages: now.languages }]
     expect(actual).toEqual(expected)
   })
+
+  test('should build `setDistributionsChannels` action', () => {
```
adnasa, line 50: it is important to note that `distributionChannels` on a Store is a list of referenced distribution channels.
It is not intended that the update actions are generated to be `changeDistributionChannel`.. instead, we set the list of references

adnasa:
cc @svranesevic

adnasa:
let's set up a release 😉 
thanks for the fast response 💟

Issue #1439, opened 2019-11-20
Error thrown when synching categories


katmatt:
### Description

When I call the `sync-actions` function `categorySync.buildActions` with this test data:
```js
      const before = {
        assets: [],
      }
      const now = {
      }
      const actual = categorySync.buildActions(now, before)
```
the following error is thrown:
```
Cannot read property '0' of undefined
```

### Expected Behavior

No error should be thrown and the function should return an empty array.

### Context

We tried to import the categories.csv from sunrise-data (https://github.com/commercetools/commercetools-sunrise-data/blob/master/data/categories.csv) and then found this issue when trying to updat an existing category with an empty `assets` array.

PR #1532, opened 2020-04-23
Adnasa null check copy array props

adnasa:
#### Summary

- caused by https://github.com/commercetools/nodejs/pull/1525

### DIFF
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -7,30 +7,32 @@ import isNil from 'lodash.isnil'
  * @returns {Array} Ordered Array [oldObj, newObj]
  */
 export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
-  const nextObjectWithEmptyArray = Object.entries(oldObj).reduce(
-    (nextObject, [key, value]) => {
-      const isArray = Array.isArray(value)
-      const newObjectValueIsUndefined = newObj[key] === undefined
+  if (!isNil(oldObj) && !isNil(newObj)) {
+    const nextObjectWithEmptyArray = Object.entries(oldObj).reduce(
+      (nextObject, [key, value]) => {
+        const isArray = Array.isArray(value)
+        const newObjectValueIsUndefined = newObj[key] === undefined
 
-      if (isArray && newObjectValueIsUndefined) {
-        return { ...nextObject, [key]: [] }
-      }
+        if (isArray && newObjectValueIsUndefined) {
+          return { ...nextObject, [key]: [] }
+        }
 
-      if (
-        !isNil(value) &&
-        !isArray &&
-        typeof value === 'object' &&
-        !isNil(newObj[key])
-      ) {
-        // recursion, so we could check for nested objects
-        const [, returnedNewObjCopy] = copyEmptyArrayProps(value, newObj[key])
-        return { ...nextObject, [key]: returnedNewObjCopy }
-      }
+        if (
+          !isNil(value) &&
+          !isArray &&
+          typeof value === 'object' &&
+          !isNil(newObj[key])
+        ) {
+          // recursion, so we could check for nested objects
+          const [, returnedNewObjCopy] = copyEmptyArrayProps(value, newObj[key])
+          return { ...nextObject, [key]: returnedNewObjCopy }
+        }
 
-      return nextObject
-    },
-    { ...newObj }
-  )
-
-  return [oldObj, nextObjectWithEmptyArray]
+        return nextObject
+      },
+      { ...newObj }
+    )
+    return [oldObj, nextObjectWithEmptyArray]
+  }
+  return [oldObj, newObj]
 }
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -11,9 +11,8 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
     const nextObjectWithEmptyArray = Object.entries(oldObj).reduce(
       (nextObject, [key, value]) => {
         const isArray = Array.isArray(value)
-        const newObjectValueIsUndefined = newObj[key] === undefined
 
-        if (isArray && newObjectValueIsUndefined) {
+        if (isArray && isNil(newObj[key])) {
           return { ...nextObject, [key]: [] }
         }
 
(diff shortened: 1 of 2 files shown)


packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,36 @@
...
  * @function copyEmptyArrayProps
  * @description Takes two objects and if there are Array props in oldObj which doesn't exist in newObj, then copy it with an empty value.
  * @param {Object} oldObj
  * @param {Object} newObj
  * @returns {Array} Ordered Array [oldObj, newObj]
  */
-export default function copyEmptyArrayProps(oldObj, newObj) {
-  const newObjWithFixedEmptyArray = Object.entries(oldObj).reduce(
-    (acc, [key, value]) => {
+export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
```
adnasa, line 9: if we intend to spread `newObj`, we need to fallback to an `Object` if the passed value is undefined

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,36 @@
...
-        const returnedNewObjCopy = copyEmptyArrayProps(value, newObj[key])[1]
-        return { ...acc, [key]: returnedNewObjCopy }
+        const [, returnedNewObjCopy] = copyEmptyArrayProps(value, newObj[key])
+        return { ...nextObject, [key]: returnedNewObjCopy }
       }
 
-      return acc
+      return nextObject
     },
     { ...newObj }
```
adnasa, line 32: then this won't break

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,36 @@
...
   )
 
-  return [oldObj, newObjWithFixedEmptyArray]
+  return [oldObj, nextObjectWithEmptyArray]
```
adnasa, line 35: and since the consumer expects and object, what we essentially do is forward back what was given to us

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,36 @@
...
+    (nextObject, [key, value]) => {
       const isArray = Array.isArray(value)
       const newObjectValueIsUndefined = newObj[key] === undefined
 
       if (isArray && newObjectValueIsUndefined) {
-        return { ...acc, [key]: [] }
+        return { ...nextObject, [key]: [] }
       }
 
-      if (!isArray && typeof value === 'object' && newObj[key]) {
```
adnasa, line 18: careful!

`typeof null === object`, so this will execute the recursion when `*Obj` is null

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,37 @@
...
-        // recursion, so we could check for nested objects
-        const returnedNewObjCopy = copyEmptyArrayProps(value, newObj[key])[1]
-        return { ...acc, [key]: returnedNewObjCopy }
-      }
+        if (
+          !isNil(value) &&
+          !isArray &&
+          typeof value === 'object' &&
+          !isNil(newObj[key])
+        ) {
```
adnasa, line 24: ☝🏽
this if statement doesn't make much sense and someone has to explain it...

- we check if `newObject[key]` is null or undefined.
- we check if `value` is null or undefined.

`value` derives from `newObject[key]`,because we use entries with the `newObj` as entry point

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,37 @@
+import isNil from 'lodash.isnil'
```
daern91, line 1: Good to see some battle-tested solutions here 👍

packages/sync-actions/src/utils/copy-empty-array-props.js
daern91: Technically it actually works to spread `undefined`, but I agree that this makes it easier to follow.

```
const a = {...undefined};
console.log(a);
// {}
let b;
const c = {...b}
console.log(c);
// {}
```

packages/sync-actions/src/utils/copy-empty-array-props.js
daern91: Oufff! Great find!

packages/sync-actions/src/utils/copy-empty-array-props.js
daern91: hmm, I'm reading it now and see that the `value` check makes sense since we only want to go one level deeper if its an object in the `oldObj`.

The check for `newObject[key]` then makes sure that it's actually undefined. Since if it's already defined we shouldn't copy the values from `oldObj` and instead allow the new values to take priority.

Would above make sense for you too, @adnasa?

Then my follow-up question is, are we allowed to set it to `null` in the CTP? If so, we should keep the first check and not replace it with `isNil`

adnasa: I have no context here, so I'm not sure how to proceed on this topic. 
it's separate to me, and the scope of this PR is to have the null/undefined check

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,37 @@
...
  * @description Takes two objects and if there are Array props in oldObj which doesn't exist in newObj, then copy it with an empty value.
  * @param {Object} oldObj
  * @param {Object} newObj
  * @returns {Array} Ordered Array [oldObj, newObj]
  */
-export default function copyEmptyArrayProps(oldObj, newObj) {
-  const newObjWithFixedEmptyArray = Object.entries(oldObj).reduce(
-    (acc, [key, value]) => {
-      const isArray = Array.isArray(value)
-      const newObjectValueIsUndefined = newObj[key] === undefined
```
adnasa, line 12: aah I see now... 
I'll fix this

packages/sync-actions/src/utils/copy-empty-array-props.js
MarkusAzer: it were introduced to category, fixing this issue [1439](https://github.com/commercetools/nodejs/issues/1439)

adnasa:
in the last commit, I simplified the implementation so that it is easier to reason about
https://github.com/commercetools/nodejs/pull/1532/commits/29066db804d941947469570c683bfae1b38de593

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -1,30 +1,38 @@
...
+          }
+        }
+        if (!isNil(newObj[key]) && typeof value === 'object') {
+          const [, nestedObject] = copyEmptyArrayProps(value, newObj[key])
+          return {
+            ...merged,
+            [key]: nestedObject,
+          }
+        }
+        return { ...merged }
```
MarkusAzer, line 31:
```suggestion
        return merged
```

adnasa:
I will release this

amine-benselim:
H8IT :trollface:

adnasa:
HAHAHAHAHA ❤️

PR #1533, opened 2020-04-24
Feat support custom price

MarkusAzer:
#### Summary

Add support for adding custom fields on price

#### Description

Support these two actions: `setProductPriceCustomType` ,`setProductPriceCustomField`
As mentioned in platform docs :
[docs](https://docs.commercetools.com/http-api-projects-products#set-price-custom-type)
[tutorial](https://docs.commercetools.com/tutorial-custom-fields-prices)


packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -217,6 +217,56 @@ describe('Actions', () => {
...
+          //   custom: {
+          //     type: {
+          //       typeId: 'type',
+          //       id: 'customType1',
+          //     },
+          //     fields: {
+          //       source: "shop",
+          //     },
+          //   },
+          // },
```
MarkusAzer, line 269: I didnt add this test cause sth weird happened, when i send this actions to platform 
`
{
  "version": 15,
  "actions": [
{
    "action": "setProductPriceCustomType",
    "priceId": "f28c248b-91e9-4f33-aeda-7237ecdabb60"
  }
  ]
}
`
This deletes the price object, i will give it more time to make sure its the platform

packages/sync-actions/src/utils/action-map-custom.js
```diff
@@ -26,21 +26,32 @@ const extractTypeFields = (diffedFields, nextFields) =>
...
+  newObj,
+  oldObj,
+  customActions = {}
+) {
   const actions = []
+  const priceId = oldObj.id || undefined
+  if (customActions.setCustomType)
+    Actions.setCustomType = customActions.setCustomType
+  if (customActions.setCustomField)
+    Actions.setCustomField = customActions.setCustomField
```
MarkusAzer, line 40: add customAction as param, so i can override the default actions

packages/sync-actions/src/product-actions.js
```diff
@@ -273,6 +274,7 @@ function _buildVariantPricesAction(
         // still has other values, otherwise simply return
         const filteredPrice = { ...price }
         delete filteredPrice.discounted
+        delete filteredPrice.custom
```
katmatt, line 277: Is this necessary so that we still create `changePrice` update actions?

packages/sync-actions/src/utils/action-map-custom.js
katmatt: Good idea 👍

packages/sync-actions/src/utils/action-map-custom.js
```diff
@@ -26,21 +26,32 @@ const extractTypeFields = (diffedFields, nextFields) =>
...
   if (!diff.custom) return actions
   if (hasSingleCustomFieldChanged(diff)) {
     // If custom is not defined on the new or old category
     const custom = diffpatcher.getDeltaValue(diff.custom, oldObj)
-    actions.push({ action: Actions.setCustomType, ...custom })
+    actions.push({ action: Actions.setCustomType, priceId, ...custom })
```
katmatt, line 45: This works because `priceId` is `undefined` if the `newObj`/`oldObj` aren't prices? I know that the assignment above `const priceId = oldObj.id || undefined` already existed before your changes, but to me it looks a bit dangerous. A lot of our types have an `id`... Whot do you think about it @MarkusAzer ?

packages/sync-actions/test/product-sync-base.spec.js
```diff
@@ -13,6 +13,7 @@ describe('Exports', () => {
       'meta',
       'references',
       'prices',
+      'pricesCustom',
```
katmatt, line 16: I'm not sure if we really need a new action group here, we could have reused the existing `prices` action group here instead.

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -217,6 +217,56 @@ describe('Actions', () => {
             validFrom: dateNow,
             validUntil: twoWeeksFromNow,
           },
+          {
+            // set price custom type
```
katmatt, line 221: What does this comment mean? I don't see any `custom` in the lines below?

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -217,6 +217,56 @@ describe('Actions', () => {
...
+            id: '888',
+            value: { currencyCode: 'GBP', centAmount: 1000 },
+            country: 'UK',
+            validFrom: dateNow,
+            validUntil: twoWeeksFromNow,
+          },
+          {
+            // set price custom type and field
```
katmatt, line 229: Same as above: What does this comment mean? I don't see any `custom` in the lines below?

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -217,6 +217,56 @@ describe('Actions', () => {
...
+            id: '999',
+            value: { currencyCode: 'GBP', centAmount: 1000 },
+            country: 'UK',
+            validFrom: dateNow,
+            validUntil: twoWeeksFromNow,
+          },
+          {
+            // change price custom field
```
katmatt, line 237: But I can understand this comment 😄

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -217,6 +217,56 @@ describe('Actions', () => {
...
+                typeId: 'type',
+                id: '5678',
+              },
+              fields: {
+                source: 'shop',
+              },
+            },
+          },
+          {
+            // remove price custom field
```
katmatt, line 254: But this comment confuses me again: How does this test the removal of a custom field when the lines below still have a `custom` property?

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -257,6 +307,62 @@ describe('Actions', () => {
             validFrom: twoWeeksFromNow,
             validUntil: threeWeeksFromNow,
           },
+          {
+            // set price custom type
```
katmatt, line 311: 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -257,6 +307,62 @@ describe('Actions', () => {
...
+            validUntil: twoWeeksFromNow,
+            custom: {
+              type: {
+                typeId: 'type',
+                id: '5678',
+              },
+            },
+          },
+          {
+            // set price custom type and field
```
katmatt, line 325: 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -257,6 +307,62 @@ describe('Actions', () => {
...
+                typeId: 'type',
+                id: '5678',
+              },
+              fields: {
+                source: 'random',
+              },
+            },
+          },
+          {
+            // remove price custom field and type
```
katmatt, line 359: 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -257,6 +307,62 @@ describe('Actions', () => {
...
+                typeId: 'type',
+                id: '5678',
+              },
+              fields: {
+                source: 'shop',
+              },
+            },
+          },
+          {
+            // change price custom field
```
katmatt, line 342: 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -267,7 +373,7 @@ describe('Actions', () => {
     })
 
     test('should build five update actions', () => {
-      expect(actions).toHaveLength(5)
+      expect(actions).toHaveLength(9)
```
katmatt, line 376: 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -340,6 +446,63 @@ describe('Actions', () => {
       )
     })
 
+    test('should build `setProductPriceCustomType` action without fields', () => {
```
katmatt, line 449: Nice test 👍

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -567,7 +730,7 @@ describe('Actions', () => {
               fractionDigits: 2,
             },
             id: '7960b455-ab04-4722-a8b2-431460d60012',
-            custom: {
+            customField: {
```
katmatt, line 733: This looks wrong, please revert this change.

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -595,7 +758,7 @@ describe('Actions', () => {
                 fractionDigits: 2,
               },
               id: '38a9667a-976d-4bcc-8d2a-b04701e41f18',
-              custom: {
+              customField: {
```
katmatt, line 761: This looks wrong, please revert this change.

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -640,7 +803,7 @@ describe('Actions', () => {
               centAmount: 6495,
               currencyCode: 'EUR',
             },
-            custom: {
+            customField: {
```
katmatt, line 806: This looks wrong, please revert this change.

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -664,7 +827,7 @@ describe('Actions', () => {
                 centAmount: 8995,
                 currencyCode: 'EUR',
               },
-              custom: {
+              customField: {
```
katmatt, line 830: This looks wrong, please revert this change.

packages/sync-actions/src/product-actions.js
MarkusAzer: i think i wont be right to create `changePrice` action for `custom` prop, since we now create specific action for `custom` props.
i even tried it with platform api to send this action 
`{
  "action": "changePrice",
  "priceId": "111",
  "price": {
    "custom": {
      "type": {
        "id": "2222"
      },
      "fields": {
        "promotionDiscountType": "Tiefpreis",
        "promotionRegularPrice": 8995,
        "promotionValidToExclusive": "SKU-111",
        "promotionValidFromInclusive": "2018-10-02T00:00:35.570Z"
      }
    },
  }
}`

the respond were like they didnt even add anything
and i think as they mentioned on doc, this field `custom` for price can only change with these two actions `setProductPriceCustomType` `setProductPriceCustomField`

thats why i exclude it here, and pass it to `actionsMapPricesCustom`

packages/sync-actions/src/utils/action-map-custom.js
MarkusAzer: Thats totally right, should i pass it as optional params like this maybe
`export default function actionsMapCustom(
  diff,
  newObj,
  oldObj,
  custom = {
    customActions = {},
    priceId
  }
)`

what do you think?!

packages/sync-actions/test/product-sync-base.spec.js
MarkusAzer: yea true 👍 , i just saw it like this on category `export const actionGroups = ['base', 'references', 'meta', 'custom', 'assets']`, so i thought to separate it like category.
I also think it should be same as within price actions

packages/sync-actions/test/product-sync-prices.spec.js
MarkusAzer: regarding these comments, we have two objects `before` and `now`, this actions created cause of the diff between these two, so i comment the same comment on both `before` and `now`

check line `311`, its same price id `888`, i just mention the reason why this comment exist, got me?!

katmatt: thanks for your explanation and makes sense now 👍

packages/sync-actions/src/product-actions.js
MarkusAzer: Please ignore what i commented above, i need to make sure of sth

packages/sync-actions/src/utils/action-map-custom.js
katmatt: I think it's a good idea, but would like to hear opinions from the others too. 
CC: @daern91 @jherey @danrleyt

packages/sync-actions/src/product-actions.js
MarkusAzer: you totally right, thanks :), i even added test to make sure that custom prop wont be deleted while action `changePrice`

packages/sync-actions/test/product-sync-base.spec.js
MarkusAzer: as we discussed, and as you told me :), i would be better to separate it with action, so we give flexibility choosing the actions

packages/sync-actions/test/product-sync-prices.spec.js
MarkusAzer: done

packages/sync-actions/test/product-sync-prices.spec.js
MarkusAzer: done

packages/sync-actions/test/product-sync-prices.spec.js
MarkusAzer: done

packages/sync-actions/test/product-sync-prices.spec.js
MarkusAzer: done

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -340,6 +480,95 @@ describe('Actions', () => {
       )
     })
 
+    test('should build `changePrice` action without deleting `custom` prop', () => {
```
MarkusAzer, line 483: this test is to validate that `custom` wont be changed or removed if `changePrice` action produced.

packages/sync-actions/src/utils/action-map-custom.js
danrleyt: Yeah, I can't think of anything else, I like your idea @MarkusAzer. This way we can guarantee that has a value.

jherey: I think your approach works, let's go with this. We can make updates later if we find a better way.

packages/sync-actions/src/product-actions.js
```diff
@@ -635,6 +636,57 @@ export function actionsMapPrices(diff, oldObj, newObj, variantHashMap) {
...
+            index,
+            oldVariant.prices,
+            newVariant.prices
+          )
+
+          if (
+            price.custom &&
+            (REGEX_UNDERSCORE_NUMBER.test(index) || REGEX_NUMBER.test(index))
+          ) {
+            const a = actionsMapCustom(price, newPrice, oldPrice, {
```
jherey, line 676: Can we change the variable name `a` here? I believe something descriptive will be better.

packages/sync-actions/test/product-sync-base.spec.js
jherey: What is an `action group` if I may ask? Sorry for derailing the conversation here.

MarkusAzer: Don't be :), as i understood, `actionGroups` is a way to custom or default the running actions `doMapActions`, which then choose which actions should get generated ex: `updatePrice` which belong to `prices` action group

packages/sync-actions/src/product-actions.js
MarkusAzer: Done 👍

packages/sync-actions/test/product-sync-base.spec.js
MarkusAzer: Honestly until now, not sure which is better, but i think giving the option is better!

packages/sync-actions/src/products.js
```diff
@@ -18,6 +18,7 @@ const actionGroups = [
   'meta',
   'references',
   'prices',
+  'pricesCustom',
```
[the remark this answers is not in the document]
MarkusAzer: when Matthias and i had conversation, at first glance we thought merging it with `price` would be better, but then he suggests that leaving it is better for more optional.
Correct me if i'm wrong @katmatt ?!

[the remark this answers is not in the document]
katmatt: To provide a bit more context:
- our platform has a limit on the maximum number of update actions that you can send with one request (see https://docs.commercetools.com/http-api-contract#update-actions-per-request)
- we now added two more update actions
- we should at least provide an option to our users to disable these new extra update actions
- that's why we should keep a separate action group for the custom fields on prices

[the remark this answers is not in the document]
daern91: Thanks a lot :) This extra information makes it super clear 👍

Issue #1534, opened 2020-04-24
Handle BOM nicely for importing prices through csv-parser-price


VineetKumarKushwaha:
For importing products and stock / inventory, BOMs are ignored. For prices, they lead to an error message. This is leading to problems and confusions, and should be changed.

Support ticket: https://jira.commercetools.com/browse/SUPPORT-835

Package Name:- **csv-parser-price**

PR #1535, opened 2020-04-24
fix(csv-parser-price): handle UTF-8 with BOM encoding while price import

VineetKumarKushwaha:
#### UTF-8 with BOM encoding while doing price import

Fixes: #1534

### DIFF
--- a/packages/csv-parser-price/src/main.js
+++ b/packages/csv-parser-price/src/main.js
@@ -7,6 +7,7 @@ import memoize from 'lodash.memoize'
 import isNil from 'lodash.isnil'
 import { unflatten } from 'flat'
 import fetch from 'node-fetch'
+import stripBom from 'strip-bom-stream'
 
 import { createAuthMiddlewareForClientCredentialsFlow } from '@commercetools/sdk-middleware-auth'
 import { createClient } from '@commercetools/sdk-client'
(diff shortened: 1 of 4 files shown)


packages/csv-parser-price/src/main.js
```diff
@@ -55,13 +56,14 @@ export default class CsvParserPrice {
 
     highland(input)
       // Parse CSV return each row as object
+      .through(stripBom())
```
daern91, line 59: awesome! 👍

packages/csv-parser-price/src/main.js
```diff
@@ -55,13 +56,14 @@ export default class CsvParserPrice {
...
       .through(
         csv({
           separator: this.delimiter,
           strict: true,
         })
       )
-      .map((a) => {
+      .map(a => {
```
daern91, line 66: hmm, these formatting changes shouldn't happen. Could you please run `yarn format` in the root package folder to undo these changes? and maybe `yarn install` if you're running some old dependencies locally?

VineetKumarKushwaha: It happens automatically. Sure will update.

VineetKumarKushwaha: done

PR #1539, opened 2020-04-27
fix(ts-sdk): Update the sdk following sdk changes

### LINKED ISSUE #1536: [typescript-sdk] Add support for /{projectKey}/shipping-methods/matching-cart
### Expected Behavior

I would expect that the new endpoint for the shippingmethods that matches cart is supported:

`/{projectKey}/shipping-methods/matching-cart?cartId=`

https://docs.commercetools.com/http-api-projects-shippingMethods#get-shippingmethods-for-a-cart

### Current Behavior

The current `typescript-sdk` does only have supported for the deprecated endpoint:
`'/shipping-methods?cartId=`

https://github.com/commercetools/nodejs/blob/08b63751b131fbefee9f46c5e89a32f02ea8e749/packages/typescript-sdk/src/gen/client/shipping-methods/ByProjectKeyShippingMethodsRequestBuilder.ts#L45

which gaves a deprecated warning:
`'x-deprecation-notice': [
[start:watch]       "The usage of '/shipping-methods?cartId=' endpoint is deprecated. Please use '/shipping-methods/matching-cart?cartId=' instead."
[start:watch]     ]`

### Context

We cannot use the typescript sdk for using the new endpoint, only for the deprecated one

### Possible Solution

Regenerate the types for shipping methods according to the current situation to have it supported

acbeni:
#### Summary

Sdk Update

#### Description

Update the sdk to follow latest spec changes

fixes #1536 

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


packages/typescript-sdk/generate-sdk.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# to install the rmf-codegen please refer to https://github.com/commercetools/rmf-codegen#install-rmf-codegen-cli
+rmf-codegen generate $API_RAML_FILE -o src/generated -t typescript_client -s shared -m models -c client
```
daern91, line 4: Thanks for adding this one 👍

PR #1542, opened 2020-04-28
fix(ts-sdk): add return type for shipping method

acbeni:
#### Summary

Add return type for shipping method sub endpoints

#### Description

This PR tries to fix the lack of  return type for sub shipping method endpoints reported by our user in https://github.com/commercetools/nodejs/issues/1536


packages/sync-actions/test/cart-discounts-sync.spec.js
```diff
@@ -100,8 +100,8 @@ describe('Cart Discounts Exports', () => {
         expect.arrayContaining([
           {
             action: 'setKey',
-            key: 'key'
-          }
+            key: 'key',
```
daern91, line 103: hmm these formatting changes shouldnt be here, could you please run `yarn format` inside the root folder to fix them?

acbeni: i run `yarn format` inside the root folder, but i got them still, i dont know why i got them

PR #1543, opened 2020-04-28
feat(SyncActions): Add CustomType actions in inventory entries

NickDevG:
#### CustomType actions in inventory entries
Currently changes over `CustomType` and `CustomField` are not being detected by the sync actions

#### Description

Adding the group of custom actions and the corresponding tests

### DIFF
--- a/packages/sync-actions/src/customers.js
+++ b/packages/sync-actions/src/customers.js
@@ -80,7 +80,7 @@ export default (
   // this resulting function mapActionGroup will call the callback function
   // for whitelisted action groups and return the return value of the callback
   // It will return an empty array for blacklisted action groups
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/customers.js
```diff
@@ -80,7 +80,7 @@ export default (
   // this resulting function mapActionGroup will call the callback function
   // for whitelisted action groups and return the return value of the callback
   // It will return an empty array for blacklisted action groups
-  const mapActionGroup = createMapActionGroup(syncActionConfig)
+  const mapActionGroup = createMapActionGroup(actionGroupList)
```
daern91, line 83: oh man, good find 👍

PR #1558, opened 2020-05-25
fix(ts-sdk): organise imports in ts sdk

acbeni:
#### Summary

Organise imports for TS sdk.
#### Description

Add organize imports scripts to allow, stricter configs in the future for the compiler.

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


packages/typescript-sdk/package.json
```diff
@@ -28,6 +28,9 @@
     "dist"
   ],
   "scripts": {
+    "organize_imports": "find src/generated -type f -name '*.ts' | xargs organize-imports-cli",
```
katmatt, line 31: Interesting that you have to do so much shell magic to organize imports 😄 I would expect that this is a usual task and would be better supported by `organize-imports-cli` 😞

acbeni: Actually whenever i searched for organizing imports it was UI instructions, and the typescript linter removed it, so i was stuck with this 😅 . not perfect but it does the job 👍

acbeni:
Thank you @katmatt  the review 🎉

Issue #1561, opened 2020-06-02
sync-actions for prices is not working properly if the custom field is removed/unset


ahmetoz:
### Description
With the latest release of (4.7.1) sync-actions, price sync actions are causing the error described below, 
Here an unset/removing a custom field with a different price creates additional setProductPriceCustomField after changePrice.

```
{
      code: 'InvalidOperation',
      message: "Cannot remove custom field 'touchpoints' because it does not exist.",
      action: {
        action: 'setProductPriceCustomField',
        priceId: '9fe6610f-b9dd-4830-8c4b-6276fb61e909',
        staged: true,
        name: 'touchpoints'
      }
}
```
## Sample code
```js
const { createSyncProducts } = require('@commercetools/sync-actions')

const now = {
  "masterVariant": {
    "prices": [
      {
        "value": {
          "type": "centPrecision",
          "currencyCode": "EUR",
          "centAmount": 5678,
          "fractionDigits": 2
        },
        "id": "9fe6610f-b9dd-4830-8c4b-6276fb61e909",
        "country": "DE",
        "custom": {
          "type": {
            "typeId": "type",
            "id": "218d8068-9fbf-4083-bdd6-9e18f2e33817"
          },
          "fields": {
            "published": false
          }
        }
      }
    ]
  }
}
const before = {
  "masterVariant": {
    "prices": [
      {
        "value": {
          "type": "centPrecision",
          "currencyCode": "EUR",
          "centAmount": 1900,
          "fractionDigits": 2
        },
        "id": "9fe6610f-b9dd-4830-8c4b-6276fb61e909",
        "country": "DE",
        "custom": {
          "type": {
            "typeId": "type",
            "id": "218d8068-9fbf-4083-bdd6-9e18f2e33817"
          },
          "fields": {
            "touchpoints": [
              "value"
            ]
          }
        }
      }
    ]
  }
}

const syncProducts = createSyncProducts()
const actions = syncProducts.buildActions(now, before)

console.log(JSON.stringify(actions, null, 4))


```

### Expected Behavior

Not sure but maybe it should not create `setProductPriceCustomField` together with changePrice, because `changePrice` already updating the custom fields, or order should be different 🤔

### Output from the sample code

```json
[
    {
        "action": "changePrice",
        "priceId": "9fe6610f-b9dd-4830-8c4b-6276fb61e909",
        "price": {
            "value": {
                "type": "centPrecision",
                "currencyCode": "EUR",
                "centAmount": 5678,
[...]

PR #1562, opened 2020-06-02
fix(sync-actions): generate PriceCustom actions before changePrice

MarkusAzer:
#### Summary

Fix for issue [1561](https://github.com/commercetools/nodejs/issues/1561)

#### Description

Generating `PriceCustom` actions before `changePrice` to fix actions overloading each other.
The ordering of the actions tested via Impex

### DIFF
--- a/packages/sync-actions/test/product-sync-prices.spec.js
+++ b/packages/sync-actions/test/product-sync-prices.spec.js
@@ -169,6 +169,101 @@ describe('Actions', () => {
     expect(actions).toEqual([])
   })
 
+  test('should generate PriceCustom actions before changePrice action', () => {
+    const before = {
+      id: '123',
+      masterVariant: {
+        prices: [
+          {
+            value: {
+              type: 'centPrecision',
+              currencyCode: 'EUR',
+              centAmount: 1900,
+              fractionDigits: 2,
+            },
+            id: '9fe6610f',
+            country: 'DE',
+            custom: {
+              type: {
+                typeId: 'type',
+                id: '218d8068',
+              },
+              fields: {
+                touchpoints: ['value'],
+              },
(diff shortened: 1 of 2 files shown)


MarkusAzer:
> I guess here ordering update was the simplest solution and looks good to me, I have not tried it but do we know it will pass an integration test, did you have a chance to run some tests against platform?
> 
> Also even it works, another concern for me here is if we change prices with the same custom fields, do we actually need to run another action here before ? Especially performance-wise 🤔

yep regarding tests with platform, I simulated it with Impex and so it works as expected.

regarding actions, we wanted to cover all action which platform accepts.

there is solution regarding ignoring some actions while generating it, which is basically just pickup actions which you need with specify actionGroups.

MarkusAzer:
> thanks for the fix @MarkusAzer 🎉

Big Thanks @ahmetoz :) for reporting the issue 🥇

ahmetoz:
@MarkusAzer what is the plan for release?

MarkusAzer:
> @MarkusAzer what is the plan for release?

I'm just waiting for the team to review it and then, it should be realised, i believe mostly today

packages/sync-actions/test/product-sync-prices.spec.js
```diff
@@ -169,6 +169,101 @@ describe('Actions', () => {
     expect(actions).toEqual([])
   })
 
+  test('should generate PriceCustom actions before changePrice action', () => {
```
katmatt, line 172: Perfect and thanks for adding this test 👍 This way we can be sure that further changes won't break it 😎

daern91, line 172: Super!

PR #1571, opened 2020-06-17
feat(sync-actions/stores): add set supply channels

adnasa:
#### Summary

- adds `setSupplyChannels` for stores

#### Description

- same deal as https://github.com/commercetools/nodejs/pull/1530#issue-407171862. It may look like it is missing documentation, but Context team will wait until the MC feature is released before we release the doc update.


adnasa:
will remove my custom label on merge 😉

adnasa:
thanks for the reviews. Unless there is any objections, I will release this today cc @daern91

Issue #1528, opened 2020-04-16
How to use node sdk-auth to get token after error


ashishhk:
### Support reference number
[Support-8396](https://jira.commercetools.com/browse/SUPPORT-8396)

### Description
I am trying to use node sdk-auth for sunrise.

Initially I try to use an invalid refresh token when that fails I try to get a new token.

I'd expect getting the new token would make an xhr request but what happens is that the getTokenInfo fails with the same error as the refresh token request without making a request. After that you can "run again" as much as you like but the sdk will never make another request and keeps failing with "The refresh token was not found. It may have expired." even though anonymous token flow was set.

```js
import SdkAuth, {
  TokenProvider
} from "@commercetools/sdk-auth";

const tokenProvider = new TokenProvider({
  sdkAuth: new SdkAuth({
    host: "https://auth.europe-west1.gcp.commercetools.com",
    projectKey: "sunrise-spa",
    credentials: {
      clientId: "XXXXX",
      clientSecret: "XXXXX"
    },
    scopes: [
      "manage_my_profile:sunrise-spa create_anonymous_token:sunrise-spa"
    ]
  }),
  onTokenInfoChanged: tokenInfo => {
    console.log("token info changed", tokenInfo);
  }
});

export default () => {
  tokenProvider.fetchTokenInfo = sdkAuth =>
    sdkAuth.refreshTokenFlow(
      encodeURIComponent("invalid refresh token")
    );
  tokenProvider.invalidateTokenInfo();
  return tokenProvider.getTokenInfo().catch(error => {
    console.log("expected this error:", error);
    // refresh token did not work, get a brand new token
    tokenProvider.fetchTokenInfo = sdkAuth =>
      sdkAuth.anonymousFlow();
    tokenProvider.invalidateTokenInfo();
    // this makes no xhr and just fails with previous error
    return tokenProvider.getTokenInfo();
  });
};
```
more detailed code [here](https://github.com/harm-meijer/harm.meijer-commercetools.de) 

### Possible Solution
#1279 could be a good reference point, currently fixed

VineetKumarKushwaha:
This issue is fixed and released in v3.0.8

harm-meijer:
@valoriecarli Is it true that with this version you no longer need to url encode the refresh token? I has an [issue](https://jira.commercetools.com/browse/SUPPORT-8367) with that before and with this version it **doesn't** work if I url encode the refresh token when passing it to `sdkAuth.refreshTokenFlow(`

[...]

PR #1579, opened 2020-06-30
fix(sdk-auth): Invalidate cache even after promise reject

VineetKumarKushwaha:
#### Invalidate cache even after promise reject

Fixes: #1528


packages/sdk-auth/src/tokenProvider.js
```diff
@@ -79,6 +79,8 @@ export default class TokenProvider {
     return this.fetchTokenInfoPromise.then((tokenInfo) => {
       this.fetchTokenInfoPromise = null
       return tokenInfo
+    }).catch(() => {
+        this.fetchTokenInfoPromise = null
```
daern91, line 83:
```suggestion
    }).catch((e) => {
        this.fetchTokenInfoPromise = null
        throw e
```
Just swallowing the error looks dangerous to me, how about above?

packages/sdk-auth/src/tokenProvider.js
```diff
@@ -91,6 +93,8 @@ export default class TokenProvider {
     return this.refreshTokenFlowPromise.then((refreshTokenInfo) => {
       this.refreshTokenFlowPromise = null
       return refreshTokenInfo
+    }).catch(() => {
+        this.refreshTokenFlowPromise = null
```
daern91, line 97:
```suggestion
    }).catch((e) => {
        this.fetchTokenInfoPromise = null
        throw e
```
Just swallowing the error looks dangerous to me, how about above?

daern91:
I also added the whole SDK team for review. Me and Matthias are also of course here to help :)

packages/sdk-auth/src/tokenProvider.js
jenschude: C&P error as your suggestion resets the wrong promise

packages/sdk-auth/src/tokenProvider.js
VineetKumarKushwaha: Didn't realize and even the resetting the promise so changed the behaviour

packages/sdk-auth/src/tokenProvider.js
VineetKumarKushwaha: Didn't realize and even that resetting the promise so updated the code.

daern91: Interesting, what does C&P error mean, @jenschude?

jenschude: Copy&Paste. I meant your suggestion proposes instead of refreshTokenPromise a fetchTokenPromise

daern91: Ooops. Good one 👍

packages/sdk-auth/src/tokenProvider.js
```diff
@@ -79,6 +79,9 @@ export default class TokenProvider {
     return this.fetchTokenInfoPromise.then((tokenInfo) => {
       this.fetchTokenInfoPromise = null
       return tokenInfo
+    }).catch((error) => {
+        this.fetchTokenInfoPromise = null
+        throw new Error(error)
```
daern91, line 84: Why do we wrap the error inside a new error? Wouldnt it make more sense to just rethrow it?

`throw error`

packages/sdk-auth/test/token-provider.spec.js
```diff
@@ -41,6 +41,34 @@ describe('Token Provider', () => {
...
+        const _tokenProvider = new TokenProvider({
+            sdkAuth,
+            fetchTokenInfo: jest
+                .fn()
+                .mockImplementation(() => Promise.reject(new Error("Invalid Token"))),
+        })
+        expect(_tokenProvider.fetchTokenInfoPromise).toBeFalsy();
+        try {
+            await _tokenProvider._performFetchTokenInfo()
+        } catch(e) { /** */}
```
daern91, line 54: and here I would recommend wrapping the below code inside the catch block instead of commenting it out. IMO it would make it more clear.

packages/sdk-auth/test/token-provider.spec.js
```diff
@@ -41,6 +41,34 @@ describe('Token Provider', () => {
...
+
+        expect(_tokenProvider.fetchTokenInfoPromise).toBeFalsy();
+    })
```
daern91, line 57:
```suggestion
        } catch(e) { 
          expect(_tokenProvider.fetchTokenInfoPromise).toBeFalsy();
        }
    })
```

packages/sdk-auth/src/tokenProvider.js
VineetKumarKushwaha: it is not error  but error message so we can not throw only string.

packages/sdk-auth/test/token-provider.spec.js
VineetKumarKushwaha: Done

packages/sdk-auth/test/token-provider.spec.js
VineetKumarKushwaha: Done

packages/sdk-auth/src/tokenProvider.js
VineetKumarKushwaha: Done

daern91: huh?

```
const a = () => {throw new Error('failing')}
const b = () => {try { a();} catch (error) {console.log(error instanceof Error)}; }

b()
// true
```

VineetKumarKushwaha:
@daern91 Resolved all the discussions.

packages/sdk-auth/src/tokenProvider.js
VineetKumarKushwaha: @daern91 didn't realize that check above comment I already did ;-)

PR #1580, opened 2020-07-02
Add array check for copy empty array props

markus-azer:
#### Description

Add functionality to `copyEmptyArrayProps` to check nested object inside the array.
Which will solve `Cannot read property '0' of undefined`


packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
           ...newObj,
           ...nextObject,
         }
+
+        if (key === 'custom') {
```
MarkusAzer, line 19: Ignore `custom` as this object is dynamic.

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
...
           return {
             ...merged,
             [key]: isNil(newObj[key]) ? [] : newObj[key],
           }
         }
-        if (!isNil(newObj[key]) && typeof value === 'object') {
+        if (
+          !isNil(newObj[key]) &&
+          typeof value === 'object' &&
+          !(value instanceof Date)
```
MarkusAzer, line 59: Check if its not date, as `typeof date === 'object' ` return `true`

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
...
+        if (Array.isArray(value) && newObj[key] && newObj[key].length >= 1) {
+          /* eslint-disable no-plusplus */
+          for (let i = 0; i < newObj[key].length; i++) {
+            if (
+              !isNil(newObj[key][i]) &&
+              typeof newObj[key][i] === 'object' &&
+              !isNil(newObj[key][i].id)
+            ) {
+              /* eslint-disable eqeqeq */
+              const foundObject = value.find((v) => v.id == newObj[key][i].id)
```
MarkusAzer, line 34: Get same `object` inside the `oldObj` array

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
...
           ...nextObject,
         }
+
+        if (key === 'custom') {
+          return {
+            ...merged,
+          }
+        }
+
+        if (Array.isArray(value) && newObj[key] && newObj[key].length >= 1) {
```
MarkusAzer, line 25: Just wondering, should i add comments, to explain whats going on, or is the code explains itself?!

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
           ...newObj,
           ...nextObject,
         }
+
+        if (key === 'custom') {
+          return {
+            ...merged,
```
adnasa, line 21: just `return merged` ;)

packages/sync-actions/src/utils/copy-empty-array-props.js
adnasa: I would move this to a constant, maybe?

packages/sync-actions/src/utils/copy-empty-array-props.js
adnasa: your question already answered the question 😉 
if this introduce an ounce of cognitive overload, then absolutely.

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -15,13 +15,49 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
...
+
+        if (Array.isArray(value) && newObj[key] && newObj[key].length >= 1) {
+          /* eslint-disable no-plusplus */
+          for (let i = 0; i < newObj[key].length; i++) {
+            if (
+              !isNil(newObj[key][i]) &&
+              typeof newObj[key][i] === 'object' &&
+              !isNil(newObj[key][i].id)
+            ) {
+              /* eslint-disable eqeqeq */
```
adnasa, line 33: what's this sourcery

daern91: I also thought it looked funny, turns out it's valid: https://eslint.org/docs/rules/eqeqeq

packages/sync-actions/src/utils/copy-empty-array-props.js
adnasa: what if we `Number(foo) === Number(bar)` and get rid of the eslint-disable. would that help?
I recommend you test it.

MarkusAzer: Hmm, yep, that definitely will work, Just another dynamic types 😄

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -16,12 +18,14 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
...
-          return {
-            ...merged,
-          }
+        if (key === CUSTOM) {
+          return merged
         }
 
+        /**
+         * Check if `oldObj` value is an array and that same array exist on `newObj` with more than one element
+         * then Loop over `newObj` array to copy empty array props from `oldObj` to `newObj`
```
adnasa, line 27: maybe discuss a real life use case to describe what this solves would help describe part of the feature as well.

MarkusAzer: so you mean, it would be better to explain the reason of the code, something like `Loop through the elements in newObj and make sure to copy undefined array props from oldObj to ignore errors when comparing arrays`

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -4,7 +4,8 @@ const CUSTOM = 'custom'
 
 /**
  * @function copyEmptyArrayProps
- * @description Takes two objects and if there are Array props in oldObj which doesn't exist in newObj, then copy it with an empty value.
+ * @description Create new key with empty array value on `newobj` for the arrays exist on `oldObj` and doesnt exist on `newobj`
+ * One use case is to easly compare two object without generating this error `Cannot read property '0' of undefined`
```
adnasa, line 8:
```suggestion
 * One use case is to easily compare two object without generating this error `Cannot read property '0' of undefined`
```

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -18,14 +19,10 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
           ...nextObject,
         }
 
-        if (key === CUSTOM) {
-          return merged
-        }
+        // Ignore CUSTOM key as this object is dynamic and its up to the user to dynamically change it
+        // todo, i would be better if we pass it as ignored keys param
```
adnasa, line 23: 👍

PR #1581, opened 2020-07-04
fix(sdk-auth): to encode request body

tdeekens:
#### Summary

This pull request fixes the `sdk-auth` package to encode the request's body at all times.

#### Description

The `sdk-auth` is used to perform request authorization in our Node.js SDK. For that it often uses parameters such as `scope`, `refresh_token` or `grant_type` among others.

The parameters mentioned above can often contain strings which need encoding. This holds true for e.g. the `<scope>:<projectKey>` or also the `token` itself.

So far the `sdk-auth` just concatinated strings to build up the request. For example as `body += `&scope=${scope}`. Then sometimes it encodes a parameter as `request.body += `&refresh_token=${encodeURIComponent(token)}`. Our APIs can deal with this inconsistency but it can also cause problems when a string is not rightfully detected as not encoded. This is what I think @cneijenhuis noticed when having problems with token encoding and investigating it.

As a result of this this pull request suggests to always encode the body's paramters as it's send as `      'Content-Type': 'application/x-www-form-urlencoded'`.

There are usually two options going about it:

1. Use a library
2. Use `URLSearchParams` (web standard)

I propose to use [qss](https://github.com/lukeed/qss/blob/master/src/index.js) which is a tiny library which works both in Node.js and the browser. We have good experience with it in the Merchant Center.
Using `URLSearchParams` would need a polyfill as it's not supported in all browsers. Usually leading to larger bundles while we do not need most of it's functionality/API surface.

The pull request also updates the tests affected by this which outlines the change.

This overall instead of for example:

```
grant_type=client_credentials&scope=view_products:sample-project manage_types:sample-project
```

gives us

```
grant_type=client_credentials&scope=view_products%3Asample-project%20manage_types%3Asample-project
```

### DIFF
--- a/packages/sdk-auth/src/auth.js
+++ b/packages/sdk-auth/src/auth.js
@@ -9,6 +9,7 @@ import type {
   AuthRequest,
   UserAuthOptions,
 } from 'types/sdk'
+import { decode, encode } from 'qss'
 import defaultsDeep from 'lodash.defaultsdeep'
 import { getErrorByCode } from '@commercetools/sdk-middleware-http'
 import * as constants from './constants'
@@ -106,11 +107,28 @@ export default class SdkAuth {
       config.token || SdkAuth._encodeClientCredentials(credentials)
     const authType = config.authType || constants.DEFAULT_AUTH_TYPE
 
-    let body = `grant_type=${grantType}`
-    if (grantType !== 'refresh_token') body += `&scope=${scope}`
-    if (disableRefreshToken === true) body += '&refresh_token=false'
+    const isNotRefreshTokenGrantType = grantType !== 'refresh_token'
+    const initialBody = encode({
+      grant_type: grantType,
+      ...(disableRefreshToken && { refresh_token: false }),
+      ...(isNotRefreshTokenGrantType && { scope }),
+    })
 
-    return { basicAuth, authType, uri, body, headers }
+    return { basicAuth, authType, uri, body: initialBody, headers }
+  }
+
+  static _appendToRequestBody(
+    request: AuthRequest,
+    toAppend: Object
+  ): AuthRequest {
+    const previousDecodedRequestBody = request.body ? decode(request.body) : {}
+    const nextEncdedRequestBody = encode({
+      ...previousDecodedRequestBody,
+      ...toAppend,
+    })
+    request.body = nextEncdedRequestBody
+
+    return request
   }
 
   _process(request: AuthRequest) {
@@ -159,29 +177,25 @@ export default class SdkAuth {
     return SdkAuth._parseResponseJson(response).then((jsonResponse) => {
       if (SdkAuth._isErrorResponse(response))
         throw SdkAuth._createResponseError(jsonResponse, uri, response.status)
+
       return SdkAuth._enrichTokenResponse(jsonResponse)
     })
   }
 
   static _appendUserCredentialsToBody(
-    body: string,
+    request: AuthRequest,
     username: string,
     password: string
-  ): string {
+  ): AuthRequest {
     if (!(username && password))
       throw new Error('Missing required user credentials (username, password)')
 
-    return [
-      body,
-      `username=${encodeURIComponent(username)}`,
-      `password=${encodeURIComponent(password)}`,
-    ]
-      .filter(Boolean)
-      .join('&')
+    return SdkAuth._appendToRequestBody(request, { username, password })
   }
 
   static _enrichUriWithProjectKey(uri: string, projectKey: ?string): string {
     if (!projectKey) throw new Error('Missing required option (projectKey)')
+
     return uri.replace('--projectKey--', projectKey)
   }
 
@@ -216,7 +230,7 @@ export default class SdkAuth {
 
   anonymousFlow(anonymousId: string = '', config: CustomAuthOptions = {}) {
     const _config = this._getRequestConfig(config)
-    const request = SdkAuth._buildRequest(
+    let request = SdkAuth._buildRequest(
       _config,
       SdkAuth._enrichUriWithProjectKey(
         this.ANONYMOUS_FLOW_URI,
@@ -224,13 +238,18 @@ export default class SdkAuth {
       )
     )
 
-    if (anonymousId) request.body += `&anonymous_id=${anonymousId}`
+    if (anonymousId)
+      request = SdkAuth._appendToRequestBody(request, {
+        anonymous_id: anonymousId,
+      })
+
     return this._process(request)
   }
 
   clientCredentialsFlow(config: CustomAuthOptions = {}) {
     const _config = this._getRequestConfig(config)
     const request = SdkAuth._buildRequest(_config, this.BASE_AUTH_FLOW_URI)
+
     return this._process(request)
   }
 
@@ -240,13 +259,9 @@ export default class SdkAuth {
     url: string
   ) {
     const { username, password } = credentials || {}
-    const request = SdkAuth._buildRequest(config, url, 'password')
+    let request = SdkAuth._buildRequest(config, url, 'password')
 
-    request.body = SdkAuth._appendUserCredentialsToBody(
-      request.body,
-      username,
-      password
-    )
+    request = SdkAuth._appendUserCredentialsToBody(request, username, password)
 
     return this._process(request)
   }
@@ -269,6 +284,7 @@ export default class SdkAuth {
     config: CustomAuthOptions = {}
   ) {
     const _config = this._getRequestConfig(config)
+
     return this._passwordFlow(credentials, _config, this.BASE_AUTH_FLOW_URI)
   }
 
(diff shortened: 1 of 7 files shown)


packages/sdk-auth/src/auth.js
```diff
@@ -106,11 +107,28 @@ export default class SdkAuth {
...
 
-    let body = `grant_type=${grantType}`
-    if (grantType !== 'refresh_token') body += `&scope=${scope}`
-    if (disableRefreshToken === true) body += '&refresh_token=false'
+    const isNotRefreshTokenGrantType = grantType !== 'refresh_token'
+    const initialBody = encode({
+      grant_type: grantType,
+      ...(disableRefreshToken && { refresh_token: false }),
+      ...(isNotRefreshTokenGrantType && { scope }),
+    })
```
tdeekens, line 115: Building up the initial request encoding all params.

packages/sdk-auth/src/auth.js
```diff
@@ -106,11 +107,28 @@ export default class SdkAuth {
...
+    toAppend: Object
+  ): AuthRequest {
+    const previousDecodedRequestBody = request.body ? decode(request.body) : {}
+    const nextEncdedRequestBody = encode({
+      ...previousDecodedRequestBody,
+      ...toAppend,
+    })
+    request.body = nextEncdedRequestBody
+
+    return request
```
tdeekens, line 131: A small helper reused within the module just to append to the request cause that's what we do most of the time here.

packages/sdk-auth/src/auth.js
```diff
@@ -106,11 +107,28 @@ export default class SdkAuth {
...
+    })
 
-    return { basicAuth, authType, uri, body, headers }
+    return { basicAuth, authType, uri, body: initialBody, headers }
+  }
+
+  static _appendToRequestBody(
+    request: AuthRequest,
+    toAppend: Object
+  ): AuthRequest {
```
tdeekens, line 123: We return the request not just the body here. That makes this easier to compose from my perspective.

packages/sdk-auth/src/auth.js
```diff
@@ -216,21 +230,26 @@ export default class SdkAuth {
 
   anonymousFlow(anonymousId: string = '', config: CustomAuthOptions = {}) {
     const _config = this._getRequestConfig(config)
-    const request = SdkAuth._buildRequest(
+    let request = SdkAuth._buildRequest(
```
tdeekens, line 233: We build the request, then append to its body and get a new request. That's what I mean by composing.

tdeekens:
@daern91 please assign anybody else if needed. You seem to be most familiar with this however so maybe a good reviewer 😄.

packages/sdk-auth/src/auth.js
```diff
@@ -106,11 +107,28 @@ export default class SdkAuth {
...
-    return { basicAuth, authType, uri, body, headers }
+    return { basicAuth, authType, uri, body: initialBody, headers }
+  }
+
+  static _appendToRequestBody(
+    request: AuthRequest,
+    toAppend: Object
+  ): AuthRequest {
+    const previousDecodedRequestBody = request.body ? decode(request.body) : {}
+    const nextEncdedRequestBody = encode({
```
daern91, line 125: nitpick typo:
~~nextEncdedRequestBody~~
nextEnc**o**dedRequestBody

packages/sdk-auth/src/auth.js
daern91: I like this change a lot 👍

packages/sdk-auth/src/auth.js
```diff
@@ -159,29 +177,25 @@ export default class SdkAuth {
...
     })
   }
 
   static _appendUserCredentialsToBody(
-    body: string,
+    request: AuthRequest,
     username: string,
     password: string
-  ): string {
+  ): AuthRequest {
```
daern91, line 189: Here too 👍

packages/sdk-auth/src/auth.js
```diff
@@ -216,21 +230,26 @@ export default class SdkAuth {
...
       _config,
       SdkAuth._enrichUriWithProjectKey(
         this.ANONYMOUS_FLOW_URI,
         _config.projectKey
       )
     )
 
-    if (anonymousId) request.body += `&anonymous_id=${anonymousId}`
+    if (anonymousId)
+      request = SdkAuth._appendToRequestBody(request, {
```
daern91, line 242: I guess we could technically leave `request` as a constant and return here instead of re-assigning and returning below. Just a matter of flavor tho.

packages/sdk-auth/src/auth.js
```diff
@@ -240,13 +259,9 @@ export default class SdkAuth {
...
     const { username, password } = credentials || {}
-    const request = SdkAuth._buildRequest(config, url, 'password')
+    let request = SdkAuth._buildRequest(config, url, 'password')
 
-    request.body = SdkAuth._appendUserCredentialsToBody(
-      request.body,
-      username,
-      password
-    )
+    request = SdkAuth._appendUserCredentialsToBody(request, username, password)
```
daern91, line 264: same with this one, but might just make things too nested and unreadable.

packages/sdk-auth/src/auth.js
```diff
@@ -313,12 +329,12 @@ export default class SdkAuth {
...
 
-    const request = SdkAuth._buildRequest(_config, uri)
+    let request = SdkAuth._buildRequest(_config, uri)
     request.body = body || '' // let user to build their own body
 
     if (credentials)
-      request.body = SdkAuth._appendUserCredentialsToBody(
-        request.body,
+      request = SdkAuth._appendUserCredentialsToBody(
+        request,
```
daern91, line 337: I see, we keep this pattern throughout 👍

daern91:
Adding @commercetools/clients-team as FYI

packages/sdk-auth/src/auth.js
tdeekens: Yes, didn't want to change such things much in this PR. Generally the composition in this code can be improve. Ideally, the `AuthRequeust` could have a `map` and `value` function. Then you could just `request.map(SdkAuth._appendToRequestBody(...))` which gives you a mapped request and then `value()` when you pass it somewhere. That's another story though.

packages/sdk-auth/src/auth.js
tdeekens: The conditionals are a bit hard to improve without better patterns of compositions. I will adjust those which don't involve them.

packages/sdk-auth/src/auth.js
tdeekens: Thanks. Feel free to use GitHub's review suggestions next time. Helps to get better diffs of what you carefully markdowned yerself :)

cneijenhuis:
Encoding LGTM 👍

tdeekens:
> Would also be nice to have the ability to use the, supposedly faster, querystring.stringify native solution in node. As described here: lukeed/qss#qss-

That's a Node.js utility. We also run in the browser. Given we encode three things - as you mentioned - I don't see this, comparing to other impacts, relevant for performance of this lib. With 100k ops yes, not with 3.

> Thanks for this one. Much cleaner now with a proper lib and not the hacky template strings 👍

I am still not super happy with this. Encapsulation is sub-optimal:

1. The `AuthRequest` should have a `map` and `value` and sort of act as a Functor
2. The encoding chould be "inside of" it. Not part of the surrouding code

Still this is better than before and step-wise is how to improve things 🐱.

daern91:
@jenschude any comments on this one before we merge?

packages/sdk-auth/test/client-password-flow.spec.js
```diff
@@ -40,9 +44,12 @@ describe('Client Password flow', () => {
       .post(
         `/oauth/token`,
         // expected body
-        `grant_type=password&scope=manage_project:${config.projectKey}` +
-        `&username=user%204%5El*aJ%40ETso%2B%2F%5CHdE1!x0u4q5` + // encoded
-          `&password=pass%204%5El*aJ%40ETso%2B%2F%5CHdE1!x0u4q5` // encoded
+        encode({
```
jenschude, line 47: Also it may make the test more readable you are testing the encoding with the function to encode itself. I'm pretty sure that this works correct but a non working the encode function would result in a still working test.

But I also can see that the encode function may not produce a canonical string to compare it to a static string.

tdeekens: Good point. Let me suggest something later :)

jenschude: To be on the safe side you could use https://nodejs.org/api/querystring.html#querystring_querystring_encode

tdeekens: Yet another good point 💋. Thanks for being persistent. Here we go: 2913cfe.

tdeekens: I take the thumbs up as an approval.

PR #1589, opened 2020-07-15
feat(sync-actions): publish product actions

markus-azer:
#### Description

Give the ability to publish product actions directly by simply mark `staged` flag to false.

One use case for this, it would make it easy for package user to just publish the updates instead of sending update actions then send another action to publish product updates plus this feature would shorten the platform calls

### DIFF
--- a/packages/sync-actions/src/products.js
+++ b/packages/sync-actions/src/products.js
@@ -43,6 +43,7 @@ function createProductMapActions(
   ): Array<UpdateAction> {
     const allActions = []
     const { sameForAllAttributeNames } = options
+    const { publish } = newObj
 
     const variantHashMap = findMatchingPairs(
       diff.variants,
@@ -128,6 +129,9 @@ function createProductMapActions(
         productActions.actionsMapAssets(diff, oldObj, newObj, variantHashMap)
       )
     )
+
+    if (publish === true)
+      return flatten(allActions).map((action) => ({ ...action, staged: false }))
 
     return flatten(allActions)
   }
--- a/packages/sync-actions/test/product-sync-base.spec.js
+++ b/packages/sync-actions/test/product-sync-base.spec.js
@@ -102,6 +102,26 @@ describe('Actions', () => {
     expect(actions).toEqual([{ action: 'changeName', ...now }])
   })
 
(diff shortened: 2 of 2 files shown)


packages/sync-actions/src/products.js
```diff
@@ -129,6 +130,9 @@ function createProductMapActions(
       )
     )
 
+    if (publish === true)
+      return flatten(allActions).map((action) => ({ ...action, staged: false }))
```
daern91, line 134: Will this also work for the [setKey](https://docs.commercetools.com/http-api-projects-products#set-key) and [setTaxCategory](https://docs.commercetools.com/http-api-projects-products#set-taxcategory) actions? 

I see they are in the list here
https://github.com/commercetools/nodejs/blob/2008434e12e469d4d035ed4658cff51069ba7985/packages/sync-actions/src/product-actions.js#L26
and here
https://github.com/commercetools/nodejs/blob/2008434e12e469d4d035ed4658cff51069ba7985/packages/sync-actions/src/product-actions.js#L36

but am not sure how it works.

markus-azer: Thats good question, i thought its only `setkey` that doesnt have this flag and i already tested it and it worked, seems that platform ignored  this flag, i should also try it with `setTaxCategory `

packages/sync-actions/src/products.js
```diff
@@ -43,6 +43,7 @@ function createProductMapActions(
   ): Array<UpdateAction> {
     const allActions = []
     const { sameForAllAttributeNames } = options
+    const { publish } = newObj
```
markus-azer, line 46: regarding this flag naming, i thought that `publish` is more clear than using `staged`.

packages/sync-actions/src/products.js
daern91: Sounds good to me. And we don't have integration tests, right? So I guess it's all up to us to manually test it and make sure the platform doesn't throw on us.

markus-azer: yep so far we dont have integration tests, and i just tried it with `setTaxCategory` and it worked fine

PR #1594, opened 2020-07-22
test(sync-actions): add test for empty categoryOrderHints

markus-azer:
#### Summary

Fix generating wrong actions for categoryOrderHints

#### Description

When passing `before` with empty `categoryOrderHints` while `categoryOrderHints` doesnt exist on `now` it generate wrong actions.


markus-azer:
> To me it looks dangerous to completely ignore changes on `categoryOrderHints`. Can you explain the use case why this change was necessary?

yes, totally agree its dangerous change, just want to mention, its not ignoring `categoryOrderHints`, `categoryOrderHints` is an object prop, so according to [jsondiffpatch docs](https://www.npmjs.com/package/jsondiffpatch), the diff should also be object containing props changes, in case of array, it means for ex: `oldobj` have empty obj for `categoryOrderHints` while `newobj` doesnt have `categoryOrderHints` and in case not ignoring this array, it generate very wrong actions.
regarding usecase, there is this new test.

packages/sync-actions/src/product-actions.js
```diff
@@ -490,6 +490,8 @@ export function actionsMapCategories(diff) {
 
 export function actionsMapCategoryOrderHints(diff) {
   if (!diff.categoryOrderHints) return []
+  // Ignore this pattern as its means no changes happened [{},0,0]
```
katmatt, line 493: The `[{},0,0]` in this comment confused me a bit and I didn't see this anywhere else in this project. But we can keep i!

PR #1601, opened 2020-08-18
refactor(sync-actions): rename terminology to allow/ignore

emmenko:
I think there is no need to explain this.


emmenko:
Hmm any idea why the integration tests are failing with this error?

```
npm ERR! cb.apply is not a function
```

emmenko:
It seems to happen also in other PRs...

emmenko:
Looks like it's an issue with node: https://github.com/nodejs/help/issues/2874

daern91 (approved):

PR #1603, opened 2020-08-18
chore: upgrade flowtype to 0.131.0 and yarn to 1.22.4

emmenko:
Some maintenance work:

- upgrade to flowtype `0.131.0`. This comes with some overdue required changes, like https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb
- upgrade yarn bin to `1.22.4`


packages/product-json-to-csv/src/main.js
```diff
@@ -197,7 +197,7 @@ export default class ProductJsonToCsv {
     )
   }
 
-  async _resolveVariantReferences(variant: Variant) {
+  async _resolveVariantReferences(variant: Variant): Promise<Variant> {
```
emmenko, line 200: Flow requires exported functions/interfaces to have an explicit return type declaration

```
Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type annotation at function return:Flow(signature-verification-failure)
```

types/resourceDeleter.js
```diff
@@ -23,7 +23,7 @@ export type ApiConfigOptions = {
 
 /* Client */
 export type CustomClientResult = ClientResult & {
-  id?: string,
+  id: string,
```
emmenko, line 26: This should have been required, as far as I can see in the code

packages/sdk-auth/src/auth.js
```diff
@@ -17,7 +16,7 @@ import * as constants from './constants'
 export default class SdkAuth {
   // Set flowtype annotations
   config: AuthOptions
-  fetcher: ConfigFetch
+  fetcher: typeof fetch
```
emmenko, line 19: We should be able to use the built-in type declarations for the fetch object.

Because of that, we can remove all our custom type declarations.

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -206,7 +206,7 @@ describe('Http', () => {
         expect(res.request).toEqual(
           expect.objectContaining({
             headers: {
-              'Content-Type': ['application/json'],
+              'Content-Type': 'application/json',
```
emmenko, line 209: I couldn't recall why we were using an array of values for some headers...I think it makes sense to keep it simple and declare it as a single value

packages/product-json-to-csv/src/main.js
katmatt: Makes somehow sense to be explicit about the return type.

packages/sdk-middleware-http/test/http.spec.js
katmatt: Your argument makes sense to me too 👍

.flowconfig
```diff
@@ -55,8 +55,24 @@ packages/*/src
...
+esproposal.nullish_coalescing=enable
+
+server.max_workers=2
+
+module.name_mapper='^types/\(.*\)$' -> '<PROJECT_ROOT>/types/\1.js'
+
+suppress_type=$FlowIssue
+suppress_type=$FlowFixMe
+
+# https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb
```
daern91, line 76: Thanks a lot for adding the links/context

packages/sdk-auth/src/auth.js
```diff
@@ -53,12 +52,24 @@ export default class SdkAuth {
...
+    if (fetcher) {
+      fetchFunction = fetcher
+    } else {
+      // `fetcher` is set here rather than the destructuring to ensure fetch is
+      // declared before referencing it otherwise it would cause a `ReferenceError`.
+      // For reference of this pattern: https://github.com/apollographql/apollo-link/blob/498b413a5b5199b0758ce898b3bb55451f57a2fa/packages/apollo-link-http/src/httpLink.ts#L49
+
+      /* eslint-disable */
+      fetchFunction = fetch
+      /* eslint-enable */
```
daern91, line 70: do we disable here because of the reassignment? Doesn't flow allow casting, maybe?

packages/sdk-middleware-http/src/http.js
```diff
@@ -106,27 +111,33 @@ export default function createHttpMiddleware({
...
+    if (timeout || abortController) {
+      fetchOptions.signal = abortController.signal
+    }
+    if (body) {
+      fetchOptions.body = body
     }
     let retryCount = 0
     // wrap in a fn so we can retry if error occur
     function executeFetch() {
-      // $FlowFixMe
```
daern91, line 128: super nice to see this one removed

daern91 (approved):
LGTM 👍 Thx a lot for putting in the effort here

PR #1604, opened 2020-08-18
chore: use latest node versions

emmenko:
This should hopefully fix the issue we're currently having on CI when installing tar packages:

```
cb.apply is not a function
```

See related issue: https://github.com/nodejs/help/issues/2874


emmenko:
Alright, this seems to fix the issue. Merging to unblock master and the other PRs.

PR #1618, opened 2020-09-11
[fix] abort controller issue

hakobpogh:
#### Summary

the current version of `createHttpMiddleware` requires a single abortController and uses that for all requests that's why when a single request is getting aborted then all the requests are aborted with it. So my current change is adding a function called getAbortController which returns a new AbortController each time. And I've moved to get the abort controller to the function returned from the `createHttpMiddleware`.

#### Description

Most probably you'll need to make the abortController property deprecated but I left that decision on you :D.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


packages/sdk-middleware-http/src/http.js
```diff
@@ -68,14 +68,15 @@ export default function createHttpMiddleware({
...
 }: HttpMiddlewareOptions): Middleware {
   if (!fetcher && typeof fetch === 'undefined')
     throw new Error(
       '`fetch` is not available. Please pass in `fetch` as an option or have it globally available.'
     )
-  if (timeout && !_abortController && typeof AbortController === 'undefined')
+  if (timeout && !getAbortController && !_abortController && typeof AbortController === 'undefined')
     throw new Error(
-      '`AbortController` is not available. Please pass in `AbortController` as an option or have it globally available when using timeout.'
+      '`AbortController` is not available. Please pass in `getAbortController` as an option or have AbortController globally available when using timeout.'
```
jenschude, line 79: This change will fail this test

https://github.com/commercetools/nodejs/blob/693f630b2c8700bcb68d2c79048b075c92ca558f/packages/sdk-middleware-http/test/http.spec.js#L34-L42

jenschude (commented):
As the change is made backwards compatible it should be fine. Could you please add an additional test with the callback to http.spec.js

jenschude:
I fixed a small issue and added 2 tests for the timeout with the getAbortController function

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -134,6 +166,38 @@ describe('Http', () => {
...
     }))
 
+  test('execute a get request with getAbortController short timeout (fail)', () =>
+    new Promise((resolve, reject) => {
+      const request = createTestRequest({
+        uri: '/foo/bar',
+      })
+      const response = { resolve, reject }
+      const next = (req, res) => {
+        expect(res).toEqual({
```
tdeekens, line 176: With these "delayed" `expect` blocks it could be useful to add a `expect.assertions(1)` on L170 to ensure that the expectation is actually called and we don't silently pass the test.

tdeekens (commented):
🙇

packages/sdk-middleware-http/test/http.spec.js
```diff
@@ -102,6 +102,38 @@ describe('Http', () => {
       httpMiddleware(next)(request, response)
     }))
 
+  test('execute a get request with getAbortController timeout (success)', () =>
```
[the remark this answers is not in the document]
jenschude: Both tests are c&p from the other ones. Would like to have them in the same style else it would make sense to adjust all.

packages/sdk-middleware-http/test/http.spec.js
jenschude: Added the expect.assertions

packages/sdk-middleware-http/test/http.spec.js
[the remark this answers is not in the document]
tdeekens: Tru dat.

tdeekens (approved):

hakobpogh:
Hey @tdeekens @jenschude. I'm sorry I've been inactive for a few days. I've noticed that you've made some changes there. Is there anything else I can do? Also, what should I do to get this PR merged?

Issue #1627, opened 2020-09-28
`customerId` and `cartId` params not being reset between calls to `build`


jimmythomson:
### Description

When using `byCustomerId` or `byCartId` in `api-request-builder`, the locally stored params are not cleared before the next call to `build`. 

### Expected Behavior

The `setDefaultParams` function should reset the `customerId` and `cartId` params.

### Current Behavior

The `customerId` and `cartId` params are not cleared between calls to `build`.

### Context

This bug was discovered when attempting to build multiple queries, one after the other.

### Possible Solution

Will create a PR shortly.

PR #1628, opened 2020-09-28
fix(api-request-builder): resetting `customerId` and `cartId` params

jimmythomson:
Fixes: #1627

Resets the `customerId` and `cartId` params to null after the `build` method has generated a uri.


daern91 (approved):
LGTM, thanks a lot for the contribution! 💯

PR #1629, opened 2020-09-29
small linter fixes

jenschude:
#### Summary

Fixes the broken linting on CircleCI

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -92,7 +92,7 @@ export default function createHttpMiddleware({
     request: MiddlewareRequest,
     response: MiddlewareResponse
   ) => {
-    let abortController
+    let abortController: any
     if (timeout || getAbortController || _abortController)
       // eslint-disable-next-line
       abortController = (getAbortController ? getAbortController(): null) || _abortController || new AbortController()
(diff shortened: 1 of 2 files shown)


packages/sdk-middleware-http/src/http.js
```diff
@@ -120,7 +120,7 @@ export default function createHttpMiddleware({
     if (credentialsMode) {
       fetchOptions.credentials = credentialsMode
     }
-    if (timeout || abortController) {
+    if (abortController) {
```
jenschude, line 123: FYI: if timeout is set, the abortController will be initialized anyway

daern91: Good catch 👍

daern91 (approved):
LGTM! Thanks a lot for the quick fix 🥇

packages/sdk-middleware-http/src/http.js
```diff
@@ -92,7 +92,7 @@ export default function createHttpMiddleware({
     request: MiddlewareRequest,
     response: MiddlewareResponse
   ) => {
-    let abortController
+    let abortController: any
```
katmatt, line 95: This is TypeScript and I'm wondering why JavaScript accepts it at all?

jenschude: Cause the node SDK uses flow syntax ;)

Issue #1632, opened 2020-10-12
sdk-middleware-logger: Customise http request-response logs in createLoggerMiddleware


VineetKumarKushwaha:
### https://jira.commercetools.com/browse/SUPPORT-9830

VineetKumarKushwaha:
As per the one of the devlopers this is not needed. Please check this https://jira.commercetools.com/browse/SUPPORT-9830

PR #1633, opened 2020-10-12
feat(sdk-middleware-logger): support custom logging

VineetKumarKushwaha:
#### support custom logging

Fix: #1632

### DIFF
--- a/packages/sdk-middleware-logger/src/logger.js
+++ b/packages/sdk-middleware-logger/src/logger.js
@@ -6,16 +6,29 @@ import type {
   Next,
 } from 'types/sdk'
 
-export default function createLoggerMiddleware(): Middleware {
+type CallbackFn = (
+  request: MiddlewareRequest,
+  response: MiddlewareResponse
+) => void
+
+const defaultCb = (
+  request: MiddlewareRequest,
+  response: MiddlewareResponse
+) => {
+  const { error, body, statusCode } = response
+  console.log('Request: ', request)
+  console.log('Response: ', { error, body, statusCode })
+}
+
+export default function createLoggerMiddleware(
+  cb: CallbackFn = defaultCb
+): Middleware {
   return (next: Next): Next => (
     request: MiddlewareRequest,
     response: MiddlewareResponse
   ) => {
-    const { error, body, statusCode } = response
-    /* eslint-disable */
-    console.log('Request: ', request)
-    console.log('Response: ', { error, body, statusCode })
-    /* eslint-enable */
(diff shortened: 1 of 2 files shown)


packages/sdk-middleware-logger/src/logger.js
```diff
@@ -6,13 +6,19 @@ import type {
...
   return (next: Next): Next => (
     request: MiddlewareRequest,
     response: MiddlewareResponse
   ) => {
     const { error, body, statusCode } = response
     /* eslint-disable */
+    // Pass function to customize logging.
+    if (cb) {
+      cb(request, response)
+    }
```
daern91, line 21: Hmm, this should work but I'm wondering why the customer wouldn't just implement their own logging middleware instead?

packages/sdk-middleware-logger/src/logger.js
```diff
@@ -6,13 +6,19 @@ import type {
...
     console.log('Request: ', request)
     console.log('Response: ', { error, body, statusCode })
```
daern91, line 23: Do they still want these to execute too? Even if they pass in their own logic?

daern91 (commented):
Hi Vineet, thanks for the PR. I left some questions below.

packages/sdk-middleware-logger/src/logger.js
jenschude: I think it's still fine to allow custom logging in this middleware, but then I would also create a default callback function which logs to the console

jenschude (changes requested):

VineetKumarKushwaha:
I have resolved the comments.

emmenko (changes requested):
The way it's implemented now does not make much sense to me. You can simply create your own middleware then, which is equivalent to pass a callback function.

VineetKumarKushwaha:
@emmenko this was the requirement so I implemented like that. Could you please suggest me some other approach ?

emmenko:
Well simply create your own custom middleware logger. There is no point in customizing the default logger middleware.

emmenko:
This is the all point of a middleware, you can create your own that fits your needs. See https://commercetools.github.io/nodejs/sdk/Middlewares.html

VineetKumarKushwaha:
@emmenko  could you suggest customer your view here https://jira.commercetools.com/browse/SUPPORT-9830.

emmenko:
I left a comment. You can close this PR.

VineetKumarKushwaha:
@emmenko Thanks

Issue #1634, opened 2020-10-31
Add ga ip anonymization to the github pages


VineetKumarKushwaha:
### Add ga ip anonymization to the github pages

PR #1635, opened 2020-10-31
1634 Github Pages: add ip anonymization to the ga

VineetKumarKushwaha:
Fixes: #1634

### DIFF
--- a/book.json
+++ b/book.json
@@ -19,6 +19,9 @@
       "url": "https://github.com/commercetools/nodejs/"
     },
     "ga": {
+      "configuration": {
+        "anonymizeIp": true
+      },
       "token": "UA-38285631-20"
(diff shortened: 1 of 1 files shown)


book.json
```diff
@@ -19,6 +19,7 @@
       "url": "https://github.com/commercetools/nodejs/"
     },
     "ga": {
+      "anonymizeIp": true,
```
emmenko, line 22: From what I understand it should be like this:

```suggestion
      "configuration": {
        "anonymizeIp": true,
      },
```

https://www.npmjs.com/package/gitbook-plugin-ga

emmenko (changes requested):

emmenko (approved):

PR #1648, opened 2021-01-16
fix: bump node js version

VineetKumarKushwaha:
### Current node js version has some vulnerability so bumped to the non vulnerable version

Fix: #1647


VineetKumarKushwaha:
@daern91 @ashishhk I checked but didn't get anything why this test cases are failing. I can see the test cases are failing on node 12 on master as well. I tried to update the circle node 12 version but then it started throwing some new node related errors. So Can we merge this as this change does not breaking anything new ?  we can debug the test cases issue later as I can see it may need more debugging time.

ashishhk (approved):
Thank you for updaing.

ashishhk:
@daern91 I don't see a issue merging this PR since the master was already failing with the test case, WDYT ?

daern91:
This PR #1654 is also blocked by the tests being red on master. So I think it would be best to unblock master first then merge these two... just to make sure we're not piling on more bugs/failing tests with the upcoming PRs.

daern91 (approved):

PR #1662, opened 2021-03-09
fix(sync-actions): changing order of actions

danrleyt:
#### Summary
We had this error reported by one of our customers of the Import-API

While changing the master variant of a given product, the sdk is generating a removeVariant action (trying to remove the master variant) before it has been actually changed, which results in an error coming from the platform saying that it is not possible to remove the master variant. 

The behavior should be. First, change the master variant, then remove it. 

#### Description

This PR changes the order of actions to:
1. AddVariant
2. ChangeMasterVariant (if any)
3. RemoveVariant

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR


tdeekens (approved):
Makes sense to me. Thanks.

daern91 (approved):
Very nice, thanks a lot 🙏

PR #1668, opened 2021-03-12
1667 personal data eraser timeout with sorting

### LINKED ISSUE #1667: Personal Data Eraser Timeout With Sorting
### https://jira.commercetools.com/browse/SUPPORT-11559

daern91:
#### Summary

PR introduces `disableSort` flag to the `client.process` method. This one is to be used when deleting larger amounts of data where we don't need to pass in a sort param as it just slows down the db through scans.

Please let me know if you think that introducing this option is too dangerous, as misusing it could introduce infinite loops with lots of requests to the CTP. 

An alternative solution would be that we implement a separate loop within the `personal-data-eraser` lib instead of using `client.process`.

resolves #1667

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [x] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/docs/sdk/api/sdkClient.md
+++ b/docs/sdk/api/sdkClient.md
@@ -69,7 +69,6 @@ Returns a `Promise` with the accumulated result of each `processFn` calls.
 - `options` _(Object)_
   - `accumulate` _(Boolean)_: (default `true`) a flag to indicate whether all the results of the iterations should be accumulated. This is useful if you want to e.g. fetch all the entities of an API endpoint and do something with it at the end. _Be careful that this might lead to memory problems if the fetched data gets too big. If it's not necessary to have all the data when the process function resolves, it's recommended to disable this option_.
   - `total` _(Number)_: a number to indicate the total amount of items to be fetched from all API calls
-  - `disableSort` _(Boolean)_: (default `false`) disables sorting during iteration, useful when deleting multiple entries. Please do not use it otherwise as it will introduce an infinite loop.
 
 #### Usage example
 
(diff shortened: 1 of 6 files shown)


ashishhk (approved):
Thanks for taking care of this on priority, looks good to me. I didn't understood the syntax completely but I just hope its an optional parameter.

docs/sdk/api/sdkClient.md
```diff
@@ -69,6 +69,7 @@ Returns a `Promise` with the accumulated result of each `processFn` calls.
 - `options` _(Object)_
   - `accumulate` _(Boolean)_: (default `true`) a flag to indicate whether all the results of the iterations should be accumulated. This is useful if you want to e.g. fetch all the entities of an API endpoint and do something with it at the end. _Be careful that this might lead to memory problems if the fetched data gets too big. If it's not necessary to have all the data when the process function resolves, it's recommended to disable this option_.
   - `total` _(Number)_: a number to indicate the total amount of items to be fetched from all API calls
+  - `disableSort` _(Boolean)_: (default `false`) disables sorting during iteration, useful when deleting multiple entries. Please do not use it otherwise as it will introduce an infinite loop.
```
cneijenhuis, line 72: I don't think this is correct - Mongo will have some sorting internally that is consistent, unless the underlying data changes. The sorting provided by Mongo may or may not be a sorting by `id`.

The more likely result is that you'll miss elements, not that you'll end in an infinite loop (as the `lastId` should increase in every iteration)?

cneijenhuis (approved):

docs/sdk/api/sdkClient.md
daern91: Ahh, very good question. Here I was trying to point out that with the PR's implementation it also won't use `lastId` but keep fetching until there's no element left to fetch. 

So the infinite loop could come from the library side - in the case where the user is not deleting every element it's getting.

Does that make sense?

jenschude: I already said to Daniel directly I don't think this is a good solution.

The best way would be to implement a distinct loop method which uses the proposed way by deleting till the result count is below the limit parameter. Even a graphql query could be used to just retrieve the id & version to save bandwidth

jenschude (approved):
Fixes the possible issue while deleting. Also now the "dry-run" differs from the actual delete. But this could be tackled separately.

PR #1678, opened 2021-04-26
fix(sync-actions): fix customer error for Cannot read property '0' of undefined

### LINKED ISSUE #1602: Error thrown when synching customer
### Description

When I call the `sync-actions` function `customerSync.buildActions` with this test data:
```js
      const before = {
        addresses: [],
      }
      const now = {
      }
      const actual = customerSync.buildActions(now, before)
```
the following error is thrown:
```
Cannot read property '0' of undefined
```

### Expected Behavior

No error should be thrown and the function should return an empty array.

### Context

We tried to import customer and then found this issue when trying to update an existing customer with an empty `addresses` array.

markus-azer:
#### Summary

fix sync customer error
#### Description

same issue like this https://github.com/commercetools/nodejs/pull/1525
Resolve #1602 

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -69,7 +69,7 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
         }
         return merged
       },
-      {}
+      { ...newObj }
     )
     return [oldObj, nextObjectWithEmptyArray]
   }
--- a/packages/sync-actions/test/customer-sync.spec.js
+++ b/packages/sync-actions/test/customer-sync.spec.js
@@ -481,4 +481,29 @@ describe('Actions', () => {
       },
     ])
   })
+
+  test('should build not throw error for empty array', () => {
+    const before = {
+      stores: [
+        {
+          typeId: 'store',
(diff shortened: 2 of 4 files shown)


packages/sync-actions/test/customer-sync.spec.js
```diff
@@ -481,4 +481,29 @@ describe('Actions', () => {
       },
     ])
   })
+
+  test('should build not throw error for empty array', () => {
```
markus-azer, line 485: this test to simulate the problem or the error

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -69,7 +69,7 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
         }
         return merged
       },
-      {}
+      { ...newObj }
```
markus-azer, line 72: this to fix bug in this function `copyEmptyArrayProps`

tdeekens (approved):
💯

jherey (approved):

PR #1679, opened 2021-04-28
Add getting started example

anilkk:
#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


stmeissner:
@anilkk If the integration with the TypeScript SDK is the approach we highly recommend, I would go for #1681 right away and drop this. WDYT?

PR #1682, opened 2021-05-04
fix(sync-actions): organize customer sync actions, ex: addAdress, then setDefaultShippingAddress

markus-azer:
#### Summary

make sure that customer sync-actions are generating in right order.

#### Description

this PR, make sure that these two actions `setDefaultShippingAddress` and `setDefaultBillingAddress` will be in order after adding new address, so ctp platform wont fire error complaining that address id not found.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/customer-actions.js
+++ b/packages/sync-actions/src/customer-actions.js
@@ -22,6 +22,14 @@ export const baseActionsList = [
   { action: 'setLocale', key: 'locale' },
   { action: 'setVatId', key: 'vatId' },
   {
+    action: 'setStores',
+    key: 'stores',
+  },
+  { action: 'setKey', key: 'key' },
+]
+
+export const setDefaultBaseActionsList = [
+  {
     action: 'setDefaultBillingAddress',
     key: 'defaultBillingAddressId',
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/customers.js
```diff
@@ -44,6 +44,17 @@ function createCustomerMapActions(
       )
     )
 
+    allActions.push(
+      mapActionGroup('base', (): Array<UpdateAction> =>
```
markus-azer, line 48: same action group so it wont be breaking change and its ordered after `addresses` actions

packages/sync-actions/src/customer-actions.js
```diff
@@ -31,11 +39,6 @@ export const baseActionsList = [
     key: 'defaultShippingAddressId',
     actionKey: 'addressId',
   },
```
markus-azer, line 41: Introduce new action list to make it easy to control the order

danrleyt (approved):

PR #1687, opened 2021-05-12
feat(sync-actions): create more order update actions

jherey:
#### Summary
Adds functionality to create update actions for order's `returnInfo` and `delivery parcel`.

#### Description
- add `returnInfo` update actions
- add `parcel` update actions
- write tests for both update actions

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/order-actions.js
+++ b/packages/sync-actions/src/order-actions.js
@@ -84,7 +84,6 @@ function _buildDeliveryParcelsAction(
     }
 
     if (isRemoveAction(key, parcel)) {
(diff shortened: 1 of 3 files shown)


markus-azer (commented):
seems this PR needs more tests, maybe add tests to make sure there is no action generated in case of same length and other test could be test it with empty arrays, wdyt?

tdeekens (approved):

PR #1702, opened 2021-06-18
test(sync-action): test adding newly master variant

markus-azer:
#### Summary

- test to make sure that when adding newly master variant, it will be added as first `addVariant` action.

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/test/product-sync-variants.spec.js
+++ b/packages/sync-actions/test/product-sync-variants.spec.js
@@ -634,6 +634,71 @@ describe('Actions', () => {
     })
   })
 
+  describe('adding newly master variant', () => {
+    const before = {
+      id: '123',
+      version: 1,
+      masterVariant: {
+        id: 1,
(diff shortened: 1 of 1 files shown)


packages/sync-actions/test/product-sync-variants.spec.js
```diff
@@ -634,6 +634,71 @@ describe('Actions', () => {
...
+        id: 2,
+        sku: 'v2',
+        attributes: [{ name: 'fooo', value: 'baar' }],
+      },
+    }
+
+    test('should add newly added masterVariant as first `addVariant` ', () => {
+      const actions = productsSync.buildActions(now, before)
+
+      expect(actions[0]).toEqual({
```
islam3zzat, line 672: Is this redundant since we do an assertion on all the actions?

Issue #1694, opened 2021-05-21
Personal Data eraser tool-  Cart Not deleted


ashishhk:
https://jira.commercetools.com/browse/SUPPORT-12343

as per documentation it should clear any cart info too https://commercetools.github.io/nodejs/cli/personal-data-erasure.html

PR #1705, opened 2021-07-01
fix(personal-data-erasure): remove cart also while deleting all customer data

VineetKumarKushwaha:
#### Delete cart data when deleting customer data

Fix: #1694


danrleyt (approved):
💯 Thanks Vineet!

PR #1712, opened 2021-10-12
fix: retry for aborted requests

kamaz:
#### Summary

Fix retry for aborted requests. 

#### Description

The PR fixes the issues with retry on aborted requests by moving the `AbortController` to `executeFetch`. The reason for that is so that the same signal is not being used. Otherwise, retry are just aborted instantaneously. 

#### Todo

- Tests
  - [x] Unit


jenschude:
Closed with #1714

PR #1717, opened 2021-10-31
fix(discount-importer): handle exception show user correct error

VineetKumarKushwaha:
#### Discount Imported module throws exception when there is an error

Fix: https://github.com/commercetools/express-impex/issues/767


VineetKumarKushwaha:
Aborting this PR because the customer is already moved to done so no need of this for now. We can re-open it again if customer comes for asking this

PR #1722, opened 2021-11-04
Setup Changesets

ajimae:
#### Summary

This PR is sets up changesets for the `commercetools/nodejs` repository

#### Description

This is the initial attempt to setup the minimal tools and configs needed to enable changesets in this repository

#### Todo

- [ ] Migrate test workflow from CircleCI to Github workflow*
- [ ] Documentation the new release process


package.json
```diff
@@ -28,7 +28,6 @@
     "test:coverage": "yarn build && npm test -- --no-watchman --coverage",
     "test:ci": "yarn test -- --ci --no-watchman --maxWorkers=10",
     "test:coverage:ci": "yarn test:ci -- --coverage && codecov",
-    "release": "yarn check-gh-auth-token && cross-env NODE_ENV=cli yarn build && lerna publish --registry=https://registry.npmjs.org/ && yarn build:changelog && yarn push:changelog && yarn docs:publish",
```
emmenko, line 31: We should also remove those other scripts/commands that are not necessary anymore, as well as dependencies (e.g. `lerna-changelog`).

.changeset/spotty-flies-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools/sdk-middleware-auth': minor
```
emmenko, line 2: You need to make an actual change in the package, otherwise it won't work.

I would suggest to not mix it up with this PR though.

emmenko:
Can you also update the CONTRIBUTING part about releasing?

package.json
ajimae: I have removed all commands associated with lerna changelog.

emmenko: What about the `yarn docs:publish` command? Do you still need it?

emmenko (approved):
Looks good, just left a small comment

PR #1751, opened 2022-01-21
fix(sdk-middleware-http): allow to unset content-type

MicheleRezk:
#### Summary
Don't set the content-type by default to the requests with Null content type, and let the client to identify the correct content-type based on the request body

#### Description
Adjust the http middleware to check for requests with Null Content-Type, if this passed in the request headers then we don't need to set the content-type for the request to be 'application/json' and let the client identify the correct content-type based on the request body.

This is useful when the request is of type 'multipart/form-data' and it's better to let the client Identify the correct content-type & boundary and also calculate the Content-Length. 

so while calling file-upload service we will pass this Content-Type as Null in the request headers without identifying the content-type and the sdk client will add the correct content-type while calling the service.

- Tests
  - [ ] Unit

### DIFF
--- a/.changeset/orange-years-rush.md
+++ b/.changeset/orange-years-rush.md
@@ -1,0 +1,5 @@
+---
+"@commercetools/sdk-middleware-http": patch
+---
+
+Allow requests with empty content-type
--- a/.changeset/orange-years-rush.md
+++ b/.changeset/orange-years-rush.md
@@ -1,5 +1,0 @@
----
-"@commercetools/sdk-middleware-http": patch
----
-
-Allow requests with empty content-type
--- a/.changeset/plenty-wasps-refuse.md
+++ b/.changeset/plenty-wasps-refuse.md
@@ -1,0 +1,6 @@
+---
+"@commercetools/sdk-middleware-http": patch
+---
+
+- Allow requests with empty content-type
+- To unset a request header content-type, explicitly pass `null` as the value. 
--- a/.changeset/plenty-wasps-refuse.md
+++ b/.changeset/plenty-wasps-refuse.md
@@ -1,6 +1,7 @@
 ---
-"@commercetools/sdk-middleware-http": patch
+'@commercetools/sdk-middleware-http': patch
 ---
 
-- Allow requests with empty content-type
-- To unset a request header content-type, explicitly pass `null` as the value. 
+Allow to unset the `content-type` HTTP header by explicitly passing `null` as the value.
+
+A use case for that is using `FormData` as the request body, for example to perform a file upload. The browser generally sets the `content-type` HTTP header automatically.
--- a/integration-tests/cli/discount-codes.it.js
+++ b/integration-tests/cli/discount-codes.it.js
@@ -51,6 +51,8 @@ describe('DiscountCode tests', () => {
     ])
 
     cartDiscount = data[0].body
+
+    console.log("""""""""""""""""""""""""""""""""""" + cartDiscount + """"""""""""""""""""""""""""""""""")
   }, 15000)
 
   afterAll(async () => {
--- a/integration-tests/cli/discount-codes.it.js
+++ b/integration-tests/cli/discount-codes.it.js
@@ -52,7 +52,6 @@ describe('DiscountCode tests', () => {
 
     cartDiscount = data[0].body
 
-    console.log("""""""""""""""""""""""""""""""""""" + cartDiscount + """"""""""""""""""""""""""""""""""")
   }, 15000)
 
   afterAll(async () => {
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -100,13 +100,10 @@ export default function createHttpMiddleware({
   ) => {
     let abortController: any
     const url = host.replace(/\/$/, '') + request.uri
-
-    // if it's true, the client will detect the correct content-type, will be used in sending files (ex: multipart/form-data)
-    const hasNullContentType = request.headers["Content-Type"] === null
     const body =
       typeof request.body === 'string' ||
         Buffer.isBuffer(request.body) ||
-        hasNullContentType
+        (request.body instanceof FormData)
         ? request.body
         : // NOTE: `stringify` of `null` gives the String('null')
         JSON.stringify(request.body || undefined)
@@ -115,10 +112,10 @@ export default function createHttpMiddleware({
     if (!Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type')) {
       requestHeader['Content-Type'] = 'application/json'
     }
-    if(hasNullContentType) {
+    if(request.headers["Content-Type"] === null) {
       delete requestHeader['Content-Type']
     }
-    else if (body) {
+    if (body && !(body instanceof FormData)) {
       requestHeader['Content-Length'] = Buffer.byteLength(body).toString()
     }
     const fetchOptions: RequestOptions = {
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -35,9 +35,9 @@ function calcDelayDuration(
   if (backoff)
     return retryCount !== 0 // do not increase if it's the first retry
       ? Math.min(
-          Math.round((Math.random() + 1) * retryDelay * 2 ** retryCount),
-          maxDelay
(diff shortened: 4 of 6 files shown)


tdeekens:
Thanks for the changes @MicheleRezk! I can follow along the concepts however it feels diffs contain a lot of changes due to intention. Can you maybe perform a self review and highlight and explain important changes? That would ease and speed up the review. Thanks.

MicheleRezk:
sorry @tdeekens, I adjusted the file format and spaces, it should be easier now for review, please let me know if you have any other comment, thanks a lot

emmenko:
Quick feedback: I wouldn't use a custom header to disable another header.

I would implement it so that you pass `null` as the `Content-Type` header and the middleware then unsets the header when `null` is passed.

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,17 +100,25 @@ export default function createHttpMiddleware({
...
     let abortController: any
     const url = host.replace(/\/$/, '') + request.uri
+
+    // if it's true, the client will detect the correct content-type, will be used in sending files (ex: multipart/form-data)
+    const hasNullContentType = request.headers["Content-Type"] === null
     const body =
-      typeof request.body === 'string' || Buffer.isBuffer(request.body)
+      typeof request.body === 'string' ||
+        Buffer.isBuffer(request.body) ||
+        hasNullContentType
```
emmenko, line 109:
```suggestion
        request.body instanceof FormData
```

tdeekens (approved):
From what I can tell and knowledge of the goal and topic this implementation should do the job.

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,17 +100,25 @@ export default function createHttpMiddleware({
...
+        JSON.stringify(request.body || undefined)
 
     const requestHeader: HttpHeaders = { ...request.headers }
     if (!Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type')) {
       requestHeader['Content-Type'] = 'application/json'
     }
-    if (body) {
+    if(hasNullContentType) {
+      delete requestHeader['Content-Type']
+    }
```
emmenko, line 120:
```suggestion
    if (requestHeader['Content-Type'] === null) {
      delete requestHeader['Content-Type']
    }
```

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,17 +100,25 @@ export default function createHttpMiddleware({
...
+    else if (body) {
```
emmenko, line 121:
```suggestion
    if (body && !(body instanceof FormData)) {
```

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,17 +100,25 @@ export default function createHttpMiddleware({
   ) => {
     let abortController: any
     const url = host.replace(/\/$/, '') + request.uri
+
+    // if it's true, the client will detect the correct content-type, will be used in sending files (ex: multipart/form-data)
+    const hasNullContentType = request.headers["Content-Type"] === null
```
emmenko, line 105:
```suggestion
```

emmenko (changes requested):
The `hasNullContentType` logic was confusing me a bit. I tried to make it a bit more explicit (suggestions).

packages/sdk-middleware-http/src/http.js
MicheleRezk: Thanks @emmenko for feedback, I didn't want to restrict it to FormData as we may need to refactor this again if we introduced a new service with application/x-www-form-urlencoded content type

packages/sdk-middleware-http/src/http.js
```diff
@@ -101,16 +101,21 @@ export default function createHttpMiddleware({
     let abortController: any
     const url = host.replace(/\/$/, '') + request.uri
     const body =
-      typeof request.body === 'string' || Buffer.isBuffer(request.body)
+      typeof request.body === 'string' ||
+        Buffer.isBuffer(request.body) ||
+        (request.body instanceof FormData)
```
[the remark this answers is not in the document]
emmenko: Hmm yeah that sucks...maybe we can refactor this a bit and use the `content-type` info to run the check.

For instance, only if `application/json` we need to `JSON.stringify`.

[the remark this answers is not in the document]
emmenko:
```js
if (!Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type')) {
  requestHeader['Content-Type'] = 'application/json'
}
if(request.headers["Content-Type"] === null) {
  delete requestHeader['Content-Type']
}

const body = requestHeader['Content-Type'] === 'application/json'
  // NOTE: `stringify` of `null` gives the String('null')
  ? JSON.stringify(request.body || undefined)
  : request.body

if (typeof body === 'string' || Buffer.isBuffer(request.body)) {
  requestHeader['Content-Length'] = Buffer.byteLength(body).toString()
}
```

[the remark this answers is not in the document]
MicheleRezk: yes I think this is much better, thanks for the feedback :+1:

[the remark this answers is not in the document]
tdeekens: Yes, I am not 100% into this topic but wouldn't using the `content-type` header and `multipart/form-data` work?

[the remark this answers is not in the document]
MicheleRezk: @tdeekens if we passed the content-type explicitly as 'multipart/form-data' in the request header it wouldn't work as there is other things need to be calculated like boundary, so to bypass this we can pass content-type as null and let the client Identify the correct content-type and the boundary to make it work

emmenko (approved):
Thanks!

emmenko:
@MicheleRezk btw, you need to write a changeset so that the changes here can be released.

ajimae (dismissed):
LGTM!

.changeset/orange-years-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools/sdk-middleware-http": patch
+---
+
+Allow requests with empty content-type
```
emmenko, line 5: Please specify that you need to explicitly pass `null` to "unset" the header.

ajimae: Hi @emmenko 
This is how I worded it: To unset a request header's content-type, explicitly pass `null` as the value. 

Hope it's fine?

tdeekens (dismissed):

.changeset/plenty-wasps-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools/sdk-middleware-http": patch
+---
+
+- Allow requests with empty content-type
```
emmenko, line 5: Remove this line. Also the other changeset file

integration-tests/cli/discount-codes.it.js
```diff
@@ -51,6 +51,8 @@ describe('DiscountCode tests', () => {
     ])
 
     cartDiscount = data[0].body
+
+    console.log("""""""""""""""""""""""""""""""""""" + cartDiscount + """"""""""""""""""""""""""""""""""")
```
emmenko, line 55: Remove

packages/sdk-middleware-http/src/http.js
```diff
@@ -100,17 +100,19 @@ export default function createHttpMiddleware({
...
       requestHeader['Content-Type'] = 'application/json'
     }
-    if (body) {
+    if(request.headers["Content-Type"] === null) {
+      delete requestHeader['Content-Type']
+    }
+    const body = requestHeader['Content-Type'] === 'application/json'
+        // NOTE: `stringify` of `null` gives the String('null')
+        ? JSON.stringify(request.body || undefined)
+        : request.body
```
emmenko, line 113: The failing integration tests might be related to this change. If you want to debug the error, print out the `console.log(body)` to see what value it is.

emmenko (changes requested):

emmenko:
I think the changes here are good now. The failing integration tests are the same as in `master`, so unrelated to these changes.

emmenko (approved):

ajimae (approved):

emmenko:
@tdeekens anything else from your side?

Issue #1743, opened 2021-12-17
Retry on 5xx HTTP error


lojzatran:
nodejs sdk is currently retrying on 503 HTTP error only (https://github.com/commercetools/nodejs/blob/1167644b37db9af7baf511b8b14cdf6abf48f47c/packages/sdk-middleware-http/src/http.js#L215). According to the CTP documentation it should retry at least on 502 HTTP error (https://docs.commercetools.com/api/errors#502-bad-gateway). You may consider retrying also on other 5xx HTTP errors where it makes sense.

lojzatran:
Also is it possible to add own retry rules? E.g. we want to retry in some specific 4xx errors that we have confirmed from support that they should be retried. In the old sphere-node-sdk we have retryKeywords and we could also configure them: https://github.com/commercetools/sphere-node-sdk/blob/master/src/coffee/repeater-task-queue.coffee#L7. How could this be done in the current nodejs sdk?

ajimae:
Hi @lojzatran 

I will look into this and if it's something we need to implement I will raise a PR for it.

Thanks

butenkor:
The question is how reliable is to have error codes only since some of the errors which are listed in the the old keywords SDK are not related to the API response code (since commercetools server is not reached) but also OS limitation like amount of possible concurrent connection and similar - in such cases it would allow the user to retry automatically too.

PR #1764, opened 2022-02-10
Add retryCodes option in HttpMiddlewareOptions

ajimae:
#### Summary

- add retryCodes option to HttpMiddlewareOptions
- add unit test
- add documentation
- add flow types
- remove deprecated options from `.flowconfig`

#### Description

This adds options to allow customers retry requests on a given status (error) codes.

- Tests
  - [x] Unit
- [x] Documentation

#### Related Issue:

https://github.com/commercetools/nodejs/issues/1743


lojzatran:
Hi @ajimae I don't think this retry is enough. As I wrote in the issue (https://github.com/commercetools/nodejs/issues/1743#issuecomment-1028794106), it should be able to retry also at network errors OR at 4xx errors if it contains a particular error message. Also we should be able to define the retry rules. Basically it should work the same as in the old sphere-node-sdk (https://github.com/commercetools/sphere-node-sdk/blob/master/src/coffee/repeater-task-queue.coffee#L7).

docs/sdk/api/sdkMiddlewareHttp.md
```diff
@@ -67,6 +68,7 @@ const client = createClient({
       includeOriginalRequest: true,
       maskSensitiveHeaderData: true,
       enableRetry: true,
+      retryCodes: [504, 'ETIMEDOUT', 'ECONNREFUSED', 503],
```
jenschude, line 71: I would have made the retryCodes a property of the retryConfig. And use a default value array with the 503 status code like:

```
retryConfig: {
   retryCodes: [503],
  ...
}
```

packages/sdk-middleware-http/src/http.js
```diff
@@ -59,6 +59,7 @@ export default function createHttpMiddleware({
   maskSensitiveHeaderData = true,
   enableRetry,
   timeout,
+  retryCodes = [],
```
jenschude, line 62:
```
retryConfig: {
   retryCodes: [503],
  ...
}
```

lojzatran:
Hi @ajimae do you have an estimation when this PR will be merged and released? Thank you.

ajimae:
> Hi @ajimae do you have an estimation when this PR will be merged and released? Thank you.

I really can't say because we need @jenschude to approve.

jenschude (approved):

PR #1767, opened 2022-02-23
feat(sync-actions): add product selection sync actions

mohib0306:
#### Summary
- Product Selections are introduced to the platform since January. We also have implemented a view in the MC to enable users to create product selection and also update them. Since the product selection were not public, the sync action for updating the product selection were generated locally in the MC code base and used for the update mutations of the GraphQL API. This PR adds product selections related sync actions to the `sync-actions` package.


danrleyt (approved):
Thanks for the contribution @mohib0306! LGTM!

PR #1784, opened 2022-05-22
fix(resource-deleter): add regeneratorRuntime dependency to remove error

VineetKumarKushwaha:
#### Fixing Error "regeneratorRuntime is not defined" in resource-deleter

Recently I was investigating issue in `resource-deleter` and found that after bumping babel, resource-deletor throws an error because we use `async & await` and regeneratorRuntime import was missing in the package.

- Added `core-js`
- Added `regenerator-runtime`

More info can be found here https://stackoverflow.com/questions/53558916/babel-7-referenceerror-regeneratorruntime-is-not-defined

### DIFF
--- a/packages/resource-deleter/package.json
+++ b/packages/resource-deleter/package.json
@@ -43,6 +43,8 @@
     "pino": "^6.0.0",
     "pretty-error": "^2.1.1",
     "prompts": "^2.0.4",
-    "yargs": "^16.0.0"
+    "yargs": "^16.0.0",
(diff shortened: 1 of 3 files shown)


ajimae:
Thanks @VineetKumarKushwaha for this fix.

I saw the tests were failing due to expired project credentials. I had to update the credentials and re-ran the test and everything now seem okay.

I will go ahead and approve.

ajimae (approved):
LGTM!

PR #1796, opened 2022-07-31
fix(sync-actions): fix enum type actions

VineetKumarKushwaha:
#### Fix (sync-actions): addEnumValue should be generated when there type is ``

Fixes: https://jira.commercetools.com/browse/SUPPORT-16999

When using the sync-actions library to sync custom types wrong update actions are generated when enum values are updated.
we check for the attribute type with a lower ‘e’ instead of an capital 'E’ as required in the CustomFieldEnumType.
Also, we don't pass fieldName when generating the action so it takes type as a fieldName which is incorrect.

### DIFF
--- a/.changeset/six-melons-hope.md
+++ b/.changeset/six-melons-hope.md
@@ -1,0 +1,5 @@
+---
+'@commercetools/sync-actions': minor
+---
+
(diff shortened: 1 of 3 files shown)


ajimae (approved):
LGTM!

PR #1807, opened 2022-09-07
Fix request to allow application/graphql content-type 

ajimae:
#### Summary
- [x] include application/graphql as an acceptable header content-type

#### Related Issues
- [x] [#316](https://github.com/commercetools/commercetools-sdk-typescript/issues/316)

### DIFF
--- a/packages/sdk-middleware-http/src/http.js
+++ b/packages/sdk-middleware-http/src/http.js
@@ -114,8 +114,10 @@ export default function createHttpMiddleware({
 
       // If no content-type is provided, defaults to application/json
       if (
-        !Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type') ||
-        !Object.prototype.hasOwnProperty.call(requestHeader, 'content-type')
+        !(
+          Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type') ||
+          Object.prototype.hasOwnProperty.call(requestHeader, 'content-type')
+        )
       ) {
         requestHeader['Content-Type'] = 'application/json'
       }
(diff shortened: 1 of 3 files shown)


packages/sdk-middleware-http/src/http.js
```diff
@@ -107,24 +107,25 @@ export default function createHttpMiddleware({
...
+        delete requestHeader['Content-Type']
+      }
+
       // If no content-type is provided, defaults to application/json
       if (
-        !Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type')
+        !(
+          Object.prototype.hasOwnProperty.call(requestHeader, 'Content-Type') ||
+          Object.prototype.hasOwnProperty.call(requestHeader, 'content-type')
+        )
```
emmenko, line 120: This looks like a regression. The `requestHeader['Content-Type'] === null` must run AFTER the default.

Any reason to change that? @ajimae

ajimae: [This PR](https://github.com/commercetools/nodejs/pull/1751/files#diff-06163118d2bcb7381ae16147c21bd62bf5b2f2c5aff4c1394301f63f03527140) is where the change was first introduced. I could remember there was issue with file upload when using this package by the `#beam-me-up` team, where the `Content-Type` was sent as null so @MicheleRezk had to unset this header first.

yassinejebli: @ajimae I'm not sure if this change has something to do with unsetting the `Content-Type` header or not but setting `null` to the header seems not working as expected, please check this [thread](https://commercetools.slack.com/archives/CCL83RA06/p1676300321733559)

ajimae: Oh, I think I get it now. I will push a PR to correct this.

PR #1816, opened 2022-11-09
refactor(utils): improve performance of `copyEmptyArrayProps` function

qmateub:
#### Summary

Refactors a little bit the way `copyEmptyArrayProps` is implemented in order to seek a better performance specially for cases with a huge amount of items in an array.

#### Description

Context: We (pangolins) do a heavy use of this library in Audit Log. We were observing some customer comparisons that were taking a huge amount of time and when debugging that led us to spotting sync actions library as the one that was causing the delay.

After some investigation, we found that the library was not really performant for cases when customers with a huge amount of addresses where compared between versions. In our case it was a customer with almost 13k addresses in the old and new version.

In this PR we propose a solution by changing the implementation to remove the usage of spread operators inside a reduce in favour of attribute assignments or even the usage of a map instead of array methods which usually led to the O(n^2) problem [More info](https://prateeksurana.me/blog/why-using-object-spread-with-reduce-bad-idea/#why-is-the-spread-operator-so-slow)

#### Todo

- Tests
  - [X] Unit
  - [X] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -13,18 +13,17 @@ const CUSTOM = 'custom'
 export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
   if (!isNil(oldObj) && !isNil(newObj)) {
     const nextObjectWithEmptyArray = Object.entries(oldObj).reduce(
-      (nextObject, [key, value]) => {
-        const merged = {
-          ...newObj,
-          ...nextObject,
-        }
-
+      (merged, [key, value]) => {
         // Ignore CUSTOM key as this object is dynamic and its up to the user to dynamically change it
         // todo, it would be better if we pass it as ignored keys param
         if (key === CUSTOM) return merged
 
         if (Array.isArray(value) && newObj[key] && newObj[key].length >= 1) {
           /* eslint-disable no-plusplus */
+          const hashMapValue = value.reduce((acc, val) => {
+            acc[val.id] = val
+            return acc
+          }, {})
           for (let i = 0; i < newObj[key].length; i++) {
             if (
               !isNil(newObj[key][i]) &&
@@ -32,16 +31,14 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
               !isNil(newObj[key][i].id)
             ) {
               // Since its unordered array elements then check if the element on `oldObj` exists by id
-              const foundObject = value.find(
-                (v) => Number(v.id) === Number(newObj[key][i].id)
-              )
+              const foundObject = hashMapValue[newObj[key][i].id]
               if (!isNil(foundObject)) {
                 const [, nestedObject] = copyEmptyArrayProps(
                   foundObject,
                   newObj[key][i]
                 )
                 /* eslint-disable no-param-reassign */
-                newObj[key][i] = nestedObject
+                merged[key][i] = nestedObject
               }
             }
           }
@@ -49,10 +46,8 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
           return merged
         }
         if (Array.isArray(value)) {
-          return {
-            ...merged,
-            [key]: isNil(newObj[key]) ? [] : newObj[key],
-          }
+          merged[key] = isNil(newObj[key]) ? [] : newObj[key]
+          return merged
         }
         if (
           !isNil(newObj[key]) &&
@@ -62,10 +57,8 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
           !(value instanceof Date)
         ) {
           const [, nestedObject] = copyEmptyArrayProps(value, newObj[key])
-          return {
-            ...merged,
-            [key]: nestedObject,
-          }
+          merged[key] = nestedObject
+          return merged
         }
         return merged
       },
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -13,46 +13,41 @@ const CUSTOM = 'custom'
 export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
   if (!isNil(oldObj) && !isNil(newObj)) {
     const nextObjectWithEmptyArray = Object.entries(oldObj).reduce(
-      (nextObject, [key, value]) => {
-        const merged = {
-          ...newObj,
-          ...nextObject,
-        }
```
qmateub, line 20: This combined object was not needed as it is exactly what the accumulator does in the `reduce`.

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -13,46 +13,41 @@ const CUSTOM = 'custom'
...
         // Ignore CUSTOM key as this object is dynamic and its up to the user to dynamically change it
         // todo, it would be better if we pass it as ignored keys param
         if (key === CUSTOM) return merged
 
         if (Array.isArray(value) && newObj[key] && newObj[key].length >= 1) {
           /* eslint-disable no-plusplus */
+          const hashMapValue = value.reduce((acc, val) => {
+            acc[val.id] = val
+            return acc
+          }, {})
```
qmateub, line 26: This is a small trick to avoid the usage of `find` in a situation of array of array (of arrays). The difference between both approaches was remarkable when debugging this locally.

With hashmap it was less than 100 ms
With `.find` around 6 seconds

packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -13,46 +13,41 @@ const CUSTOM = 'custom'
...
-              )
+              const foundObject = hashMapValue[newObj[key][i].id]
               if (!isNil(foundObject)) {
                 const [, nestedObject] = copyEmptyArrayProps(
                   foundObject,
                   newObj[key][i]
                 )
                 /* eslint-disable no-param-reassign */
-                newObj[key][i] = nestedObject
+                merged[key][i] = nestedObject
```
qmateub, line 41: That means that we can just reassign this to the accumulator.

wrsenn (approved):

packages/sync-actions/test/utils/copy-empty-array-props.spec.js
```diff
@@ -306,3 +307,25 @@ test('shouldnt change objects since there is no arrays to copy', () => {
...
+  }
+
+  const start = performance.now()
+  const [old, fixedNewObj] = copyEmptyArrayProps(oldObj, newObj)
+  const end = performance.now()
+
+  expect(old).toEqual(oldObj)
+  expect(fixedNewObj).toEqual(newObj)
+  expect(end - start).toBeLessThan(100)
+})
```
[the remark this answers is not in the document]
taylor-knapp: (or checkout 97f3638ab7ce4ab428e9e6a7c744ecb5f5a56017 and run it there)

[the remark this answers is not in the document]
taylor-knapp: (or checkout 97f3638ab7ce4ab428e9e6a7c744ecb5f5a56017 and run it there)

taylor-knapp (approved):

ajimae (approved):

packages/sync-actions/src/utils/copy-empty-array-props.js
angel-marin-ct: As a pattern I will name this trick a "dictionary"... you could even do this a little trickier

```
 const valueDict = Object.assign({}, ...value.map( x=>( { [x.id]: x) ) )
 ```

Comment (taylor-knapp) [reply packages/sync-actions/src/utils/copy-empty-array-props.js:26]: fwiw the object assign method is a bit slower, probably because you are mapping through all of the values three times (once for the map, once for the spread, once for the assignment).

If it's a large array (100k elements) the reduce approach takes ~2ms whereas the object assign approach takes ~50ms.

```
// Create an array for testing.
const testArray = new Array(100_000).fill(null).map((_, index) => ({ id: index }));

// Store individual iteration times.
const reduceTimes = [];
const objectAssignTimes = [];

const testIterations = 1_000;
for (let i = 0; i < testIterations; i++) {
  // Run reduce method.
  const startTimeReduce = Date.now();
  testArray.reduce((acc, element) => {
    acc[element.id] = element;
    return acc;
  }, {});
  reduceTimes.push(Date.now() - startTimeReduce);

  // Run object assign method.
  const startTimeObjectAssign = Date.now();
  Object.assign({}, ...testArray.map((element) => ({ [element.id]: element })));
  objectAssignTimes.push(Date.now() - startTimeObjectAssign);
}

// Log results.
console.log({ totalTimeReduce: reduceTimes.reduce((acc, val) => acc + val) / testIterations });
console.log({
  totalTimeObjectAssign: objectAssignTimes.reduce((acc, val) => acc + val) / testIterations,
});
```

Comment (angel-marin-ct) [reply packages/sync-actions/src/utils/copy-empty-array-props.js:26]: Well... I just said trickier :-)

Thanks for the testing!

Another interesting approach will be if Array.prototype.group proposal gets approved. It won't return exactly a dictionary because the values will be an array...

But it will read nicely:

```
array.group( x=>x.id ) 
```

qmateub: Thanks both for the input and the testing around this. It was really a nice learning experience and I think this improves the library for those cases that we see big numbers.

Just for the record, the case that we were checking was a customer from cimpress that had almost 13k addresses defined in it and the size of the entity was almost 9MB 😨 💥

danrleyt (approved):
Thanks for the brilliant contribution, that will also help us immensely.

Issue #1653, opened 2021-01-26
Optimize client-side @commercetools/sdk-auth


studentIvan:
### Description

A lot of useless code coming with this library:
1. Buffer polyfill
2. Lodash (defaultsDeep)

### Expected Behavior

It works without these libraries or load them only on demand.

### Current Behavior

It works with these libraries and is impossible even to replace them if you use just simple `import SdkAuth, { TokenProvider } from '@commercetools/sdk-auth';` because it uses the UMD version.

### Possible Solution

https://github.com/commercetools/nodejs/pull/1649
https://github.com/commercetools/nodejs/pull/1650
https://github.com/commercetools/nodejs/pull/1651

tayzlor:
+ 1 This prevents this library from being used with build tools such as Vite - which inherently don't support node polyfills in the browser

PR #1830, opened 2023-01-13
Remove redundant dependencies

ajimae:
#### Summary
Remove redundant dependencies

#### Description
Remove `lodash.defaultsdeep` from dependency tree

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR 
  
  
  

Thanks to [@studentIvan](https://github.com/studentIvan) for the suggestions and recommended performance optimization [here](https://github.com/commercetools/nodejs/issues/1653)

### DIFF
--- a/packages/sdk-auth/package.json
+++ b/packages/sdk-auth/package.json
@@ -36,6 +36,7 @@
   },
   "dependencies": {
     "@commercetools/sdk-middleware-http": "^7.0.0",
+    "lodash.defaultsdeep": "^4.6.1",
     "qss": "2.0.3"
   },
   "devDependencies": {
--- a/packages/sdk-auth/package.json
+++ b/packages/sdk-auth/package.json
@@ -36,7 +36,6 @@
   },
   "dependencies": {
     "@commercetools/sdk-middleware-http": "^7.0.0",
-    "lodash.defaultsdeep": "^4.6.1",
     "qss": "2.0.3"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


packages/sdk-auth/src/auth.js
```diff
@@ -237,10 +236,21 @@ export default class SdkAuth {
...
-    const mergedConfig = defaultsDeep({}, config, this.config)
 
-    // handle scopes array - defaultsDeep would merge arrays together
+    // It is important to note the difference between plain old javascript Object.assign()
+    // and the lodash _.defaultsDeep() functions especially how they are both used here
+
+    // _.defaultsDeep({ 'a': { 'b': 2 } }, { 'a': { 'b': 1, 'c': 3 } }) // { a: { b: 2, c: 3 } }
+    // Object.assign({ 'a': { 'b': 2 } }, { 'a': { 'b': 1, 'c': 3 } }) // { a: { b: 1, c: 3 } }
+
+    // handle scopes array - Object.assign would merge arrays together
```
jenschude, line 246: I think this is actually not really good if a missing dependency would change the behavior here. I would take a look what's the size of the config and do the merge straightforward in a function without the need of any dependency. Or directly use Object.assign

packages/sdk-auth/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@commercetools/sdk-middleware-http": "^7.0.0",
-    "lodash.defaultsdeep": "^4.6.0",
+    "lodash.defaultsdeep": "^4.6.1",
```
jenschude, line 39: is this dependency still needed after the change to Object.assign?

ajimae: Just remove it 🙂

jenschude (approved):

Issue #1832, opened 2023-01-31
[DiscountCodeImporter] Show meaningful error


VineetKumarKushwaha:
### DiscountCodeImporter shows generic exception 
- ATM, if there is problem with input to discount-code-importer it throws an generic exception and doesn't tell us what is wrong and why it is failed. 
- This package is being used by the express-impex and hence the customers are not able to understand about the error and reach out to customer support to think its an bug.

Generic Error
- ![image](https://user-images.githubusercontent.com/8777104/215694970-47d73ad1-d023-4f19-b2af-9345359f66e2.png)
Currently user sees this exception incase of any error however specialy it should also show actual error like
`error:   Cart Discount '0bfd0919-0540-4350-a6af-16ffa913a954' doesn't require a discount code.`

PR #1833, opened 2023-02-01
fix(discount-code-importer): improve error logging

VineetKumarKushwaha:
#### Improve Error Logging for Discount Importer in Impex

- Pass error message to the consumer. Currently in impex we only see the generic coming from the Discount Importer. So with this PR we are also pushing specific error message in the array which we show in the UI

Fix: #1832


marcelogpinheiro (approved):
Nice! Thank you @VineetKumarKushwaha 💯

markus-azer (approved):
💯

PR #1836, opened 2023-02-07
Add set authentication mode sync action

nicolasnieto92:
#### Summary

Generate `setAuthenticationMode` sync action.

https://docs.commercetools.com/api/history/change-types#setauthenticationmodechange

#### Description

Unit tests and changeset has been added.

### DIFF
--- a/packages/sync-actions/src/customer-actions.js
+++ b/packages/sync-actions/src/customer-actions.js
@@ -162,6 +162,7 @@ function buildAuthenticationModeActions({ actions, diff, oldObj, newObj }) {
       const now = newObj[key]
       const isNotDefinedBefore = isEmptyValue(oldObj[key])
       const isNotDefinedNow = isEmptyValue(newObj[key])
+      const authenticationModes = ['Password', 'ExternalAuth']
 
       if (!delta) return undefined
 
@@ -171,6 +172,9 @@ function buildAuthenticationModeActions({ actions, diff, oldObj, newObj }) {
         throw new Error(
           'Cannot set to Password authentication mode without password'
         )
+
+      if (!authenticationModes.includes(newObj.authenticationMode))
+        throw new Error('Invalid Authentication Mode')
 
       if (!isNotDefinedNow && isNotDefinedBefore) {
         // no value previously set
--- a/packages/sync-actions/src/customer-actions.js
+++ b/packages/sync-actions/src/customer-actions.js
@@ -1,3 +1,4 @@
+import isNil from 'lodash.isnil'
 import {
   buildBaseAttributesActions,
   buildReferenceActions,
@@ -7,6 +8,14 @@ import createBuildArrayActions, {
   REMOVE_ACTIONS,
   CHANGE_ACTIONS,
 } from './utils/create-build-array-actions'
+import * as diffpatcher from './utils/diffpatcher'
+import clone from './utils/clone'
+
+const normalizeValue = (value) =>
+  typeof value === 'string' ? value.trim() : value
+
+export const createIsEmptyValue = (emptyValues) => (value) =>
+  emptyValues.some((emptyValue) => emptyValue === normalizeValue(value))
 
 export const baseActionsList = [
   { action: 'setSalutation', key: 'salutation' },
@@ -26,7 +35,6 @@ export const baseActionsList = [
     key: 'stores',
   },
   { action: 'setKey', key: 'key' },
-  { action: 'setAuthenticationMode', key: 'authenticationMode' },
 ]
 
 export const setDefaultBaseActionsList = [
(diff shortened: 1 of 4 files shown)


marcelogpinheiro (approved):
Nice work, @nicolasnieto92 ! 💯

packages/sync-actions/test/customer-sync.spec.js
```diff
@@ -533,4 +534,38 @@ describe('Actions', () => {
...
+    now = {
+      authenticationMode: 'Password',
+    }
+
+    actual = customerSync.buildActions(now, before)
+    expected = [
+      {
+        action: 'setAuthenticationMode',
+        authenticationMode: now.authenticationMode,
+      },
```
markus-azer, line 567: shouldnt the action also include `password` ref https://docs.commercetools.com/api/projects/customers#set-authenticationmode, wdyt?

markus-azer (changes requested):

packages/sync-actions/test/customer-sync.spec.js
nicolasnieto92: totally right. otherwise i got:

`{
    "statusCode": 400,
    "message": "'password' should not be empty.",
    "errors": [
        {
            "code": "InvalidOperation",
            "message": "'password' should not be empty.",
            "action": {
                "action": "setAuthenticationMode",
                "authMode": "Password"
            },
            "actionIndex": 1
        }
    ]
}`

thanks mate.

packages/sync-actions/src/customer-actions.js
```diff
@@ -126,3 +141,61 @@ export function actionsMapShippingAddresses(diff, oldObj, newObj) {
 
   return handler(diff, oldObj, newObj)
 }
+
+export function actionsMapAuthenticationModes(diff, oldObj, newObj) {
+  // eslint-disable-next-line no-use-before-define
```
nicolasnieto92, line 146: i though it was better to keep `buildAuthenticationModeActions` at the bottom and all the actionMaps together.

packages/sync-actions/src/customer-actions.js
```diff
@@ -126,3 +141,61 @@ export function actionsMapShippingAddresses(diff, oldObj, newObj) {
...
+  return buildAuthenticationModeActions({
+    actions: authenticationModeActionsList,
+    diff,
+    oldObj,
+    newObj,
+  })
+}
+
+function buildAuthenticationModeActions({ actions, diff, oldObj, newObj }) {
```
nicolasnieto92, line 155: had to build this modified version of `buildBaseAttributesActions` cause is a special case and this logic is only used for authentication modes.

packages/sync-actions/test/customer-sync.spec.js
```diff
@@ -533,4 +539,96 @@ describe('Actions', () => {
...
+
+    expected = []
+    expect(actual).toEqual(expected)
+
+    before = {
+      authenticationMode: 'ExternalAuth',
+    }
+    now = {
+      authenticationMode: '',
+    }
```
markus-azer, line 614: little bit doubtful here, cause in this case it should be unset, but seems there is no action for this.

packages/sync-actions/src/customer-actions.js
```diff
@@ -126,3 +141,61 @@ export function actionsMapShippingAddresses(diff, oldObj, newObj) {
...
+      if (newObj.authenticationMode === 'Password' && !newObj.password)
+        throw new Error(
+          'Cannot set to Password authentication mode without password'
+        )
+
+      if (!isNotDefinedNow && isNotDefinedBefore) {
+        // no value previously set
+        if (newObj.authenticationMode === 'ExternalAuth')
+          return { action: item.action, authMode: now }
+        return { action: item.action, authMode: now, [value]: newObj.password }
```
markus-azer, line 179: Codecov created this assumtion whats if the user input isnt trusted, ex: if user provided  authMode as 'xyz'
this will be invalid action created, just random thought, maybe 
```
if (newObj.authenticationMode === 'Password')
  return { action: item.action, authMode: 'Password', [value]: newObj.password }
```
  wdyt?

markus-azer (approved):
Looks amazing and love the testing 💪 
just left couple of comments

packages/sync-actions/test/customer-sync.spec.js
nicolasnieto92: me too, thats why i didnt generate any action for this, to avoid any unwanted behaviour. but im also not sure

packages/sync-actions/src/customer-actions.js
nicolasnieto92: as always, good catch, thanks! i added a case for this and updated a previous one. so invalid strings (e.g. 'xyz')  and empty strings will throw an error. and if key is not present, it will not create an action. wdyt?

PR #1839, opened 2023-02-24
fix: failing to sync froozen arrays for prices

tdeekens:
#### Summary

At times consumers may pass in frozen arrays. Even though this is discouraged it worked before. Syncing frozen arrays doesn't work as we copy properties from one to another. Hence we need to slice a frozen array.

CleanShot 2023-02-24 at 13 43 47@2x
CleanShot 2023-02-24 at 13 44 30@2x

### DIFF
--- a/packages/sync-actions/src/utils/copy-empty-array-props.js
+++ b/packages/sync-actions/src/utils/copy-empty-array-props.js
@@ -37,6 +37,10 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
                   foundObject,
                   newObj[key][i]
                 )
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/utils/copy-empty-array-props.js
```diff
@@ -37,6 +37,10 @@ export default function copyEmptyArrayProps(oldObj = {}, newObj = {}) {
                   foundObject,
                   newObj[key][i]
                 )
+                if (Object.isFrozen(merged[key])) {
```
qmateub, line 40: ❄️

qmateub (approved):
🔥 🔥 🔥

Issue #1766, opened 2022-02-18
[sync-actions] Type FieldDefinition change from required false to true failed


lojzatran:
### Description

When I have a fieldDefintion changed from required=false to required=true
```
{
  "name": {
    "en": "Subscription order"
  },
  "description": {
    "en": "Subscription order definition"
  },
  "resourceTypeIds": ["order"],
  "fieldDefinitions": [
    {
      "name": "deliveryDate",
      "type": {
        "name": "DateTime"
      },
      "required": true,   // changed here from false to true
      "label": {
        "en": "deliveryDate"
      },
      "inputHint": "SingleLine"
    }
  ]
}
```

sync-actions throws an error `"Cannot read properties of undefined (reading 'values')` and fails at this line: https://github.com/commercetools/nodejs/blob/4974bb700aa9469282f38da4e40ecbb8bba7d513/packages/sync-actions/src/types-actions.js#L131 When I inspect that code, diffValue has this value:
```
diffValue = `{
  "required": [
    false,
    true
  ]
}
```

### Expected Behavior

Since there is [no update action](https://docs.commercetools.com/api/projects/types#update-actions) for this change I would expect following:
1. It would try to remove the fieldDefinition and re-add it with the correct required value. This might not work since there could be resources created based on this type.
1. If the previous point doesn't work, it should fail with a nice error describing what is the problem and what could be done.

lojzatran:
Hi team, any changes here? Thank you.

christina-grannas:
I am getting this as well, would be great if it could be fixed

PR #1852, opened 2023-04-26
BMU-401 fix(sync-actions): fix type sync-actions errors

markus-azer:
#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/types-actions.js
+++ b/packages/sync-actions/src/types-actions.js
@@ -136,7 +136,7 @@ export function actionsMapFieldDefinitions(
           label: extractedPairs.newObj.label,
           fieldName: extractedPairs.oldObj.name,
         })
-      } else if (diffValue.type.values) {
+      } else if (diffValue?.type?.values) {
         actions.push(
           ...actionsMapEnums(
             extractedPairs.oldObj.name,
@@ -152,7 +152,7 @@ export function actionsMapFieldDefinitions(
         if (diffValue.length === 3 && diffValue[2] === 3) {
           actions.push({
             action: 'changeFieldDefinitionOrder',
-            fieldNames: next,
+            fieldNames: next.map((n) => n.name),
           })
         } else {
           const deltaValue = diffPatcher.getDeltaValue(diffValue)
--- a/packages/sync-actions/test/types-sync-fields.spec.js
+++ b/packages/sync-actions/test/types-sync-fields.spec.js
@@ -336,7 +336,14 @@ describe('Actions', () => {
       ])
     })
   })
-  describe('should ignore required field in fieldDefinition', () => {
+
+  /**
+   * there is no update action for fieldDefinition -> required,
+   * so this field is immutable and unchangeable.
+   * in case of changing it, this were throwing `Cannot read properties of undefined` cause its nested field.
+   * below test is making sure this field is ignored and without any internal package errors.
+   */
+  describe('should ignore changes in required field in fieldDefinition', () => {
     beforeEach(() => {
       before = createTestType({
         fieldDefinitions: [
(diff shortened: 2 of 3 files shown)


packages/sync-actions/src/types-actions.js
```diff
@@ -136,7 +136,7 @@ export function actionsMapFieldDefinitions(
           label: extractedPairs.newObj.label,
           fieldName: extractedPairs.oldObj.name,
         })
-      } else if (diffValue.type.values) {
+      } else if (diffValue?.type?.values) {
```
markus-azer, line 139: That one fixes this issue https://github.com/commercetools/nodejs/issues/1766

packages/sync-actions/src/types-actions.js
```diff
@@ -152,7 +152,7 @@ export function actionsMapFieldDefinitions(
         if (diffValue.length === 3 && diffValue[2] === 3) {
           actions.push({
             action: 'changeFieldDefinitionOrder',
-            fieldNames: next,
+            fieldNames: next.map((n) => n.name),
```
markus-azer, line 155: doc: https://docs.commercetools.com/api/projects/types#change-the-order-of-fielddefinitions

packages/sync-actions/src/types-actions.js
nicolasnieto92: good catch! although this is not throwing error anymore, i think the `required` field is still not being updated.
so [this](https://github.com/commercetools/commercetools-importer/pull/2552/files#diff-b9adf055e61d9e3c62b4ffb433f171742bd5f8b04d226bc3919c0c257d476ac5R75-R97) particular test is not failing anymore, but if we check the `required` field, its still `false` if im not mistaken.

packages/sync-actions/test/types-sync-fields.spec.js
```diff
@@ -336,4 +336,28 @@ describe('Actions', () => {
...
+            name: 'first',
+            required: true,
+          },
+        ],
+      })
+      now = createTestType({
+        fieldDefinitions: [
+          {
+            name: 'first',
+            required: false,
```
tdeekens, line 353: Sorry, I am not following: we change the value but don't create an update action. Can you elaborate?

packages/sync-actions/src/types-actions.js
markus-azer: Yep very true, this field is unchangeable, so the current behaviour is just ignore these fields

packages/sync-actions/test/types-sync-fields.spec.js
markus-azer: totally agree the test is already confusing 👍 (gonna rename it better and add more context to test), there is some fields that is immutable or unchangeable so there is no `update actions` for it, which apply for the this field `FieldDefinition->required` (maybe also that should be mentioned in the public doc cause there is no clue cause it also apply for `resourceTypeIds`, `FieldDefinition->Name`).

and the current behaviour is to just ignore it for example in product entity , `[productType](https://docs.commercetools.com/api/projects/products#product)`  field `Cannot be changed.`

for above field `required`, changing it causing throwing errors 😞, and thought of adding test for it.

already been thinking about adding util for that case and pass it as config option, so it doesnt require major upgrade.

does it make sense?

tdeekens: Yeah thanks makes sense.

packages/sync-actions/src/types-actions.js
nicolasnieto92: oh i see! good then 👍

nicolasnieto92 (approved):
looks good! thanks 🚀

tdeekens (approved):

PR #1861, opened 2023-06-25
release(product-json-to-xlsx): minor version bump

VineetKumarKushwaha:

### DIFF
--- a/.changeset/little-cars-obey.md
+++ b/.changeset/little-cars-obey.md
@@ -1,0 +1,5 @@
+---
+'@commercetools/product-json-to-xlsx': minor
+---
+
+Upgrage exceljs


.changeset/little-cars-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools/product-json-to-xlsx': minor
+---
+
+Upgrage exceljs
```
markus-azer, line 5: hmm, is this PR only about changeset or did the upgrade happened in older PR?

VineetKumarKushwaha: Yes upgrade didn't happen because I missed to commit this file so with this commit it will be released

PR #1864, opened 2023-06-26
feat: add missing sync action for StandalonePrice entity `setPriceTiers`

ARRIOLALEO:
#### Summary
This PR adds support for StandalonePrice `setPriceTiers`,

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/prices-actions.js
+++ b/packages/sync-actions/src/prices-actions.js
@@ -1,12 +1,10 @@
 /* eslint-disable max-len */
-import {
-  buildBaseAttributesActions,
-} from './utils/common-actions'
-
+import { buildBaseAttributesActions } from './utils/common-actions'
 
 export const baseActionsList = [
   { action: 'changeValue', key: 'value' },
   { action: 'setDiscountedPrice', key: 'discounted' },
+  { action: 'setPriceTiers', key: 'tiers' },
 ]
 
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/prices-actions.js
```diff
@@ -1,12 +1,10 @@
...
-import {
-  buildBaseAttributesActions,
-} from './utils/common-actions'
-
+import { buildBaseAttributesActions } from './utils/common-actions'
 
 export const baseActionsList = [
   { action: 'changeValue', key: 'value' },
   { action: 'setDiscountedPrice', key: 'discounted' },
+  { action: 'setPriceTiers', key: 'tiers' },
```
markus-azer, line 7: Thats pretty easy to follow 👍 , and it will do the job, maybe also add todo to add the other two actions `addPriceTier`, `removePriceTier` which will make it more accurate, wdyt?

markus-azer (approved):
Looks good 👍 , just left one nitpick

dogayuksel (commented):
We should add a test case when price tiers on now and then are equal and created sync actions are empty.

dogayuksel (approved):

PR #1874, opened 2023-07-11
fix: handle long text values performantly

taylor-knapp:
#### Summary

Audit Log has run into a few edge cases where clients set VERY long `values` on product attributes. The result is that diffing takes a long time (can be > 7-10s or more).

#### Description

After running down where the delay is coming from, it appears it is due to `jsondiffpatch` uses `diff-match-patch` for long text diffing, which uses a very fine-grained diffing algorithm for long to easily identify what changed between two strings (this is actually the library originally built by google for google docs comparisions). However, we don't actually care about the specifics of what changed between two strings - we only care if the string changed at all. 

This PR adds a test in the first commit which shows the edge case (long values with a failing result b/c it should take ~5ms and it takes 5s). The second commit effectively disables the long text diffing approach by using the max safe integer as the value for `jsondiffpatch` to use `diff-match-patch` instead of the default diff. Now, long text strings diff very quickly.

All tests pass with this change, and based on the [getDeltaValue](https://github.com/commercetools/nodejs/blob/afe751ac52775eaaffa3f5769150e100858c6576/packages/sync-actions/src/utils/diffpatcher.js#L42-L70) method, I don't think there is any impact (the type of text diff used is resolved by that method, so disabling the long diff method will produce the same result - the new state of the string is what the method returns). That said - I'm not as familiar with the intricacies of this library and if disabling the long text diff algorithm produces any unintended consequences, so please think on that. 🙏 

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/utils/diffpatcher.js
+++ b/packages/sync-actions/src/utils/diffpatcher.js
@@ -20,11 +20,14 @@ const diffpatcher = new DiffPatcher({
     includeValueOnMove: false,
   },
   textDiff: {
-    // If the value to diff has a bigger length,
-    // a text diffing algorithm is used
-    // See https://github.com/benjamine/jsondiffpatch/
-    // blob/master/docs/deltas.md#text-diffs
-    minLength: 300,
+    /**
+     * jsondiffpatch uses a very fine-grained diffing algorithm for long strings to easily identify
+     * what changed between strings. However, we don't actually care about what changed, just
+     * if the string changed at all. So we set the minimum length to diff to a very large number to avoid
+     * using the very slow algorithm.
+     * See https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs.
+     */
+    minLength: Number.MAX_SAFE_INTEGER,
   },
 })
 
--- a/.changeset/violet-timers-act.md
+++ b/.changeset/violet-timers-act.md
@@ -1,0 +1,5 @@
+---
+'@commercetools/sync-actions': minor
+---
+
+Handle long text values performantly
--- a/packages/sync-actions/test/product-sync-variants.spec.js
+++ b/packages/sync-actions/test/product-sync-variants.spec.js
@@ -116,6 +116,78 @@ describe('Actions', () => {
     ])
   })
 
+  test('should handle long text values performantly', () => {
+    const longText = 'a'.repeat(10_000)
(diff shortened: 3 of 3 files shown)


taylor-knapp:
Hmm. Code coverage is complaining b/c now we never use the long text diff patch method. 🤔

qmateub (approved):
Given the added test and that this actually does not break any other stuff, I would vote for having it in 🚀 and Audit Log probably will be happy too 😅

markus-azer (approved):
@taylor-knapp Thats an amazing catch 🥇, and will be great help 👍 
`if disabling the long text diff algorithm produces any unintended consequences`,
@tdeekens , will this by any chance have side effect, wdyt?

packages/sync-actions/src/utils/diffpatcher.js
```diff
@@ -20,11 +20,14 @@ const diffpatcher = new DiffPatcher({
...
-    // a text diffing algorithm is used
-    // See https://github.com/benjamine/jsondiffpatch/
-    // blob/master/docs/deltas.md#text-diffs
-    minLength: 300,
+    /**
+     * jsondiffpatch uses a very fine-grained diffing algorithm for long strings to easily identify
+     * what changed between strings. However, we don't actually care about what changed, just
+     * if the string changed at all. So we set the minimum length to diff to a very large number to avoid
+     * using the very slow algorithm.
+     * See https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs.
```
tdeekens, line 28: Thanks for the comment 🏅.

tdeekens (approved):
I don't see there being a side-effect. Actually I don't know why we would ever use unidiff which is a character-based diffing thing to efficiently detect changes in parts of the text. We only care - to my knowledge - if there was a change and not where. The where would also result in a return value with the shape of `delta = [ unidiff, 0, 2 ]` which I haven't seen being used here at all. As it's actually costly I think we shouldn't be using it. I suspect the `300` being a somewhat 🪄al number with a reasoning of maybe "300 sounds large. I don't think we'll have larger values" when added longer time ago.

Keep in mind, last change here: 7 years ago.

CleanShot 2023-07-13 at 09 51 08@2x

PR #1890, opened 2024-04-30
chore: to replace fs.access with fs.exist

Rombelirk:
## Relates to 
- https://github.com/commercetools/fe-chapter-notes/issues/490

### DIFF
--- a/packages/csv-parser-discount-code/src/cli.js
+++ b/packages/csv-parser-discount-code/src/cli.js
@@ -6,6 +6,15 @@ import yargs from 'yargs'
 
 import CsvParserDiscountCode from './main'
 
+const doesFileExist = (filePath) => {
+  try {
+    fs.accessSync(filePath)
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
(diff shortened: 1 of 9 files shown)


ajimae (approved):

packages/csv-parser-discount-code/src/cli.js
```diff
@@ -6,6 +6,15 @@ import yargs from 'yargs'
 
 import CsvParserDiscountCode from './main'
 
+const doesFileExist = (filePath) => {
```
ajimae, line 9: Just a little observation, it will be nice if we can add a simple unit test to test this function.

ajimae:
Also, to make the testing easier, I will suggest you move `doesFileExist` as an exported function into a util or helper file and import them everywhere else where it is used.

packages/csv-parser-discount-code/src/cli.js
tdeekens: Yes but then we have a cross package dependency. We don't seen to have a shared package so duplication should be fine.

tdeekens (approved):

PR #1896, opened 2024-06-07
Add removeStagedChanges action for StandalonePrice

nima-ct:
#### **Summary**

Following the requirements here https://commercetools.atlassian.net/browse/PRC-2676, 
we want to be able to clear a staged value from an existing standalone price.
This PR adds a `removeStagedChanges` action.

#### Description

Docs https://docs.commercetools.com/api/projects/standalone-prices#remove-staged-changes
#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/test/price-sync.spec.js
+++ b/packages/sync-actions/test/price-sync.spec.js
@@ -8,14 +8,14 @@ const twoWeeksFromNow = new Date(Date.now() + 12096e5)
 /* eslint-disable max-len */
 describe('price actions', () => {
   test('action group list', () => {
-    expect(actionGroups).toEqual(['base', 'custom'])
+    expect(actionGroups).toStrictEqual(['base', 'custom'])
   })
 
   test('should not build actions if prices are not set', () => {
     const before = {}
     const now = {}
     const actions = pricesSync.buildActions(now, before)
-    expect(actions).toEqual([])
+    expect(actions).toStrictEqual([])
   })
 
   test('should not build actions if now price is not set', () => {
@@ -30,7 +30,7 @@ describe('price actions', () => {
     }
     const now = {}
     const actions = pricesSync.buildActions(now, before)
-    expect(actions).toEqual([])
+    expect(actions).toStrictEqual([])
   })
 
   test('should not build actions if there is no change', () => {
@@ -80,7 +80,7 @@ describe('price actions', () => {
       },
     }
     const actions = pricesSync.buildActions(now, before)
-    expect(actions).toEqual([])
+    expect(actions).toStrictEqual([])
   })
 
   describe('changeValue', () => {
@@ -106,7 +106,7 @@ describe('price actions', () => {
       }
 
       const actions = pricesSync.buildActions(now, before)
-      expect(actions).toEqual([
+      expect(actions).toStrictEqual([
         {
           action: 'changeValue',
           value: {
@@ -143,7 +143,7 @@ describe('price actions', () => {
       }
 
       const actions = pricesSync.buildActions(now, before)
-      expect(actions).toEqual([
+      expect(actions).toStrictEqual([
         {
           action: 'setDiscountedPrice',
           discounted: {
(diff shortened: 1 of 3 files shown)


packages/sync-actions/test/price-sync.spec.js
```diff
@@ -925,4 +925,43 @@ describe('price actions', () => {
...
+
+      const now = {
+        id: '1234',
+        value: {
+          type: 'centPrecision',
+          currencyCode: 'USD',
+          centAmount: 5000,
+          fractionDigits: 2,
+        },
+        staged: null,
```
ragafus, line 955: A couple of questions:

1. The object of the action `removeStagedChanges` does not include a `staged` property in the docs. Shouldn't we omit it?
2. Let's imagine we have the following now object:
```
const now = {
        id: '1234',
        value: {
          type: 'centPrecision',
          currencyCode: 'USD',
          centAmount: 5000,
          fractionDigits: 2,
        },
        staged: {
          type: 'centPrecision',
          currencyCode: 'EUR',
          centAmount: 5555,
          fractionDigits: 2,
        },
}
```
Which action is generated?
Is the

nima-ct: @ragafus 
1. yes, removed staged property
2. Do you mean adding a `should not build removeStagedChanges action` test?

ragafus: > @ragafus
> 
> 1. yes, removed staged property
> 2. Do you mean adding a `should not build removeStagedChanges action` test?

1.  I'm not sure if having `staged: undefined` in the generated object is going to create an error on BE side.
2. Yes

nima-ct: The staged property is not generated by the action anyway. I think the Jest equality matcher considered that both 
```
        {
          action: 'removeStagedChanges',
        }
```
```
         {
          action: 'removeStagedChanges',
          staged: undefined,
        }
```
are equal. That's why the assertion with the included `staged` property passes

2. Gonna add another test for it 👍

ragafus: > The staged property is not generated by the action anyway. I think the Jest equality matcher considered that both
> 
> ```
>         {
>           action: 'removeStagedChanges',
>         }
> ```
> 
> ```
>          {
>           action: 'removeStagedChanges',
>           staged: undefined,
>         }
> ```
> 
> are equal. That's why the assertion with the included `staged` property passes
> 
> 2. Gonna add another test for it 👍

1. You are right. Maybe we should switch to `toStrictEqual` matcher.

dogayuksel: Only having the `removeStagedChanges` action without the rest of the staged related actions is a bit confusing.

If the test data was like below, do we get a `removeStagedAction` and `changeValue` 🤔 It should be `applyStagedChanges` instead.

```
const now = {
    id: '1234',
    value: {
        type: 'centPrecision',
        currencyCode: 'USD',
        centAmount: 6000,
        fractionDigits: 2,
    },
};
```

And on @ragafus 's example, I would assume by looking at how other props are configured, we would get something like below 😕 
```
const action = {
    action: 'removeStagedChanges',
    staged: {
        type: 'centPrecision',
        currencyCode: 'EUR',
        centAmount: 5555,
        fractionDigits: 2,
    },
};
```

These staged related actions are designed implicitly stateful, seems to be not so compatible with the diff'er of sync-actions as is.

nima-ct: @dogayuksel thx for your input, indeed the action builder doesn't get it right in the case you mentioned.
As far as I understand the situation and your comment, does that imply that the sync-actions we are using to determine an action aren't sufficient to handle this case?
Let's say I introduce `applyStagedChanges`
```
  { action: 'removeStagedChanges', key: 'staged' },
  { action: 'applyStagedChanges', key: 'staged' },
```
It doesn't seem to me that our action method can tell the difference... or am I missing something?

PR #1908, opened 2024-08-28
fix(projects): fix myBusinessUnits and customerSearchStatus action keys

ChristianMoll:
Summary
Fix myBusinessUnit and customerSearch project actions.

Description
Some project actions I added recently had the wrong keys. I fixed those and split the actions into groups myBusinessUnit and customerSearch because changeMyBusinessUnitStatusOnCreation and changeCustomerSearchStatus have the same key: `status`. 

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/projects-actions.js
+++ b/packages/sync-actions/src/projects-actions.js
@@ -7,15 +7,24 @@ export const baseActionsList = [
   { action: 'changeLanguages', key: 'languages' },
   { action: 'changeMessagesConfiguration', key: 'messagesConfiguration' },
   { action: 'setShippingRateInputType', key: 'shippingRateInputType' },
+]
+
+export const myBusinessUnitActionsList = [
   {
     action: 'changeMyBusinessUnitStatusOnCreation',
-    key: 'myBusinessUnitStatusOnCreation',
+    key: 'status',
   },
   {
     action: 'setMyBusinessUnitAssociateRoleOnCreation',
-    key: 'myBusinessUnitAssociateRoleOnCreation',
+    key: 'associateRole',
   },
-  { action: 'changeCustomerSearchStatus', key: 'customerSearchStatus' },
+]
+
+export const customerSearchActionsList = [
+  {
+    action: 'changeCustomerSearchStatus',
+    key: 'status',
+  },
 ]
 
 export function actionsMapBase(diff, oldObj, newObj, config = {}) {
@@ -25,5 +34,23 @@ export function actionsMapBase(diff, oldObj, newObj, config = {}) {
     oldObj,
     newObj,
     shouldOmitEmptyString: config.shouldOmitEmptyString,
+  })
+}
+
+export function actionsMapBusinessUnit(diff, oldObj, newObj) {
+  return buildBaseAttributesActions({
+    actions: myBusinessUnitActionsList,
+    diff,
+    oldObj,
+    newObj,
+  })
+}
+
+export function actionsMapCustomer(diff, oldObj, newObj) {
+  return buildBaseAttributesActions({
+    actions: customerSearchActionsList,
+    diff,
+    oldObj,
+    newObj,
   })
 }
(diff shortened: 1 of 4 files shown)


ChristianMoll:
@commercetools/developer-tooling I split the actions into groups myBusinessUnit and customerSearch because changeMyBusinessUnitStatusOnCreation and changeCustomerSearchStatus have the same key: status.
Otherwise the mapping when building the actions doesn't work because of the non unique keys.
Is this the right approach here or can you think of a better alternative to handle this?

jenschude:
You should address @commercetools/beam-me-up-team

ChristianMoll:
@commercetools/beam-me-up-team I removed the groups for now but I am still unsure in which cases we use groups.

NickDevG (approved):

packages/sync-actions/src/projects.js
```diff
@@ -16,6 +20,14 @@ function createChannelsMapActions(mapActionGroup, syncActionConfig) {
       )
     )
 
+    allActions.push(
+      mapActionGroup('base', () => actionsMapBusinessUnit(diff, oldObj, newObj))
+    )
+
+    allActions.push(
+      mapActionGroup('base', () => actionsMapCustomer(diff, oldObj, newObj))
+    )
```
markus-azer, line 29: i would suggest, to add two new ActionGroups ex: businessUnit.

`base` is mainly used for 1st lvl attributes ex: name, countries, etc..
plus `packages/sync-actions/src/projects-actions.js` shows that you already naming it perfectly diff than `base`

`actionsMapCustomer`: if im not mistaken thats more of generating the search config actions, so maybe more descriptive naming

packages/sync-actions/src/projects-actions.js
```diff
@@ -7,15 +7,26 @@ export const baseActionsList = [
...
     action: 'setMyBusinessUnitAssociateRoleOnCreation',
     key: 'myBusinessUnitAssociateRoleOnCreation',
+    actionKey: 'associateRole',
+  },
+]
+
+export const customerSearchActionsList = [
+  {
+    action: 'changeCustomerSearchStatus',
+    key: 'status',
```
markus-azer, line 28: That actions, i'm so confused about, it shows on [Representations](https://docs.commercetools.com/api/projects/project#customersearchstatus) but cant map it to the upper lvl [draft](https://docs.commercetools.com/api/projects/project#project).
its either im not getting it or documentation linking isnt right, wdyt?

packages/sync-actions/src/projects-actions.js
```diff
@@ -27,3 +38,37 @@ export function actionsMapBase(diff, oldObj, newObj, config = {}) {
...
+  const { searchIndexing } = diff
+
+  if (!searchIndexing) {
+    return []
+  }
+
+  const { customers } = searchIndexing
+  if (!customers) {
+    return []
+  }
```
markus-azer, line 66: That means there is sth wrong with doc ref: [SearchIndexingConfiguration](https://docs.commercetools.com/api/projects/project#searchindexingconfiguration), right?

markus-azer (commented):
@ChristianMoll Thanks Christian for fix 💯 💪 
Would confirm what @jenschude mentioned that sync-action package is mainly maintained/released by BMU.

Im missing little bit the context, so i left few comments.

packages/sync-actions/src/projects.js
ChristianMoll: Hi @markus-azer , thanks for the feedback! I split the actions into three groups accordingly and renamed `actionsMapCustomer` to be more descriptive.

packages/sync-actions/src/projects-actions.js
ChristianMoll: It was indeed missing from the documentation. I opened an issue to add it: https://github.com/commercetools/commercetools-docs/issues/5506

packages/sync-actions/src/projects-actions.js
ChristianMoll: I opened an issue to add the missing field: https://github.com/commercetools/commercetools-docs/issues/5506

markus-azer (approved):
looks pretty good 💯

PR #1914, opened 2024-09-13
fix: filter out unchanged taxrates from changed list

antoniolodias:
#### Summary

[support ticket](https://commercetools.atlassian.net/browse/SUPPORT-28326)
When trying to update one tax rate inside a tax category, all the tax rates for the tax category are updated by the API.

#### Description

This happens because there's no `update` action for taxRates, but only `replace`, so any taxRate sent to the API is taken to be replaced.

Tried just filtering out from the FE, but that conflicted with `remove` logic

This PR just filters out unchanged tax rates from the action

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [x] `Type` label for the PR

### DIFF
--- a/.changeset/honest-hats-push.md
+++ b/.changeset/honest-hats-push.md
@@ -1,0 +1,5 @@
+---
+"@commercetools/sync-actions": patch
+---
+
+fix: filter out unchanged taxrates from changed list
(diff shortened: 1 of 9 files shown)


Ibidapo (approved):

ajimae (approved):
Great job @antoniolodias,

Even though it's not strictly needed, I still think it will be nice if we can add an test assertion to show that the `unchanged tax rates` were completely omitted in the resulting action.

PR #1919, opened 2025-01-22
Remove unwanted "require" usage in sync-actions package

CarlosCortizasCT:
#### Summary

In this PR we're getting rid of a `require` statement that (according to the code comment above it) we wanted to remove and that's also causing some troubles for consumers who have a ES modules setup.

#### Description

Our codebase is using ES module which handles modules dependencies using `import/export` statements. However, we have one dependency (`jsondiffpatch`) that does not support such statements as it's bundles only with CommonJS exports. Thus, we needed to use an unwanted `require` statement to actually be able to use it in our `sync-actions` package.

My first attempt was to update the `jsondiffpatch` dependency to its latest version ([v0.6.0](https://github.com/benjamine/jsondiffpatch/releases/tag/v0.6.0)) as they have migrated to ES modules in that one. There's one caveat though with that version as they also dropped support for any NodeJS version older than 18 and we need to support version 14 and 16 in this repository packages.

Looking for alternatives to make a CommonJs dependency work in a ES module context, I found out we can configure `rollout` (our build tool) to use a plugin that can make the conversion from one format to the other under the hood so we can import that dependency just like the other.
Also, I realized the plugin I found about [is already configured in this repository](https://github.com/commercetools/nodejs/blob/master/rollup.config.js#L43) so actually we don't need anything extra, it's just a matter of changing how we import the `jsondiffpath` package.

I noticed that the "unwanted" import is 7 years old while the rollup plugin config is 6 years old so probably we forgot we could change that import back when we introduced the rollup plugin.

### DIFF
--- a/packages/sync-actions/package.json
+++ b/packages/sync-actions/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "fast-equals": "^2.0.0",
-    "jsondiffpatch": "^0.4.0",
+    "jsondiffpatch": "0.6.0",
     "lodash.flatten": "^4.4.0",
     "lodash.foreach": "^4.5.0",
     "lodash.isequal": "^4.5.0",
--- a/packages/sync-actions/package.json
+++ b/packages/sync-actions/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "fast-equals": "^2.0.0",
-    "jsondiffpatch": "0.6.0",
+    "jsondiffpatch": "0.4.0",
     "lodash.flatten": "^4.4.0",
     "lodash.foreach": "^4.5.0",
     "lodash.isequal": "^4.5.0",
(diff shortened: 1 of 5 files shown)


packages/sync-actions/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "fast-equals": "^2.0.0",
-    "jsondiffpatch": "^0.4.0",
+    "jsondiffpatch": "0.4.0",
```
CarlosCortizasCT, line 39: Having the caret in the version is misleading as the semantic versioning does not work for versions in the zero major version.
Even if there's a new zero version (as it is in this case), it won't be used.

The reasoning for that is that zero versions are not considered stable so breaking changes can be expected in minor versions, which is a risk from the consumers point of view.
Actually, it happens with this dependency as version `0.6.0` is a breaking change one.

markus-azer (approved):
Thanks @CarlosCortizasCT 💯 
thanks for also explaining `caret`, was about to ask

PR #1921, opened 2025-02-12
feat(sync-actions): provide option to generate empty update actions for ommitted properties

kafis:
#### Summary

The import-api is committed to unset properties of a resource if the supplied request omits the specified keys.

The sync-actions currently only generate unset actions if the supplied key is present but null. This PR introduces two flags, that controls this behaviour. The default behaviour is kept the same, if this flags are not supplied.

More information can be found in the respective [SUPPORT ticket](https://commercetools.atlassian.net/browse/SUPPORT-30603?focusedCommentId=888857)

Upon approval, I will run a changeset command and include it in this PR.

##### shouldUnsetOmittedProperties

This flag controls, wether an omitted property should lead to generating an updateAction:
 
This is how it is intended to work:

```typescript

describe('`shouldUnsetOmittedProperties`', () => {
      test('if true => should generate unset actions for omitted properties', () => {
        before = {
          name: { en: 'Foo' },
          description: 'description',
          externalId: '123',
          slug: { en: 'foo' },
          customerNumber: 'customer-number',
          quantityOnStock: 1,
        }
        now = {}
        actions = buildBaseAttributesActions({
          actions: testActions,
          diff: diffpatcher.diff(before, now),
          oldObj: before,
          newObj: now,
          shouldOmitEmptyString: false,
          shouldUnsetOmittedProperties: true,
        })
        expect(actions).toEqual([
          { action: 'changeName' },
          { action: 'setDescription' },
          { action: 'setExternalId' },
          { action: 'changeSlug' },
          { action: 'setCustomerNumber' },
          { action: 'changeQuantity' },
        ])
      })
      test('if false => should not generate unset actions for omitted properties', () => {
        before = {
          name: { en: 'Foo' },
          description: 'description',
          externalId: '123',
          slug: { en: 'foo' },
          customerNumber: 'customer-number',
          quantityOnStock: 1,
        }
        now = {}
        actions = buildBaseAttributesActions({
          actions: testActions,
          diff: diffpatcher.diff(before, now),
          oldObj: before,
          newObj: now,
          shouldOmitEmptyString: false,
          shouldUnsetOmittedProperties: false,
        })
        expect(actions).toEqual([])
      })
    })
```
##### shouldPreventUnsettingRequiredFields

This flag controls, wether required fields should be unset.

The convention in update actions, according to my research through the doc, is, that update actions that start with `set...` are allowed to unset. 
 
This is how it is intended to work:

```typescript

describe('`shouldPreventUnsettingRequiredFields`', () => {
      test('if true => should not generate unset actions for required fields', () => {
        before = {
          name: { en: 'Foo' },
          description: 'description',
          externalId: '123',
          slug: { en: 'foo' },
          customerNumber: 'customer-number',
          quantityOnStock: 1,
        }
        now = {
          name: null,
          description: null,
          externalId: null,
          slug: null,
          customerNumber: null,
          quantityOnStock: null,
        }
        actions = buildBaseAttributesActions({
          actions: testActions,
          diff: diffpatcher.diff(before, now),
          oldObj: before,
          newObj: now,
          shouldPreventUnsettingRequiredFields: true,
        })
        expect(actions).toEqual([
          { action: 'setDescription' },
          { action: 'setExternalId' },
          { action: 'setCustomerNumber' },
        ])
      })
      test('if false => should generate unset actions for required fields', () => {
        before = {
          name: { en: 'Foo' },
          description: 'description',
          externalId: '123',
          slug: { en: 'foo' },
          customerNumber: 'customer-number',
          quantityOnStock: 1,
        }
        now = {
          name: null,
          description: null,
          externalId: null,
          slug: null,
          customerNumber: null,
          quantityOnStock: null,
        }
        actions = buildBaseAttributesActions({
          actions: testActions,
          diff: diffpatcher.diff(before, now),
          oldObj: before,
          newObj: now,
          shouldPreventUnsettingRequiredFields: false,
        })
        expect(actions).toEqual([
          { action: 'changeName' },
          { action: 'setDescription' },
          { action: 'setExternalId' },
          { action: 'changeSlug' },
          { action: 'setCustomerNumber' },
          { action: 'changeQuantity' },
        ])
      })
    })
  })
```
#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR

### DIFF
--- a/packages/sync-actions/src/utils/common-actions.js
+++ b/packages/sync-actions/src/utils/common-actions.js
@@ -18,6 +18,7 @@ export const createIsEmptyValue = (emptyValues) => (value) =>
  * @param  {Object} options.oldObj - the object that needs to be updated
  * @param  {Object} options.newObj - the new representation of the object
  * @param {Boolean} options.shouldOmitEmptyString - a flag to determine if we should treat an empty string a NON-value
+ * @param {Boolean} options.shouldUnsetOmittedProperties - a flag to determine if we should unset fields which are omitted in the newObj
  */
 export function buildBaseAttributesActions({
   actions,
@@ -25,6 +26,7 @@ export function buildBaseAttributesActions({
   oldObj,
   newObj,
   shouldOmitEmptyString,
+  shouldUnsetOmittedProperties,
 }) {
   const isEmptyValue = createIsEmptyValue(
     shouldOmitEmptyString ? [undefined, null, ''] : [undefined, null]
@@ -38,6 +40,7 @@ export function buildBaseAttributesActions({
       const now = newObj[key]
       const isNotDefinedBefore = isEmptyValue(oldObj[key])
       const isNotDefinedNow = isEmptyValue(newObj[key])
+      const isOmitted = !{}.hasOwnProperty.call(newObj, key)
       if (!delta) return undefined
 
       if (isNotDefinedNow && isNotDefinedBefore) return undefined
@@ -47,10 +50,10 @@ export function buildBaseAttributesActions({
         return { action: item.action, [actionKey]: now }
 
       /* no new value */
-      if (isNotDefinedNow && !{}.hasOwnProperty.call(newObj, key))
+      if (isNotDefinedNow && isOmitted && !shouldUnsetOmittedProperties)
         return undefined
 
-      if (isNotDefinedNow && {}.hasOwnProperty.call(newObj, key))
+      if (isNotDefinedNow)
         // value unset
         return { action: item.action }
 
--- a/packages/sync-actions/src/utils/common-actions.js
+++ b/packages/sync-actions/src/utils/common-actions.js
@@ -45,7 +45,7 @@ export function buildBaseAttributesActions({
       const now = newObj[key]
       const isNotDefinedBefore = isEmptyValue(oldObj[key])
       const isNotDefinedNow = isEmptyValue(newObj[key])
-      const isOmitted = !{}.hasOwnProperty.call(newObj, key)
+      const isOmitted = !Object.keys(newObj).includes(key)
       if (!delta) return undefined
 
       if (
--- a/packages/sync-actions/src/utils/common-actions.js
+++ b/packages/sync-actions/src/utils/common-actions.js
@@ -8,6 +8,9 @@ const normalizeValue = (value) =>
 export const createIsEmptyValue = (emptyValues) => (value) =>
   emptyValues.some((emptyValue) => emptyValue === normalizeValue(value))
 
+/* actions that start with 'set' can generate undefined valued */
+export const isOptionalField = (action) => action.startsWith('set')
+
 /**
  * Builds actions for simple object properties, given a list of actions
  * E.g. [{ action: `changeName`, key: 'name' }]
@@ -19,6 +22,7 @@ export const createIsEmptyValue = (emptyValues) => (value) =>
  * @param  {Object} options.newObj - the new representation of the object
  * @param {Boolean} options.shouldOmitEmptyString - a flag to determine if we should treat an empty string a NON-value
  * @param {Boolean} options.shouldUnsetOmittedProperties - a flag to determine if we should unset fields which are omitted in the newObj
+ * @param {Boolean} options.shouldPreventUnsettingRequiredFields - a flag to determine if required fields should be unset
  */
 export function buildBaseAttributesActions({
   actions,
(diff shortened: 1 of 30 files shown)


packages/sync-actions/test/utils/common-actions.spec.js
```diff
@@ -116,6 +112,55 @@ describe('Common actions', () => {
...
+          shouldOmitEmptyString: false,
+          shouldUnsetOmittedProperties: true,
+        })
+        expect(actions).toEqual([
+          { action: 'changeName' },
+          { action: 'setDescription' },
+          { action: 'setExternalId' },
+          { action: 'changeSlug' },
+          { action: 'setCustomerNumber' },
+          { action: 'changeQuantity' },
```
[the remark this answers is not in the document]
kafis: This is a very valid concern and I'm very grateful for the feedback. I will look for a suitable solution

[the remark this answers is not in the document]
kafis: > one thing which is seems to be pattern, change is for required fields or that cant be un-setted.
> while set is always for fields that could be removed.

I went through the documentation and could not find a counter example. I do therefore agree

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -8,6 +8,9 @@ const normalizeValue = (value) =>
 export const createIsEmptyValue = (emptyValues) => (value) =>
   emptyValues.some((emptyValue) => emptyValue === normalizeValue(value))
 
+/* actions that start with 'set' can generate undefined valued */
+export const isOptionalField = (action) => action.startsWith('set')
```
markus-azer, line 12: Thats smart utils @kafis 💯

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -19,6 +22,7 @@ export const createIsEmptyValue = (emptyValues) => (value) =>
  * @param  {Object} options.newObj - the new representation of the object
  * @param {Boolean} options.shouldOmitEmptyString - a flag to determine if we should treat an empty string a NON-value
  * @param {Boolean} options.shouldUnsetOmittedProperties - a flag to determine if we should unset fields which are omitted in the newObj
+ * @param {Boolean} options.shouldPreventUnsettingRequiredFields - a flag to determine if required fields should be unset
```
markus-azer, line 25: somehow that flag helps to generate invalid actions, `{ action: 'changeName' },` 🤔.

maybe to make the package backward compatible, change this flag to be warper for the new added functionality.
so in case the user wanted new behaviour, the `xyz` flag set to true, otherwise the old behaviour works, which will enable us to pump the version as `patch`, wdyt?!

hope that i'm not trolling you :)

packages/sync-actions/test/utils/common-actions.spec.js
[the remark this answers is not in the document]
tdeekens: Sorry, I am not following: are we planning to make a change here to only include any update action that starts with `set*`?

packages/sync-actions/test/utils/common-actions.spec.js
```diff
@@ -116,6 +112,120 @@ describe('Common actions', () => {
         ])
       })
     })
+    describe('`shouldUnsetOmittedProperties`', () => {
+      test('if true => should generate unset actions for omitted properties', () => {
```
tdeekens, line 116: Nit: I'd not mimic argument naming in a test but describe intended behaviour instead. I assume this test suite doesn't follow that pattern much but for this it would mean something like:

```js
describe('omitting properties', () => {
  describe('when omitted properties should be unset', () => {});
  describe('when omitted properties should not be unset', () => {});
});
```

The advantage I see would be that tests are less coupled to implementation and can act more as an independent specification of it.

packages/sync-actions/src/utils/common-actions.js
```diff
@@ -38,19 +45,27 @@ export function buildBaseAttributesActions({
       const now = newObj[key]
       const isNotDefinedBefore = isEmptyValue(oldObj[key])
       const isNotDefinedNow = isEmptyValue(newObj[key])
+      const isOmitted = !{}.hasOwnProperty.call(newObj, key)
```
tdeekens, line 48: Nit: 
```suggestion
      !Object.keys(newObj).includes(key)
```

tdeekens (approved):

PR #1924, opened 2025-03-04
fix(sync-actions): exclude 'chalk' package from being required in final build artifact by changing import statement

kafis:
#### Summary

Changing the import statement does remove `chalk` from being included in the `sync-action.cjs.js` artefact.

#### Todo

- Tests
  - [ ] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


yarn.lock
```diff
@@ -1350,6 +1350,23 @@
   resolved "https://registry.yarnpkg.com/@commercetools/sdk-middleware-logger/-/sdk-middleware-logger-2.1.2.tgz#6b8ce77699bf5d8e5ec4f243a2c9d691094ef0c7"
   integrity sha512-uGhhAUnZzLWO3ozzz6VDMUAKr5Y9ctVyqfExXVC7t37IgppLsKrrvLl6VMoU2uXjIGSNA0VN0Kxgj93rBBejTQ==
 
+"@commercetools/sync-actions@^5.0.0":
```
tdeekens, line 1353: I don't follow why these change.

tdeekens:
Tried to integrate this into a browser bundle. We still see the error of process not being defined as `chalk` seems to be in the bundle. 

Screenshot 2025-03-04 at 09 21 24@2x

PR #1927, opened 2025-03-05
Fix sync-actions package not working on a browser environment

CarlosCortizasCT:
#### Summary

It seems the changes we introduced [here](https://github.com/commercetools/nodejs/pull/1919) made the package compatible with `vite` but created a new issue that made the `sync-actions` package not usable in a browser environment.

#### Description

Because of the new `import`, it looks like the `jsondiffpatch` package is pulling the `chalk` dependency and it's using an outdated version which relies on the `process` global object which does not exist in a browser context.

In this PR we update again the `import` and add a `resolutions` config to make sure we pull a newer version of `chalk` which does not rely on the global `process` object.

#### Todo

- Tests
  - [x] Unit
  - [ ] Integration
  - [ ] Acceptance
- [ ] Documentation
- [ ] `Type` label for the PR


packages/sync-actions/src/utils/diffpatcher.js
```diff
@@ -1,4 +1,4 @@
-import { DiffPatcher } from 'jsondiffpatch'
+import { DiffPatcher } from 'jsondiffpatch/dist/jsondiffpatch.cjs'
```
CarlosCortizasCT, line 1: At least in my tests, adding the resolution config only is not enough.

kafis (approved):
LGTM

tdeekens (approved):

PR #1930, opened 2025-03-06
Dropping support for Node.js versions which have EOL

tdeekens:
##### Requires Node.js v18 or later

This releases migrates packages to require Node.js v18 or later. Ideally you should be already using Node.js v20 or later. According to [Node.js Releases](https://nodejs.org/en/about/previous-releases) Node.js v18 will be in maintenance and reach End of Life by the end of April. 

Other than requiring Node.js v18 packages with this releases do not contain any internal breaking changes. 

For the Node.js versions for which we dropped support:

1. The end-of-life date for Node.js 14 is April 30, 2023
2. The end-of-life date for Node.js 16 is September 11, 2023

Both versions do not receive bug fixes, features or security updates. We should not recommend our customers or us using them. Updating to newer versions (v18+) as usually no breaking changes for consumers after all.


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
CarlosCortizasCT, line 1: Can you please share why we need this configuration? 🙏

tdeekens: Not needed. Came as a result of me using a newer yarn locally.

mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+node = "22"
```
CarlosCortizasCT, line 2: Can you share what is this file for? 🙏

packages/category-exporter/package.json
```diff
@@ -1,18 +1,15 @@
...
+    "node": ">=18"
   },
   "name": "@commercetools/category-exporter",
   "version": "4.0.0",
   "description": "Exports categories from the commercetools platform.",
   "main": "lib/main.js",
-  "bin": {
-    "category-exporter": "bin/category-exporter.js"
-  },
+  "bin": "bin/category-exporter.js",
```
CarlosCortizasCT, line 12: Can you explain this change? 🙏

CarlosCortizasCT (approved):
Thanks a lot for working on this 🙇

mise.toml
tdeekens: Upsa.

packages/category-exporter/package.json
tdeekens: Not sure I did this. Must have been some tooling. Let me revert.

tdeekens: [471ba06](https://github.com/commercetools/nodejs/pull/1930/commits/471ba060beef74d56f125daa3615be177a582d7d)

mise.toml
tdeekens: [7092de6](https://github.com/commercetools/nodejs/pull/1930/commits/7092de6ed2dd3464e14ea64604ba4b1d51828787)

tdeekens:
Once approved by the team I'll adjust the branch protections to match the new Node.js versions accordingly.

kafis (approved):
🥇

PR #1931, opened 2025-03-07
Migrate from `yarn` to `pnpm` and drop `lerna` in the process

tdeekens:
#### Summary

This migrates from `yarn` to `pnpm` which we adopted across repositories. It's speedier for installs are more strict about dependency resolution.

At the same time we can drop `lerna` in the process and replace it with `manypkg` which enforces dependency rules across the repository.


.circleci/config.yml
```diff
@@ -3,44 +3,6 @@ version: 2.1
...
-      - v1-yarn-cache-node_22
-
-  - &save_yarn_cache_node_22
-    key: v1-yarn-cache-node_22-{{ .Branch }}-{{ checksum "yarn.lock" }}
-    paths:
-      - ~/.cache/yarn
-
-  - &yarn_install
-    name: Installing
-    command: yarn --frozen-lockfile
```
tdeekens, line 42: We can simplify large parts of this:

1. We don't need a cache per Node.js version. Package managers don't install differently based on Node.js versions
2. We can migrate this into a command. That way it's native to CircleCI and validated by its language server also locally

.circleci/config.yml
```diff
@@ -56,72 +18,117 @@ executors:
       - image: cimg/node:22.14.0
     working_directory: *working_directory
 
+commands:
+  pnpm_install:
```
tdeekens, line 22: This is the command to replace everything above.

.circleci/config.yml
```diff
@@ -56,72 +18,117 @@ executors:
...
 jobs:
   install_test_node_18:
     executor: node_18
     steps:
       - checkout
-      - restore_cache: *restore_yarn_cache_node_18
-      - run: *yarn_install
-      - save_cache: *save_yarn_cache_node_18
+      - pnpm_install
+      - run:
```
tdeekens, line 70: I suggest not to build as a post install. It's not needed and frustrating. If we don't do that we must do it here.

.github/workflows/release.yml
```diff
@@ -26,28 +26,19 @@ jobs:
...
-        run: echo "::set-output name=dir::$(yarn cache dir)"
-
-      - uses: actions/cache@v4
+      - name: Setup Node (uses version in .nvmrc)
+        uses: actions/setup-node@v4
         with:
-          path: ${{ steps.yarn-cache.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
```
tdeekens, line 47: Large parts here can be offloaded to the action-setup.

.husky/commit-msg
```diff
@@ -0,0 +1 @@
+pnpm commitlint --edit $1
```
tdeekens, line 1: Husky has to be updated in the process. It has now files here instead of a global config file.

.nvmrc
```diff
@@ -1 +1 @@
-18
+22
```
tdeekens, line 1: Let's bump this. We just develop against the latest version we support to detect regressions early.

integration-tests/cli/product-json-to-csv.it.js
```diff
@@ -179,10 +179,8 @@ describe('CSV and CLI Tests', () => {
 
           it('contains `description`', () => {
             const description = {
-              en:
-                'The light jackets of Save the Duck keep us cozy warm. The slight',
-              de:
-                'Die leichten Freizeitjacken von Save the Duck halten uns wohlig',
+              en: 'The light jackets of Save the Duck keep us cozy warm. The slight',
```
tdeekens, line 182: A whole lot of formatting was wrong. This fixes that.

integration-tests/package.json
```diff
@@ -33,20 +33,20 @@
     "@commercetools/sdk-middleware-queue": "^3.0.0",
     "@commercetools/sdk-middleware-user-agent": "^3.0.0",
     "@commercetools/state-importer": "^4.0.0",
-    "@commercetools/sync-actions": "^5.0.0",
+    "@commercetools/sync-actions": "^6.1.1",
```
tdeekens, line 36: Dependency versions should be in sync in the repo. manypkg takes care of that.

package.json
```diff
@@ -1,46 +1,61 @@
...
   },
   "scripts": {
-    "postinstall": "check-node-version --package --print && cross-env NODE_ENV=development yarn build",
-    "clean": "rm -rf coverage; lerna exec -- rimraf lib dist",
+    "preinstall": "npx only-allow pnpm",
+    "postinstall": "./scripts/postinstall.sh",
+    "clean": "rm -rf coverage; manypkg exec rimraf lib dist node_modules; rm -rf node_modules",
     "commit": "git-cz",
-    "build": "lerna run build --no-private",
+    "build": "pnpm --recursive run build",
```
tdeekens, line 13: pnpm has this which allows us to drop lerna.

package.json
```diff
@@ -1,46 +1,61 @@
...
     "package-readme-entry": "./scripts/package-readme-entry.sh",
     "changeset": "changeset",
     "changeset:version-and-format": "changeset version && prettier --write --parser json '**/package.json'"
   },
-  "workspaces": [
-    "integration-tests",
-    "packages/*"
-  ],
-  "dependencies": {
+  "resolutions": {
```
tdeekens, line 41: Just formatting.

package.json
```diff
@@ -1,46 +1,61 @@
...
+    "npm-user-validate": "1.0.1",
+    "serialize-javascript": "6.0.0",
+    "diff": "5.0.0",
+    "object-path": "0.11.8",
+    "npm": "6.14.17",
+    "set-value": "4.0.1",
+    "chalk": "3.0.0"
+  },
+  "packageManager": "pnpm@9.15.6",
+  "devDependencies": {
```
tdeekens, line 58: Semi important: root package.json files should not contain deps but devDeps. We use things in there for development.

scripts/postinstall.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
tdeekens, line 1: Easier to maintain than an inline run script.

tsconfig.json
```diff
@@ -29,6 +29,7 @@
     "target": "esnext",
     "allowJs": false,
     "baseUrl": ".",
+    "skipLibCheck": true,
```
tdeekens, line 32: Needed at the moment for one package. I don't see a major downside. I plan to enable this later again.

.circleci/config.yml
```diff
@@ -56,72 +18,117 @@ executors:
...
       - run:
           name: Running tests
           # Limiting the workers of Jest to 10
           # as the build otherwise dies due to resource restrictions.
-          command: yarn test:ci
+          command: pnpm test:ci
       - run:
           name: Running test (with coverage report)
-          command: yarn test:ci
+          command: pnpm test:ci
```
islam3zzat, line 85: I'm not sure why we used to run the tests twice!

islam3zzat (approved):
This clearly involved a great deal of effort, thank you so much!

PR #1934, opened 2025-03-12
[Sync-actions]: adding business unit search status actions

NickDevG:
#### Summary

Adding business unit search status actions

### DIFF
--- a/packages/sync-actions/src/projects-actions.js
+++ b/packages/sync-actions/src/projects-actions.js
@@ -36,17 +36,6 @@ export const businessUnitSearchActionsList = [
   },
 ]
 
-export const searchIndexActionsList = [
-  {
-    action: 'changeCustomerSearchStatus',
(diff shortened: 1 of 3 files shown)


packages/sync-actions/src/projects-actions.js
```diff
@@ -29,6 +29,24 @@ export const customerSearchActionsList = [
...
+
+export const searchIndexActionsList = [
+  {
+    action: 'changeCustomerSearchStatus',
+    key: 'status',
+  },
+  {
+    action: 'changeBusinessUnitSearchStatus',
+    key: 'status',
+  },
```
leventekobor, line 47: why is it duplicated here? 🤔

NickDevG: Leftover, nice catch

leventekobor (approved):

PR #1947, opened 2025-03-26
fix: to remove custom import resolver for eslint

tdeekens:
#### Summary

This should be removed. It's not working with the workspace protocol.


tdeekens:
Merging to unblock main.

PR #1960, opened 2025-05-06
Downgrade `jsondiffpatch` to `v0.6.2`

tdeekens:
#### Summary

This downgrades the underlying library of `jsondiffpatch` to v0.6.2 as any future version can cause issues in non ESM contexts. 

The issue appears to be in this version. As the changelog of the library is not well maintained the root cause is hard to track down.

Screenshot 2025-05-06 at 12 56 32@2x

From the last versions this is the last working for us

Screenshot 2025-05-06 at 12 54 50@2x

Without this change in Jest you would see something like

Screenshot 2025-05-06 at 12 57 10@2x

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -19,6 +19,10 @@
     {
       "matchPackageNames": ["flow-bin"],
       "allowedVersions": "0.141.0"
+    },
+    {
+      "matchPackageNames": ["jsondiffpatch"],
+      "allowedVersions": "<= 0.7.2"
     }
   ],
   "lockFileMaintenance": {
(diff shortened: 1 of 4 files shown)


.renovaterc.json
```diff
@@ -19,6 +19,10 @@
     {
       "matchPackageNames": ["flow-bin"],
       "allowedVersions": "0.141.0"
+    },
+    {
+      "matchPackageNames": ["jsondiffpatch"],
+      "allowedVersions": "<= 0.6.2"
```
tdeekens, line 25: We should investigate but not update accidentally again.

tdeekens:
I will merge this as `master` is not working.

PR #1969, opened 2025-06-16
PDM-236 Add support for product level attributes

peter-gerhard:
https://commercetools.atlassian.net/browse/PDM-236

Add Support for importing products, product drafts with product level attributes

- [X] Add Action Generation for Product Level Attributes on Products
   - [X] Refactor while keeping tests green
   - [X] Add logic for product level attribute actions
   - [X] Add tests for product level attribute actions and get them green
- [x] Add Action Generation for Product Level Attributes on Product-Types
   - [X] The only action affected is addAttributeDefinition which takes a definition. By adding the `level` to the raml (other repo) it will automatically work to import product types with product level attributes
   
   
I suggest t review commit by commit to get a more meaningful diff

### DIFF
--- a/packages/sync-actions/src/products.js
+++ b/packages/sync-actions/src/products.js
@@ -19,6 +19,7 @@ const actionGroups = [
   'references',
   'prices',
   'pricesCustom',
+  'productAttributes',
   'variantAttributes',
   'images',
   'variants',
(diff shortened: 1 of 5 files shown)


packages/sync-actions/src/products.js
```diff
@@ -19,7 +19,8 @@ const actionGroups = [
   'references',
   'prices',
   'pricesCustom',
-  'attributes',
```
markus-azer, line 22: while it makes alot of sense to rename it, i would avoid renaming `attributes` to `productAttributes`, cause that could be breaking change which require major upgrade.

there is multi ways to introduce it without breaking, the easiest would be keep it and only introduce `productAttributes`

